PR #5, opened 2021-05-14
refactors code for reserves, adds fixed reserves

tmtmtl30:
refactors reserve code heavily. also adds code for fixed-size reserves, and integrates them with overmap dynamic event generation


code/modules/mapping/space_management/space_reservation.dm
```diff
@@ -1,35 +1,139 @@
...
+/datum/turf_reservation/proc/get_reserved_turfs()
+	SHOULD_NOT_OVERRIDE(TRUE)
+	var/turf/bottom_left = locate(bottom_left_coords[1], bottom_left_coords[2], bottom_left_coords[3])
+	var/turf/top_right = locate(top_right_coords[1], top_right_coords[2], top_right_coords[3])
+	return block(bottom_left, top_right)
+
+/datum/turf_reservation/proc/get_non_border_turfs()
+	SHOULD_NOT_OVERRIDE(TRUE)
+	var/turf/bottom_left = locate(bottom_left_coords[1]+1, bottom_left_coords[2]+1, bottom_left_coords[3])
+	var/turf/top_right = locate(top_right_coords[1]-1, top_right_coords[2]-1, top_right_coords[3])
```
MarkSuckerberg, line 56: When we add padding for mesons and such (if we do at all) we need to remember to replace +/- 1 with a #define-d or variable amount

MarkSuckerberg (approved):

code/modules/mapping/space_management/space_reservation.dm
tmtmtl30: they're already padded, it's integrated into the reserve sizes themselves -- the +/- 1 is to account for the 1-tile impassable border, which is included inside the reserve because it's placed on the tiles of the reserve

PR #18, opened 2021-05-21
maybe fixes broken bone bug

tmtmtl30:
when my broken bone was fixed during the test, it didn't speed me up. this maybe fixes that? i don't really know (because i haven't tested it), but it's worth a shot


MarkSuckerberg (approved):
Worth a shot

PR #84, opened 2021-07-16
Makes portable seed extractors craftable 

tiramisuapimancer:
## About The Pull Request
Adds a recipe to the misc tab for a portable seed extractor
image
Most the materials you need to make them are fairly easy to get if you have an autolathe, and you should probably have an autolathe. Ditto for necessary botanist equipment unrelated to seed extraction.

## Why It's Good For The Game
To be honest, it'd be better if it was just possible to huck seeds out of a fruit with a knife, but that would totally fuck with slicing code, and a whole bunch of other things, and I have utterly no idea how to do it, and there's no way it won't be annoying to even a competent coder to do. 

The point of it is just to make ghetto botany more viable, and makes it so you can get seeds without having to loot botanical equipment- just make a pile of dirt on the ground and chuck seeds into it, you shouldn't need to be fancy. And we're all about ghetto here at Shitpest

## Changelog
:cl:
add: crafting recipe for portable seed extractors
/:cl:


whitesands/code/datums/components/crafting/recipes.dm
```diff
@@ -14,3 +14,15 @@
...
+	name = "Portable seed extractor"
+	reqs = list(
+			/obj/item/storage/bag/plants,
+			/obj/item/plant_analyzer,
+			/obj/item/stock_parts/manipulator,
+			/obj/item/stack/cable_coil = 2)
+	result = /obj/item/storage/bag/plants/portaseeder //this will probably mean that you can craft portable seed extractors into themselves, sending the other materials into the void, but we still don't have a solution for recipes involving radios stealing your headset, so this is officially not my problem. "no, Tills-The-Soil, adding more analyzers and micro-manipulators to your portable seed extractor does not make it make more seeds. in fact it does exactly nothing."
+	time = 20
+	tools = list(TOOL_SCREWDRIVER, TOOL_WIRECUTTER)
+	category = CAT_MISC
```
MarkSuckerberg:
```suggestion
	category = CAT_MISC

```

PR #94, opened 2021-07-23
removes shuttle IDs, and a lot of associated code

tmtmtl30:
shuttle IDs are a black cloud that hangs over every piece of code that so much as touches shuttles. this fixes that, by removing them. it also reorganizes shuttle map files into directories, and switches the shuttle manipulator's categories to work off the same organization.

the overmap also no longer uses IDs; half the reason it used IDs is because shuttles had to.

in the process of removing shuttle IDs, a lot of code has been removed, broken, or reworked. the vast majority of this code was no longer used anyway, so it's not that big a deal. changes include but are probably not limited to:
- removed turbolifts
- removed aux base
- removed mechanical curtains (???)
- removed escape pod-specific functionality
- removed steel rain
- removed shuttle creation tool (it sucked and i don't think anyone used it anymore. the code was awful and might not have worked anymore anyway)
- removed nav computer used to select docking location on ruins, for simulated shuttles and nonsimulated shuttles
- basic shuttle computers (like the one used to control the lavaland shuttle) have had all their subtypes removed; none of them are currently functional. they weren't used anymore anyway though
- nukie war declaration can probably be repeated multiple times now. there were some dumb ass global lists and i hated them so they're gone. this will never be important ever
- the fugitive pinpointer and the functionality of the nukie pinpointer allowing it to point at the nukie ship have been removed (neither was used anymore and they're easy enough to reimplement)
- labor claim console can no longer request to move the labor shuttle
- removed shuttle loan event
- removed arrivals shuttle
- supply shuttle and emergency shuttle no longer dock anywhere. they use global variables on SSshuttle now, which are currently uninitialized since the shuttles are never important
- removed some config settings relevant to removed features
- some other stuff i've missed

the shuttle manipulator is a little anemic at the moment. i might patch it up later

there are a few tiny code quality changes in places as well. also this definitely absolutely needs to be tested live because i can't have caught every potential bug so some things are probably broken. sorry


MarkSuckerberg (approved):

MarkSuckerberg:
LGTM

PR #136, opened 2021-08-09
Adds extremely basic placeholder trading methods

MarkSuckerberg:
## About The Pull Request
Adds the selling pad and console, as well as fixes the express cargo console (which I never realized was broken until just now) for the purpose of basic trading. Not added on maps yet, will likely do before PR is merged

**CODE NOTE:** I made it so that all objects that are initialized in /area/ship areas run connect_to_shuttle() after init if they're not being loaded at map placement time. I think this will be good for consistency so that all ship-based things get it called so they do shuttle-specific initialization, but I don't know if it's good for performance/how connect_to_shuttle() was meant to be used. I have it done in a decently performant (I think) way by just checking the area for its mobile_port var, but I realize adding even one small thing to *every single atom's* init proc can really add up.

## Why It's Good For The Game
Honestly I don't think it is that good, since it'll reduce player dependency on each other for supplies, but at the moment it feels like this is somewhat needed just to make it so supplies *can* be replenished.

## Changelog
:cl:
add: Selling pads and their respective consoles to enable ship crews to sell unneeded or rare items and then purchase them with an express console.
fix: Express cargo consoles work again.
/:cl:

### DIFF
--- a/_maps/shuttles/shiptest/amogus_sus.dmm
+++ b/_maps/shuttles/shiptest/amogus_sus.dmm
@@ -446,6 +446,10 @@
 "gq" = (
 /turf/closed/wall/mineral/titanium,
 /area/ship/crew/office)
+"gC" = (
+/obj/machinery/computer/cargo/express,
+/turf/open/floor/plasteel,
+/area/ship/cargo)
 "gI" = (
 /obj/effect/turf_decal/bot,
 /obj/structure/closet/crate,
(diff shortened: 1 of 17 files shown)


MarkSuckerberg:
Multitool linkage works, and selling pad icons actually show up again

WindowsErrors:
Screenshot (937)

PR #161, opened 2021-08-14
Porting over Kryson's carpets.

PetMudstone:
## About The Pull Request

Ports over Kryson's carpets from https://github.com/tgstation/tgstation/pull/54612. Does not add the spawners nor any way to obtain them at the moment. 
Below images copied from relevant PR.
Executive NT carpet.
image
Stellar carpet.
image
Donk Co. promotional carpet.
image

## Why It's Good For The Game
Carpets look pretty. Could be used later by mappers or by players. Also, Donk carpets are *integral* to Donk Co. RP.

## Changelog
:cl: Pet Mudstone, Kryson
add: Ported over some of Kryson's special carpets: Donk Co., stellar, and executive NT.
/:cl:

PR #163, opened 2021-08-14
Add annotation feature to `check_regex.py`

martinlyra:
## About The Pull Request

Due to an issue pestering #161 over whitespaces there. I was asked to add the feature to output a more verbose log that can help annotate files that have errors or matches. Currently, it writes that info into a separate output file (name currently hardcoded into the script, I should consider adding arguments down the future), so it does not show up and clog the `stdout` log.

## Why It's Good For The Game

Makes the script less of a PITA to use and understand. What is even point of writing it, let alone using it if it only offers no help?

## Changelog
:cl:
code: check_regex.py now outputs an file with full info on files with matches and which lines were matched
/:cl:


ZephyrTFA:
Speed merging to fix master

PR #265, opened 2021-09-16
Fixes prescription glasses not being given to the jobs that should have them with nearsighted.

### LINKED ISSUE #266: [BUG] Permanent HUD from prescription healthHUD glasses and probably the other prescription huds (not mesons) from gear.
## Testmerges: 
Can't get right now but shouldn't be the fault of any of them anyways.

## Reproduction: 
Give a character nearsighted and any eyewear gear item, said item gets overriden by regular prescription. Give them prescription healthhud and they also get a permanent healthHUD from roundstart and no way to get rid of it too, very fun, security is likely to also be the same, science is untested and I'm unsure, no permanent mesons. Has been the case probably since these were added to the loadout options.

Ms-Mee:
## About The Pull Request
This PR makes it so that it gives a job it's relevant type of prescriptionhud when they have nearsighted, which was supposed to be handled by gear but these get overriden and stay as a permanent hud instead, which is no good, as such it closes #266. It also makes it look for a suitable spot for your glasses instead of just overriding them. May warrant a rewrite of the example nearsighted quirk at some point. 

## Why It's Good For The Game
It makes nearsighted characters not rely on the ship actually having the proper type of prescriptionhud AND makes it so nearsighted characters can have another type of eyewear on them at roundstart from gear, job or otherwise, why not let them have an eyepatch anyways?

## Changelog
:cl:
tweak: Removed prescriptionHUDs from gear and made them be given by the nearsighted quirk instead, which doesn't override your starting glasses anymore.
fix: No more roundstart permanent HUD from deleted gear item.
/:cl:

### DIFF
--- a/code/datums/traits/negative.dm
+++ b/code/datums/traits/negative.dm
@@ -260,6 +260,8 @@
 	name = "Nearsighted"
 	desc = "You are nearsighted without prescription glasses, but spawn with a pair."
 	value = -1
+	var/obj/item/glasses
+	var/where
 	gain_text = "<span class='danger'>Things far away from you start looking blurry.</span>"
 	lose_text = "<span class='notice'>You start seeing faraway things normally again.</span>"
 	medical_record_text = "Patient requires prescription glasses in order to counteract nearsightedness."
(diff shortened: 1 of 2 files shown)


Ms-Mee:
#266 Almost there.
Just the equip thing which I'll probably just reuse Family Heirloom instead and see if that works.

Ms-Mee:
Alright I think this should be work but untested PR, perfect for PRtest

fighterslam:
tested this, works just fine. Compiles with no errors, code works pretty well, good PR

fighterslam (approved):
tested this, works just fine. Compiles with no errors, code works pretty well, good PR

code/datums/traits/negative.dm
```diff
@@ -260,6 +260,8 @@
 	name = "Nearsighted"
 	desc = "You are nearsighted without prescription glasses, but spawn with a pair."
 	value = -1
+	var/obj/item/glasses
+	var/where
```
ZephyrTFA: Move these below the existing vars.
Its general a good idea to have new vars below existing ones.

Ms-Mee: I mean, I did it like that because Family Heirloom places those there, but checking again most quirks that have the new vars place them together below, by wanting to keep consistency I actually decreased it which is kinda funny

PR #267, opened 2021-09-16
Phazon Core-Preference Message Removal

untrius:
## About The Pull Request

Bugfix for https://github.com/Whitesands13/Whitesands/issues/825.

The Phazon says it requires a bluespace anom core to complete, but the type of anom core doesn't matter.  I removed the message saying a non-bluespace core is incorrect, and two descriptions of Phazon parts implying the anom core must be bluespace.

image
image
image

## Why It's Good For The Game

Makes misleading descriptions match actual functionality.

Lets me get my first PR on this repo out of the way, so I can start opening an inordinate number of issues.

## Changelog
:cl:
fix: Phazons no longer say they need bluespace cores
/:cl:

### DIFF
--- a/code/game/mecha/mecha_parts.dm
+++ b/code/game/mecha/mecha_parts.dm
@@ -234,12 +234,10 @@
 
 /obj/item/mecha_parts/chassis/phazon/attackby(obj/item/I, mob/user, params)
 	. = ..()
-	if(istype(I, /obj/item/assembly/signaler/anomaly) && !istype(I, /obj/item/assembly/signaler/anomaly/bluespace))
-		to_chat(user, "The anomaly core socket only accepts bluespace anomaly cores!")
 
 /obj/item/mecha_parts/part/phazon_torso
 	name="\improper Phazon torso"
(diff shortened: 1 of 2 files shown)


Sun-Soaked:
Do we actually have a way to get anomaly cores right now? Besides certain ships with SM engines.

Remember that random events are disabled, and also broken.

untrius:
I took a look and there's a couple ways in the code to get them, but I'm not sure if they all still apply.  You can get them from completing the power puzzle ruin, destroying an ash walker nest, and they sometimes drop when killing demonic watchers.

code/game/mecha/mecha_parts.dm
```diff
@@ -234,12 +234,10 @@
 
 /obj/item/mecha_parts/chassis/phazon/attackby(obj/item/I, mob/user, params)
```
ZephyrTFA: You can drop the entire proc override since all it does is can and return the super de
efinition now

untrius: Removed

Issue #232, opened 2021-09-04
[BUG] Buy a cricket, get a raven


AsciiSquid:
## Round ID:[165](https://shiptest.ga/stats/rounds/165)
Bought a cricket-class, got a raven-class with missing tiles in the main room, get Neo to force-load a new ship
Nobody joins my ship, I'm blaming this bug instead of anything reasonable.

## Testmerges:
#108: '[MDB IGNORE] Baytiles/the great ship prettyfication' by retlaw34 at commit abd9fd0b09
#218: '[DNM] Swifter's Ship Emporium!' by swifterjack at commit 3d1186454b
#130: 'bay clicky noises' by retlaw34 at commit 281b0945b6
#154: 'Disable Events' by ZephyrTFA at commit f821d7d77e
#179: 'Update Define Spacing' by ZephyrTFA at commit 289cc24feb

## Reproduction:
Buy a cricket, suffer.

untrius:
I've bought several cricket-class ships on my test server with no issue, and was able to buy one on live yesterday.

fighterslam:
Is this still an issue?

PR #279, opened 2021-09-21
Stops Issues from Getting Tags From Referenced Testmerges

untrius:
## About The Pull Request

The keywordlabeler bot has been adding the "do not merge" tag to several issues.  This was caused by it searching the body of the issues and seeing the [DNM] string included in the titles of some of the testmerges.  An example is #232.

This can be fixed by telling the bot to stop searching the body of the issue when deciding what tags to add.  However there doesn't appear to be a way to tell it to treat issues and PRs differently, so that means it will also not search a PR's body.

This will make it so the search strings for tags must be in the title of PRs and issues.  However it appears this is already the standard.  I don't know if this is the way you want this fixed, so I'm leaving this here as an option.

### DIFF
--- a/.github/keylabeler.yml
+++ b/.github/keylabeler.yml
@@ -2,7 +2,7 @@
 matchTitle: true
 
 # Determines if we search the body (optional). Defaults to true.
-matchBody: true
+matchBody: false
 
 # Determines if label matching is case sensitive (optional). Defaults to true.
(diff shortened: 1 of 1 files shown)


untrius:
As an example of said behavior, the bot added the "do not merge" tag because [DNM] is in the description of the PR.

PR #284, opened 2021-09-22
[DNM] Missing Vendor Item Sprites Workaround

untrius:
## About The Pull Request

This PR is a temporary workaround for the missing vendor item sprites.  It can be test merged while I work on the full fix.

When a vendor is created, they call Initialize() then build_inventory(), which puts references to all the vendor's items into GLOB.vending_products.

Then when the game is starting, /datum/asset/spritesheet/vending/register() is eventually called, which loads the vendor item sprites named in GLOB.vending_products into the vendor spritesheet.  This appears to be the only time this function runs.

Any vendor that is created after /datum/asset/spritesheet/vending/register() runs, will load its items into GLOB.vending_products, but those item sprites will not be loaded into the vendor spritesheet.  Therefore there is no sprite for the vendor to display.

This workaround functions by forceloading all the vendor items into GLOB.vending_products by putting all the vendors onto CentCom at round start.  This is not a good long term solution, and I will be investigating other options in the near future.

## Why It's Good For The Game

Makes it easier to tell items apart in vendors.

## Changelog
:cl:
fix: Item stickers have been taped back onto the vending machines
/:cl:

### DIFF
--- a/_maps/map_files/generic/CentCom.dmm
+++ b/_maps/map_files/generic/CentCom.dmm
@@ -4620,6 +4620,14 @@
 	},
 /turf/open/floor/plasteel,
 /area/centcom/control)
+"lQ" = (
+/obj/effect/turf_decal/corner/brown,
+/obj/effect/turf_decal/corner/brown{
+	dir = 4
+	},
+/obj/machinery/vending/donksofttoyvendor,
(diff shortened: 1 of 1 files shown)


untrius:
I have a full fix for this written, but it still needs some refactoring.

PR #314, opened 2021-10-04
People orbited by wisps now *spin two times faster

super12pl:
## About The Pull Request
Makes mobs orbited by wisps to spin faster when using *spin.
## Why It's Good For The Game
That is a very good question, but first we need to get two definitions from *legitemate* scientific sites:
Celestial body is an aggregation of matter in space that can be considered a single unit
A Satellite(moon) is a celestial body orbiting a different, larger celestial body

Therefor a spaceman in space and a wisp orbiting him can be defined as a celestial body and it's satellite
Now keeping that in mind look at this graph about planets in our solar system
https://imgur.com/J9PNOm3

You can clearly make out that the more moons a body has the faster it's rotation speed is
why wouldn't our celestial body(the player) be different?

Concluding, its only logical to assume due to laws of physics and astronomy that a person orbited by a wisp would spin faster than a normal human

## Changelog
:cl:
tweak: People orbited by wisps now spin two times faster.
/:cl:


Sun-Soaked:
The TRUE mining loot rework

Nobelium-XVIII:
speedmerge

MarkSuckerberg (approved):
lol

Issue #368, opened 2021-10-20
[BUG] Mining vendors give out stuff for free


TheNeoGamer42:
Issue reported from  Round ID: N/A

Reporting client version: 514.1557

Mining vendors give out shit for free. Specifically the newer ones ascii added.

![image](https://user-images.githubusercontent.com/58159167/138024420-92819dd1-8532-405d-9cbb-bc1cf4fc97b6.png)

This also results in advanced trollphysics where you buy the resupply vendor for this to get ALOT of free shit and a net gain of atleast a couple thousand credits.

triplezeta:
sounds like the mining vendor selections should be rebalanced too.

Sun-Soaked:
vendor selection balance is fine when it's functional- the issue is just freebies, and that specifically has to do with the fact that manually building mining vendors causes their cost to net to zero.
If I remember correctly from convos at the time, we actually removed them from most ships in the first place specifically because they were unlimited and that caused various troll physics exploits- so they should probably be safe to re-add now? 
Once this is fixed, at least.

untrius:
This is actually all the vendors.  They have a check for if the vendor is on station that always returns false.  When the vendor isn't on station, everything is free.  The on station check touches a few different things, so it may not be the most simple fix in the world.

PR #384, opened 2021-10-25
[Fix] Vendors Are No Longer Free

untrius:
## About The Pull Request

All items in both types of vendors (creds and mining points) were free.  This was due to a onstation check in the base vendor type.  If a vendor was not on the station z level, all items would be free.  I removed this check and the references to its result in the tgui js.

Overrides to onstation had to be removed from some of the map files.

## Why It's Good For The Game

Bugfix #368

## Changelog
:cl:
fix: Vendors are no longer free
/:cl:


swifterjack:
make sure that there is a way to manually activate this with vv, so that people dont need accounts for it even off station

PR #437, opened 2021-11-16
Adds Table Bells

triplezeta:
## About The Pull Request

This adds table bells. As of the time of making this PR there are two variants: normal and brass.

You click on them to make noise. Extremely spammable. Pick one up like a paper bin. It plays a small little animation when you ring it.

Ding sound effect recorded by Voltriar.

Also mappers/admins can change the sound effect used with varedit if they so choose to. I know exactly what this will be used to make. I have no regrets.

## Why It's Good For The Game

Cute little prop for places like the Boyardee to use. 

## Changelog
:cl:
add: Table bells
/:cl:


whitesands/code/modules/client/loadout/loadout_general.dm
```diff
@@ -156,6 +156,11 @@
 	path = /obj/item/dyespray
 	cost = 700
 
+/datum/gear/tablebell
+	display_name "table bell, brass"
```
triplezeta:
```suggestion
	display_name = "table bell, brass"
```

PR #471, opened 2021-11-26
IPC Burn Mod Nerf

zestybastard:
## About The Pull Request

ok im pretty sure my computer would stop working if i hit it with a welder but come on people

## Why It's Good For The Game

im tired of dying to two laser hits, 3 welder whacks, and a partridge in a pear tree (with a flamethrower)

## Changelog
:cl:
balance: Changes the IPC burn mod to 1, and the heat mod to 1.5.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/IPC.dm
+++ b/code/modules/mob/living/carbon/human/species_types/IPC.dm
@@ -24,8 +24,8 @@
 	limbs_id = "synth"
 	mutant_bodyparts = list("ipc_screen", "ipc_antenna", "ipc_chassis")
 	default_features = list("ipc_screen" = "BSOD", "ipc_antenna" = "None")
-	burnmod = 1.5
-	heatmod = 1
+	burnmod = 1
+	heatmod = 1.5
 	brutemod = 1
 	siemens_coeff = 1.5
 	reagent_tag = PROCESS_SYNTHETIC
--- a/code/modules/mob/living/carbon/human/species_types/IPC.dm
+++ b/code/modules/mob/living/carbon/human/species_types/IPC.dm
@@ -24,7 +24,7 @@
 	limbs_id = "synth"
 	mutant_bodyparts = list("ipc_screen", "ipc_antenna", "ipc_chassis")
 	default_features = list("ipc_screen" = "BSOD", "ipc_antenna" = "None")
-	burnmod = 1
+	burnmod = 1.25
 	heatmod = 1.5
 	brutemod = 1
 	siemens_coeff = 1.5
--- a/code/modules/mob/living/carbon/human/species_types/IPC.dm
+++ b/code/modules/mob/living/carbon/human/species_types/IPC.dm
@@ -24,8 +24,8 @@
 	limbs_id = "synth"
 	mutant_bodyparts = list("ipc_screen", "ipc_antenna", "ipc_chassis")
 	default_features = list("ipc_screen" = "BSOD", "ipc_antenna" = "None")
-	burnmod = 2
-	heatmod = 1.5
+	burnmod = 1.5
+	heatmod = 1
 	brutemod = 1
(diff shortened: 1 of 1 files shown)


Sun-Soaked:
salut

Sun-Soaked:
honestly I would maybe make the heat mod the 1.5 and the burn mod normal- you cannot burn a computer with a lighter, but you could totally break it via overheating

zestybastard:
> honestly I would maybe make the heat mod the 1.5 and the burn mod normal- you cannot burn a computer with a lighter, but you could totally break it via overheating

Done and done, thanks.

super12pl:
I'm all fun and flavour over balance but I really don't think removing the main downside from a race immune to like 80% of things in the game is a good change.

Oh and update changelog

ZephyrTFA:
Main issue with this is that EMPs, which are their only real other counter, aren't widely available

PrefabQuasar:
> Main issue with this is that EMPs, which are their only real other counter, aren't widely available

it takes 5 metal and 1 uranium to make a pocket EMP
i would say theyre pretty available

super12pl:
> > Main issue with this is that EMPs, which are their only real other counter, aren't widely available
> 
> it takes 5 metal and 1 uranium to make a pocket EMP
> i would say theyre pretty available

You don't really pvp most of the time now do you?

zestybastard:
> I'm all fun and flavour over balance but I really don't think removing the main downside from a race immune to like 80% of things in the game is a good change.
> 
> Oh and update changelog

I don't mean to remove the only downside so much as make it slightly less incredibly punishing. As it stands 2 stray shots from any laser will drop an IPC with relative consistency, and a misstep into a hot area is a guaranteed death sentence. Not to mention, IPCs also get their balance by exclusion (cannot benefit from chems at all, and so on) rather than just convenience and damage tweaks. Also, the brute mod is still like 2x or something.

super12pl:
> I don't mean to remove the only downside so much as make it slightly less incredibly punishing. As it stands 2 stray shots from any laser will drop an IPC with relative consistency, and a misstep into a hot area is a guaranteed death sentence. 

If you want to make it less punishing change the values to 1.5 burn and 1.5 heat rather than removing burn mod completely.

>Not to mention, IPCs also get their balance by exclusion (cannot benefit from chems at all, and so on) rather than just convenience and damage tweaks. 

Thats a fair point however welding tools and cable coils are still more common than medical supplies, while being more effective because  they can be used on a dead ipc that doesn't  require defibbing.

>Also, the brute mod is still like 2x or something.

It's 1, the code is in front of you

ZephyrTFA:
We can have a debate without attacking others or making off-topic asides.
If we can't I'll relock this PR discussion.

zestybastard:
Aight, fine I'll change the numbers again. I swear they take roughly the same brute damage as kepori though, at least as far as I've seen tested.

Sun-Soaked:
As time goes on, I have a feeling that power scarcity mechanics will help balance out IPC's, and at that point we may even have to push their raw resistance to higher than baseline

Sun-Soaked:
I also feel like we're kind of glazing over the fact that, while IPC's can definitely be more powerful in certain situations (oxygen, radiation, diseases) they're just as vulnerable to a bullet as everyone else, which is the fundamental baseline of PVP balance.

Sun-Soaked:
Just going to quickly point out that our quick healing solutions (injectors, legion cores, healmixes, styptic, etc.) have no effect on ipcs, making their only option to do a relatively slow healing based on resources that are often fairly limited for most shiptests.  In light of that, the realistic effect of damage on IPC's is already _significantly higher_ then usual, and wounding them generally means you've confirmed a kill, since self-repair is so annoying. This is combined with the IPC instant-kill effect that is EMP's, which are extremely easy to get here due to our current material glut.
I honestly see no reason to keep the burn vulnerability.

PR #553, opened 2021-12-27
Fixes infinite iron bug for budget 9mm mag. 

ghost:
## About The Pull Request
Adds a materials list to the budget 9mm magazine so that its 20000 iron that goes back into the autolathe not 30000.

## Why It's Good For The Game
Prevents people from getting infinite iron. 

## Changelog
:cl:
add: added a materials list to the zip ammo.
/:cl:


MarkSuckerberg:
Good catch

PR #560, opened 2021-12-29
Makes movement_force based on thrust capacity

MarkSuckerberg:
## About The Pull Request
Removes snowflake movement_force vars from ships, makes movement force calculated based on the amount of total thrust the ship has. Should it be based on mass too? Probably, but that's too expensive in my opinion to calculate everytime, and not consistent enough to calculate once and forget about.

## Why It's Good For The Game
Makes having fewer engines (e.g. on smaller ships) actually have an upside.

## Changelog
:cl:
tweak: Makes movement_force based on total thrust. This means that ships with a lot of thrusters will stun and possibly even throw, while smaller ships may just not stun at all.
/:cl:

### DIFF
--- a/_maps/RandomRuins/RockRuins/rockplanet_crash_cult.dmm
+++ b/_maps/RandomRuins/RockRuins/rockplanet_crash_cult.dmm
@@ -1,176 +1,37 @@
 //MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
-"ad" = (
-/obj/effect/decal/cleanable/dirt/dust,
-/obj/effect/decal/cleanable/blood,
-/obj/effect/turf_decal/corner/brown{
-	dir = 4
-	},
-/obj/effect/turf_decal/corner/brown{
-	dir = 1
+"as" = (
+/obj/effect/spawner/structure/window/shuttle,
+/obj/machinery/door/poddoor{
+	id = "whiteship_bridge"
 	},
 /turf/open/floor/plating,
 /area/ruin/unpowered)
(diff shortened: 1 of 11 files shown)


terg-but-ius:
with ships with more thrust like the osprey things get thrown around with wild abandon and can result in grievous injury, may wanna adjust that

EDIT: i also get the feeling that many a map was _not_ designed with this sorta thing in mind but this is a testmerge so m

ZephyrTFA:
If you dont want to be thrown around by the idiot captain who blasts off at the speed of light, use a chair or locker

ZephyrTFA:
or lynch them, im not your mom

PrefabQuasar:
> If you dont want to be thrown around by the idiot captain who blasts off at the speed of light, use a chair or locker

it does this on launch regardless of what people do, outside of turning off engines specifically for taking off and landing, and only applies to larger ships, because it takes only thrust into account and not mass, when it should REALLY apply thrust to ships like pills, and not the ospray

Apogee-dev:
Just wanted to note that this PR's made large ships particularly hellish to use, especially since they usually have more objects on tables to fling everywhere in the first place. I think a lot of players would appreciate if this was either based on thrust:mass ratio or removed altogether. A _lot_ of ships use tables extensively and it's especially punishing for jobs like chef that are more about laying things out in a way that's useful and pleasing. I can't tell you how many times I've walked into an Osprey's canteen only to see it littered with debris and broken glasses.

MarkSuckerberg:
I was going to adjust the numbers (and said so on the discord) but never got around to it or mentioned it on the actual PR, so it got merged before I could do that.

PR #566, opened 2021-12-29
[Fix] Prevent Elzu Infinite Charge from Syringes

untrius:
## About The Pull Request

Explanation:

Elzu used to be able to get free charge by removing their blood (liquid electricity) with a syringe, then putting it back for 1% per unit.  This could be repeated as many times as desired, to quickly raise charge.

The best solution to this issue is to separate "Elzu blood" and "charge-giving chemical" into two different reagents.  For the time being, I've decided to give them standard blood.

I like the idea of giving them a custom blood reagent with custom blood decals.  But the blood decal code likely needs to be refactored before this is done, as it is difficult to extend currently.  This is why I am not including this at this time.

Changes:

The exotic_blood var was removed from ezlu to make them default to using standard blood.

The liquid electricity blood pack and its references were removed (was also hardcoded onto Tide-Class).

Removed modularization comments I put in ethereal.dm over a year ago, and made number of newlines consistent.

Made description of blood pack crate no longer give a specific number of packs, because it will surely continue to change.

image

image

## Why It's Good For The Game

Free charging was clearly unintentional.

## Changelog
:cl:
tweak: Elzu have standard blood
fix: Elzu can no longer use a syringe for free charge
/:cl:

### DIFF
--- a/_maps/shuttles/shiptest/isv_roberts.dmm
+++ b/_maps/shuttles/shiptest/isv_roberts.dmm
@@ -605,7 +605,6 @@
 /obj/structure/closet/crate/freezer,
 /obj/item/reagent_containers/blood/AMinus,
 /obj/item/reagent_containers/blood/BPlus,
-/obj/item/reagent_containers/blood/ethereal,
 /obj/item/reagent_containers/blood/random,
 /obj/item/organ/ears/cat,
 /obj/item/organ/tail/cat,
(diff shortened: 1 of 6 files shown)


tiramisuapimancer:
blood decal refactror so elzu can have fullbright layer glowy blood decale :floshed:

terg-but-ius:
> For the time being, I've decided to give them standard blood.

Side effect: Vampire Elzuose will become real (maybe)

Ms-Mee:
> > For the time being, I've decided to give them standard blood.
> 
> Side effect: Vampire Elzuose will become real (maybe)

Doesn't really change much if they are *intended* not to have regular blood

CursedBirb:
What about giving them cellulose as blood(grinded wood). That might fit into new lore too

PR #524, opened 2021-12-17
Adds the ability for APC Powercords to drain energy off Ethereals

Kapu1178:
## About The Pull Request
APC Powercords (IPC's powercord implant) can drain energy off ethereals at a rate of 1:5.

## Why It's Good For The Game
Literal energy beings, smh/

## Changelog
:cl:
add: IPCs can drain energy off Ethereals with their powercord.
/:cl:

PR #567, opened 2021-12-30
Scale Up Elzu Charge

untrius:
## About The Pull Request

I had previously raised the amount of charge that Elzu would give and take from APCs.
https://github.com/Whitesands13/Whitesands/pull/567

Due to the syringe bug, I disabled this.
https://github.com/Whitesands13/Whitesands/pull/609

Now that it has been removed, I am turning this back on.  This is done by changing one constant, ETHEREAL_CHARGE_SCALING_MULTIPLIER.

The PR that allows IPCs to charge off of Elzu (https://github.com/shiptest-ss13/Shiptest/pull/524) was unaware of this, so I modified that code as well:
Elzu lose 2% per cycle, instead of 8%
Efficiency was lowered from 500% to 100%, so power doesn't come from nowhere (APCs have 20% efficiency)
These cancel with the charge scale multiplier being 20, so IPCs still receive 40 charge per cycle

## Why It's Good For The Game

Makes it so draining and charging things as an Elzu has an impact.

## Changelog
:cl:
tweak: Elzu can be the inducers they were born/grown to be
/:cl:


Kapu1178:
I should port the electrical species refactor from bee that stops IPCs and Ethereals from using essentially arbitrary numbers for how their charges are controlled and standardizes it.

Kapu1178 (approved):

PR #593, opened 2022-01-04
Update Ghostize and Re-Enter Corpse to check for a mob being in NullSpace

### LINKED ISSUE #592: Lag issue with bluespace jumped corpses
Issue reported from  Round ID: [560](https://shiptest.ga/stats/rounds/560) (Shiptest)

Reporting client version: 514.1568

Heya, I'm Cronos Morphis in game, PromethiusX in Byond, and Massattack52 both here and in the Discord. I'd like to report a pretty straightforward issue. My corpse was brought onto a ship that subsequently blue space jumped out of existence. At the time, I was a ghost and outside of my body. Later on, out of curiosity, I attempted the 're-enter corpse' command, and something strange happened.

I suffered a huge lag spike that disconnected me from the server. Not so bad right? Wrong. Everyone else on the server reported severe lag around the same time. Chances are whatever processes attempted to place me back into a body that no longer existed had a hell of a time doing just that.

I was able to recreate this by re-entering my corpse two more times, just for safe measure.

Proposed Solutions: 

1. Make sure data for mobs is fully deleted during blue space jumps if this doesn't already occur.
2. Somehow make it so the re-enter corpse command doesn't work if the corpse was part of a blue space jump.

TLDR; Don't attempt to enter your bluespace jumped corpse, the server will have a bad time.

ZephyrTFA:
## About The Pull Request

When a dead body has ghostize called on it with the re-enterable flag set to false, if they are already a ghost it doesnt update correctly; which can cause issues when the ghost attempts to re-enter their body, which is is NullSpace (bad).

## Why It's Good For The Game

Server has a heartattack attempting to render nullspace, because it isnt intended to be rendered. Lets prevent that.

## Changelog
:cl:
tweak: Ghostize and Re-Enter respect nullspace/existing ghosts
/:cl:

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -284,6 +284,7 @@ Works together with spawning an observer, noted above.
 			ghost.client.init_verbs()
 			if(!can_reenter_corpse)	// Disassociates observer mind from the body mind
 				ghost.mind = null
+				key = null
 			return ghost
 
 /*
@@ -343,7 +344,7 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 	set name = "Re-enter Corpse"
 	if(!client)
 		return
-	if(!mind || QDELETED(mind.current))
+	if(!mind || QDELETED(mind.current) || mind.current.loc == null)
 		to_chat(src, "<span class='warning'>You have no body.</span>")
 		return
 	if(!can_reenter_corpse)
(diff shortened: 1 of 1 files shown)


Kapu1178:
Looping through the dead mob list every time feels hacky

ZephyrTFA:
> Looping through the dead mob list every time feels hacky

Ghostize is very rarely called, and it's better than doing for(var/mob/dead/observer in world)

Kapu1178:
Why not just use locate() instead? I know its technically the same thing, but it saves on line space.

ZephyrTFA:
This is a performance fix for the server and prevents clients from breaking shit hard; so I'm self merging this

PR #595, opened 2022-01-04
Removing the skip delay on AACs

Omppusolttu:
## About The Pull Request

AACs are disliked due to their slow speed and having a delay on skipping it provides no practical benefit really as skipping the re- or depressurization is a shiptester's choice in the end.

## Why It's Good For The Game

Makes AACs more user-friendly

## Changelog
:cl:
del: Removed AAC skip delay
/:cl:

### DIFF
--- a/code/game/machinery/airlock_cycle_control.dm
+++ b/code/game/machinery/airlock_cycle_control.dm
@@ -95,7 +95,7 @@
 	var/scan_on_late_init = FALSE
 	var/depressurization_margin = 10 // use a lower value to reduce cross-contamination
 	var/overlays_hash = null
-	var/skip_delay = 300
+	var/skip_delay = 10
 	var/skip_timer = 0
 	var/is_skipping = FALSE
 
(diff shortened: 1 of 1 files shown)


code/game/machinery/airlock_cycle_control.dm
```diff
@@ -129,12 +129,12 @@
 	qdel(wires)
 	wires = null
 	cut_links()
-	SSair.stop_processing_machine(src)
```
jupyterkat: that proc is there for a reason y'know

code/game/machinery/airlock_cycle_control.dm
```diff
@@ -129,12 +129,12 @@
...
+	SSair.atmos_machinery -= src
 	return ..()
 
 /obj/machinery/advanced_airlock_controller/Initialize(mapload)
 	. = ..()
-	SSair.start_processing_machine(src)
```
jupyterkat: so is this

code/game/machinery/airlock_cycle_control.dm
Omppusolttu: Is that a really recent addition? because I didn't touch that and I updated my codebase a few days ago.

jupyterkat: yeah, just got added today

Omppusolttu: Yeah I'll update my code and then remake my changes then

Omppusolttu:
Unfucked

MarkSuckerberg (approved):
Can probably look into removing the skip delay fundamentally later, but this works for now.

PR #597, opened 2022-01-04
Messages admins when someone touches a controller

Kapu1178:
## About The Pull Request
This is really hacky, I know. I tried to do it directly on topic vv, but god references are beyond me.

## Why It's Good For The Game
Staff should *really* know when someone is playing with fire.

## Changelog
:cl:
admin: VVing a subsystem alerts admins.
code: VVing a reference now calls vv_alert_admins if it resolves as a datum.
/:cl:

### DIFF
--- a/code/modules/admin/view_variables/view_variables.dm
+++ b/code/modules/admin/view_variables/view_variables.dm
@@ -15,10 +15,6 @@
 	if(!islist && !istype(D))
 		return
 
-	if(istype(D, /datum/controller))
-		var/datum/controller/ss = D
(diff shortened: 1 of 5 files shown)


code/modules/admin/view_variables/view_variables.dm
```diff
@@ -15,6 +15,10 @@
 	if(!islist && !istype(D))
 		return
 
+	if(istype(D, /datum/controller/))
```
MarkSuckerberg:
```suggestion
	if(istype(D, /datum/controller))
```

francinum:
I'm not entirely convinced this is necessary, but work that's done is done.

PR #616, opened 2022-01-06
Shipcoin  gain 2: Coin Harder

Sun-Soaked:
## About The Pull Request

-Shipcoins are now gained at a rate of 15 every 10 minutes, over the course of time in a round as a living entity.
-The Shipcoin reward on roundend has been reduced to 50 coins, or 25 if you do not survive but stick around to the end.

Provided you survive to the end of a round, you need to play for just under 3 hours to achieve the original 300 coin reward.

This is more or less a direct port of the original bee functionality, but this may change if more powerful coders walk me through making it a different subsystem.

## Why It's Good For The Game

Rewards time played rather than rounds, mostly slays the beast that is "I would probably keep playing but I want my 300 shipcoins, transfer vote"

## Changelog
:cl:
tweak: shipcoins are now gained mostly based on time played, at a rate of 1.5 per minute. The reward on round-end is now 50 coins.
/:cl:

### DIFF
--- a/code/__DEFINES/~whitesands_defines/metacoin.dm
+++ b/code/__DEFINES/~whitesands_defines/metacoin.dm
@@ -11,4 +11,4 @@
 /// Rewarded when you don't survive the round, but stick around till the end
 #define METACOIN_NOTSURVIVE_REWARD 25
 /// Rewarded when you are alive and active for 10 minutes
-#define METACOIN_TENMINUTELIVING_REWARD 10
+#define METACOIN_TENMINUTELIVING_REWARD 15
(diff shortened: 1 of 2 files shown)


Kapu1178 (changes requested):
The whole job XP system sucks and we have a handy new `GLOB.respawn_timers` list to keep track of who has been dead for how long.

I'd rather just when SSBlackbox fires, loop through `GLOB.player_list` and for every `/mob/living`, check `GLOB.respawn_timers[mob.ckey]` against `world.timeofday` using some simple math to determine if they've been dead for 10 minutes or longer. Luckily, `GLOB.respawn_timers[ckey]` spits out `0` if the target is alive.

MarkSuckerberg (approved):
Can't believe that code was literally just sitting there

PR #628, opened 2022-01-07
quick-fixes dumb mapping mistake introduced in areas & ambience

Sun-Soaked:
dumb bish accidentally coats entire fortress of solitude map in explored subtype, creates hard vacuum on rock worlds

## About The Pull Request

:mothdonk:

## Why It's Good For The Game

begone vacuum

## Changelog
:cl:
fix: fixes vacuum spawning alongside the Fortress of Solitude ruin.
/:cl:


PrefabQuasar:
theres an issue about this, that ruin isnt the only one affected, and the cause isnt mapper moments, because the bug arose on ruins that previously worked perfectly with no changes to the ruin map

Sun-Soaked:
actually in this case it is 900% my fault, I accidentally extended the explored subtype of area(which blocks generation) over the area outside the ruin, which has turf passthroughs on it.

PR #641, opened 2022-01-09
[MDB IGNORE] Removes Tiny Fans | Adds Integrated Tiny Fans

ZephyrTFA:
## About The Pull Request

Removes all occurrences of Tiny Fans.
Adds the Integrated Tiny Fan; which is a tiny fan that uses power to block atmosphere changes.

## Why It's Good For The Game

No more free atmosphere blocking; and you can actually make these create external airlocks; just dont let your power run out!

## Changelog
:cl:
add: Integrated Tiny Fans; standard in all external airlocks and poddoors.
del: Tiny Fans
/:cl:


code/game/machinery/integrated_tiny_fan.dm
```diff
@@ -0,0 +1,32 @@
...
+	name = "Integrated Air-based Atmospheric Seal"
+	desc = "A tiny fan linked to the control circuity of the airlock."
+	icon = 'icons/obj/lavaland/survival_pod.dmi'
+	icon_state = "fan_tiny"
+	CanAtmosPass = ATMOS_PASS_YES
+	power_channel = AREA_USAGE_ENVIRON
+	use_power = IDLE_POWER_USE
+	idle_power_usage = 20
+	/// Reference to our parent airlock
+	var/obj/machinery/door/airlock/parent_airlock
```
MarkSuckerberg: Is this var actually used at all? And if it is needed, you could probably pass it as an argument to Initialize() instead of using locate

MarkSuckerberg (changes requested):

code/game/machinery/integrated_tiny_fan.dm
ZephyrTFA: It was used in an earlier version that I scraped because it was needlessly excessive. I just never actually got rid of it

triplezeta:
i think prefab was working on something similar to this but i see no issues

triplezeta:
pretend that was an approving review

AtlasRules:
very good

ZephyrTFA:
merge conflict resolves; @MarkSuckerberg :eyes:

PR #665, opened 2022-01-15
Fix dependencies/requirements discrepancy

AsciiSquid:
## About The Pull Request
It seems the version of python in dependencies.sh (3.6.8) doesn't support the version of Pillow trying to be fetched by tools/requirements.txt (9.0.0) so the dependency check keeps failing.

This PR rolls back the version of Pillow to 3.4.0, the last compatible version for python 3.6.8

## Why It's Good For The Game
Allows coders to actually commit their changes.

## Changelog
:cl:
fix: fixed a conflict in library versions regarding python
/:cl:


ZephyrTFA:
This touches changelogs aswell for some reason so I just made a revert PR. This is no longer needed.

PR #671, opened 2022-01-16
Kepori Frying Code

AsciiSquid:
## About The Pull Request
Adds more functionality to mob container items to allow for additional interactions, mostly relevant to Kepori.

Mob containers will now check if they are inside a movable and call `relay_container_resist_act` if there is any special resist function. A base case has also been added to all movable atoms which just auto resolves and releases them.

To test this Kepori (and any other mob with TRAIT_HOLDABLE) can drag-drop themselves into a container if it can fit them. Breakout just has a slightly different prompt but is otherwise just as instant.

Deep fryers now work more intuitively, which is to say will burn the mob alive in boiling oil. This won't make them into a consumable item though, if there is enough demand for this I will try to return this feature.

Also contains #665 because I actually couldn't commit and push this without it.

## Why It's Good For The Game
Holdable mob functionality is super cool.

## Changelog
:cl:
add: Kepori can now store themselves, be wary around duffel bags.
add: Consequently, players can stuff kepori into bags as well.
fix: Deep fryers are no longer pet safe
/:cl:


code/modules/food_and_drinks/kitchen_machinery/deep_fryer.dm
```diff
@@ -28,6 +28,7 @@
 	idle_power_usage = 5
 	layer = BELOW_OBJ_LAYER
 	var/obj/item/reagent_containers/food/snacks/deepfryholder/frying	//What's being fried RIGHT NOW?
+	var/obj/item/clothing/head/mob_holder/frying_mob					//Who's bring fried RIGHT NOW?
```
ZephyrTFA: use the correct form of documentation please.

code/game/objects/items/storage/storage.dm
```diff
@@ -60,3 +60,27 @@
...
+			return
+		M.visible_message("<span class='notice'>[M] starts to crawl into [src]...</span>", "<span class='notice'>You start crawling into [src]...</span>")
+		if(!do_after(M, 30, TRUE, src))
+			holder.release()
+			return
+		holder.forceMove(src)
+		ST.signal_insertion_attempt(ST, holder, M)
+		to_chat(M, "<span class='notice'>You have tucked yourself away into [src].</span>")
+
+/obj/item/storage/relay_container_resist_act(mob/living/user, obj/O)
```
ZephyrTFA: Use better var names

code/game/objects/items/storage/storage.dm
```diff
@@ -60,3 +60,27 @@
 		var/custom_data = item.on_object_saved(depth++)
 		dat += "[custom_data ? ",\n[custom_data]" : ""]"
 	return dat
+
+/obj/item/storage/MouseDrop_T(mob/living/M, mob/living/user)
```
ZephyrTFA: Use better var names.

code/modules/mob/living/living.dm
```diff
@@ -1333,6 +1333,10 @@
 			return//dont open the mobs inventory if you are picking them up
 	. = ..()
 
+/mob/living/proc/mob_to_item()
```
ZephyrTFA, line 1341: document this

tools/requirements.txt
```diff
@@ -1,6 +1,6 @@
 pygit2==1.0.1
 bidict==0.13.1
-Pillow==9.0.0
+Pillow==8.4.0
```
ZephyrTFA: atomization, this is already in a different PR

ZephyrTFA (changes requested):

tools/requirements.txt
AsciiSquid: I literally cannot make commits without this change.

code/modules/food_and_drinks/kitchen_machinery/deep_fryer.dm
AsciiSquid: There isn't one so I used the same method as the above variable.

code/game/objects/items/storage/storage.dm
AsciiSquid: var names inherited, changing them would cause inconsistency

code/game/objects/items/storage/storage.dm
AsciiSquid: var names inherited, changing them would cause inconsistency

code/modules/mob/living/living.dm
AsciiSquid: What about this function needs clarification?

ZephyrTFA: https://github.com/shiptest-ss13/Shiptest/blob/master/.github/AUTODOC_GUIDE.md

code/modules/food_and_drinks/kitchen_machinery/deep_fryer.dm
ZephyrTFA: https://github.com/shiptest-ss13/Shiptest/blob/master/.github/AUTODOC_GUIDE.md

code/modules/mob/living/living.dm
AsciiSquid: This article does not contain an answer to my question. What about this function needs documentation?

ZephyrTFA (approved):
I can overlook missing a single proc for documentation as it can be added later. I'd still wish you'd update the var names but I won't require you to do so.

AsciiSquid:
> I'd still wish you'd update the var names

I'm just kind of confused what the issue is with them. They're both generic control functions so isn't it standard to keep them as single letters/generic then casted into a proper variable after a type check?

MarkSuckerberg:
I have concerns with this PR, but I'm a bit too sick to bring them up at this time. DNMing for now so the other maints don't merge it beforehand.

AsciiSquid:
Can I get a reason why?

MarkSuckerberg:
*Just* said I'm too sick to bring it up at the moment.

AsciiSquid:
Not even a hint so I could try fixing it?

MarkSuckerberg:
No because I have a feeling you're going to fight me tooth and nail on it and I already have a headache

MarkSuckerberg:
Right, so, my concerns: (apologies for the delay)

I believe that this PR adds features that will only stand to worsen a few of the issues with kepori, most significantly them being taken seriously as a race. I know that I was who started it by making them holdable in the first place, and as a result led to the joke of kepori frying/microwaving/washing machining, but I had honestly never planned for the lattermost three to exist and wanted to remove them but never got around to it. 

Additionally, being able to put three of them into a duffel and one into a backpack feels far too much imo, I remember not a week ago someone poking fun at how absurd even fitting one of them into a duffel bag was, and now we're at the opposite end with three to a duffel and one to a normal backpack, which, I believe your motive with that was to make it so it could happen more often (see: most people don't use duffel bags due to the slowdown) the explicit reason I limited them to being containable in duffels in the first place was to limit the occurrence to a rarity at most.

tl;dr: I don't think this PR would be good for Kepori as a race, some of it's good I guess and I can point that out in a little bit

P.S. I have the backing of the entire Kepori lore team on these concerns.

AsciiSquid:
There seems to be a misconception regarding my intent here.
>  I believe that this PR adds features that will only stand to worsen a few of the issues with kepori, most significantly them being taken seriously as a race.

This change does not exclusively affect kepori, but all holdable_mobs. Which have had a buggy interaction with deep fryers for a long time now in which you can generate infinite fried food by dipping a holdable mob (such as Kepori, but also Dwarfs and certain simplemobs like Ian) and removing them to take a bite and repeat to no harm to the mob. The intent of this PR is to fix this bug and turn frying holdables from a joke exploit to a serious action that actually harms them.

Additionally, this PR does not affect container limits on holdables and it's still restricted to exclusively one-per duffle. But does develop a simple interface for more consistent holdable interactions across all atoms to make the feature more robust for future use regardless of direction. I **specifically avoided making changes to the container limit on holdables because I knew it was a strong opinion of the team** so I'm confused why it's being used to hold back this PR that has nothing to do with that.

MarkSuckerberg:
Only being held back because I wanted my concerns addressed, which we are doing right here at the current time, chill, it would need to be code reviewed first anyways. I didn't realize you didn't change container limits, I have not gone over the code yet and I seem to remember you talking about changing it in this PR previously on the discord. As for the deep fryer, I do understand your opinion on that and admire your effort on fixing it, but I still believe that it should at most just be handled as a normal mob being face-dunked into the fryer. And as I said before,
> but I had honestly never planned for [kepori frying/microwaving/washing machining] to exist and wanted to remove them but never got around to it.

So basically what I'm saying here is that I would rather you go about fixing the deep frier/other stuff in a different way, but the rest of the PR **(conceptually, I have not even looked at the code yet)** is fine with me.

CMDRYouth:
I don't know how I ended up here, but anything bird-shaped belongs in the Deep Fryer, and then in a bucket.

AsciiSquid:
Unfortunately holdable mobs don't fit into reagent containers so buckets are off the list.

code/game/atoms_movable.dm
```diff
@@ -833,11 +833,21 @@
...
+/**
+  * Called when a mob resists while inside a container that is itself inside something.
+  *
+  * Arguments:
+  * * resistor - Mob calling the container_resist_act
+  * * container - containing object the mob is nested within
+  */
+/atom/movable/proc/relay_container_resist_act(mob/living/resistor, obj/container)
+	//Generic case for mob_holders
+	if(istype(container, /obj/item/clothing/head/mob_holder))
```
MarkSuckerberg, line 845: Couldn't you just make this be in /obj/item/clothing/head/mob_holder/container_resist_act()?

code/modules/food_and_drinks/kitchen_machinery/deep_fryer.dm
```diff
@@ -27,12 +27,20 @@
 	use_power = IDLE_POWER_USE
 	idle_power_usage = 5
 	layer = BELOW_OBJ_LAYER
-	var/obj/item/reagent_containers/food/snacks/deepfryholder/frying	//What's being fried RIGHT NOW?
+	/// What's being fried RIGHT NOW?
+	var/obj/item/reagent_containers/food/snacks/deepfryholder/frying
+	/// Who's bring fried RIGHT NOW?
+	var/obj/item/clothing/head/mob_holder/frying_mob
```
MarkSuckerberg, line 33: Still conflicted about this

code/modules/mob/living/living.dm
```diff
@@ -1333,16 +1333,37 @@
...
+/**
+  * Creates a mob_holder for the mob and gives it to the picker-upper.
+  *
+  * Arguments:
+  * * user - mob currently picking up this mob
+  */
+/mob/living/proc/mob_pickup(mob/living/user)
+	var/obj/item/clothing/head/mob_holder/holder = mob_to_item()
+	user.visible_message("<span class='warning'>[user] scoops up [src]!</span>")
+	user.put_in_hands(holder)
```
MarkSuckerberg, line 1354: Not entirely sure what reason there is for this to be broken up into two procs

MarkSuckerberg (changes requested):

Zevotech:
kentucky fried kepi... :thepleadster:

code/modules/mob/living/living.dm
AsciiSquid: Only needed that snippet of the code to implement into the storage obj code. Opted to make it a proc over copy+paste as any future atoms creating mob holders would also need that snippet.

code/game/atoms_movable.dm
AsciiSquid: This actually is in essence what that proc does currently....

However, container_resist_act doesn't proc recursively and the mob_holder is considered its own atom. So if the atom needs to make any checks for a container_resist_act on a mob, the holder has to relay it, which is what this is for.

code/modules/food_and_drinks/kitchen_machinery/deep_fryer.dm
AsciiSquid: Have you seen the current state of this?
https://user-images.githubusercontent.com/39249137/152250368-80910e83-f94c-49b7-bc10-7ec66dc285ff.mp4

MarkSuckerberg: I more meant that I was conflicted about them being truly fryable *at all,* not that it's a regression or anything.

AsciiSquid: I'm generally of the disposition that it's better to change the outcome of an option than removing the option. The problem with deepfrying held mobs isn't the act itself, but the fact that it has silly behaviour and is of no real consequence to both parties.

You mentioned an issue with the current state of kepori:
> ...most significantly them being taken seriously as a race.

This PR was made to resolve that make actually taking the idea seriously and making the actions have consequence to consider.

PrefabQuasar: i think as long as the frying is mechanically the same as the normal neck-grab fry, it should be fine, since it just makes an unintended possibility work in line with similar actions

triplezeta:
this one is literally testmerged you wh***

PR #681, opened 2022-01-19
Renormalize all files to Unix EOL

ZephyrTFA:
git literally has functions to normalize line endings for the entire work tree
```git
git rm --cached -rf .
git diff --cached --name-only -z | xargs -n 50 -0 git add -f
```


ZephyrTFA:
it appears to also normalize the files stat to the folder default, which means non-executable, pain

PR #693, opened 2022-01-20
Kepori jump good

AsciiSquid:
## About The Pull Request
Gives Kepori the innate ability to tackle. Borrowed from the tackling gloves effect.
Additional tweaks were made to the tackling component, such as better defined target damage and conditional support for Kepori. One such addition being the ability for kepori to cleanly jump over tables (given there isn't too much stuff on them).

The base stats for kepori tackling has been added to their species file, and will likely need tweaking for balance. As of posting the current spread is:

- Stamina Cost: 10
- Base Knockdown: 0.2 Seconds
- Range: 8
- Speed: 2
- Skill Mod (Affects chance of successful tackle on hit): 2 

Tackle ingame by attempting to throw with an empty hand.

## Why It's Good For The Game
More mobility for the mobility focused species. Made balanced by how easy it is to break your fragile bird skull by accidentally jumping head-first into a wall.

## Changelog
:cl:
add: Kepori now know how to jump.
tweak: Tackle/Pouncing now more accurately induces head trauma.
/:cl:


triplezeta (approved):
I suggested it, it's only right I show my approval.

TallDeekus:
One the one hand, I have seen tackling and it's effects on other races before and found them too potent...

However, for Kepori it might actually work out just fine, plus it doesn't look too powerful as of now... Only time will tell, I approve.

PrefabQuasar:
> One the one hand, I have seen tackling and it's effects on other races before and found them too potent...
> 
> However, for Kepori it might actually work out just fine, plus it doesn't look too powerful as of now... Only time will tell, I approve.

while tackling is pretty powerful, literally anyone can access it, as improvised grippers are _really_ easy to make

TallDeekus:
> 

Yea I was more talking about my trauma involving Vox Tackles from Urist, and all the times I abused it >:)

code/modules/mob/living/carbon/human/species_types/kepori.dm
```diff
@@ -47,3 +61,11 @@
 			return FALSE
 		return equip_delay_self_check(I, H, bypass_equip_delay_self)
 	. = ..()
+
+/datum/species/kepori/on_species_gain(mob/living/carbon/C, datum/species/old_species, pref_load)
+	tackler = C.AddComponent(/datum/component/tackler, stamina_cost= tackle_stam_cost, base_knockdown= base_knockdown, range= tackle_range, speed= tackle_speed, skill_mod= skill_mod, min_distance= min_distance)
+	. = ..()
```
Kapu1178:
```suggestion
    ..()
	tackler = C.AddComponent(/datum/component/tackler, stamina_cost= tackle_stam_cost, base_knockdown= base_knockdown, range= tackle_range, speed= tackle_speed, skill_mod= skill_mod, min_distance= min_distance)

```

Kapu1178 (changes requested):
Im preeeeetty sure all child procs of on_species_gain should do their shit *after* the parent proc is called.

code/datums/components/tackle.dm
```diff
@@ -476,6 +478,11 @@
 			if(messes.len >= MAX_TABLE_MESSES)
 				break
 
+	// Kepori have mastered the art of the table-tackle and will instead fly over gracefully (most of the time)
+	if(iskepori(owner))
+		// However if there is too much junk on the table, it can't be helped
+		if(rand(1, MAX_TABLE_MESSES) > messes.len)
+			return
```
Kapu1178: if possible i'd like this to be moved to something like `datum/species/proc/tackle_act()`

code/modules/mob/living/carbon/human/species_types/kepori.dm
```diff
@@ -31,6 +31,20 @@
...
+	/// See: [/datum/component/tackler/var/base_knockdown]
+	var/base_knockdown = 0.2 SECONDS
+	/// See: [/datum/component/tackler/var/range]
+	var/tackle_range = 8
+	/// See: [/datum/component/tackler/var/min_distance]
+	var/min_distance = 1
+	/// See: [/datum/component/tackler/var/speed]
+	var/tackle_speed = 2
+	/// See: [/datum/component/tackler/var/skill_mod]
+	var/skill_mod = 2
```
Kapu1178: aghhh this is nasty. But it's.. probably fine?

Kapu1178 (changes requested):

code/datums/components/tackle.dm
AsciiSquid: Wasn't sure if tackling warranted dedicated procs on the species datum but that's certainly something I can do.

code/modules/mob/living/carbon/human/species_types/kepori.dm
AsciiSquid: I mean, the only thing that *has* to be there is the component var so I can just make the rest defined constants if that's preferable.

PrefabQuasar:
now that I think of it, It would probably be good if the tackling ability was toggleable with an action button, clicking on something on throw intent with an empty hand happens pretty often on accident, and it being integrated to a race would probably cause a lot of accidental hyperspace incidents.

AsciiSquid:
True, however this can reasonably be considered a skill issue. Careless clicking isn't a good habit to have in this game after all.

Also having to move your mouse all the way to the top left before moving it back to your target would have a negative impact on it's use in time-critical situations like combat, which tackling is somewhat made for.

code/modules/mob/living/carbon/human/species_types/kepori.dm
```diff
@@ -31,6 +31,20 @@
 	no_equip = list(ITEM_SLOT_BACK)
 	mutanttongue = /obj/item/organ/tongue/kepori
 	species_language_holder = /datum/language_holder/kepori
+	/// For storing our tackler datum so we can remove it after
+	var/datum/component/tackler
```
Fikou: we usually dont want to store components on anything, weakrefs would be better but still eh, you could also just qdel(C.GetComponent(/datum/component/tackler))

code/datums/components/tackle.dm
```diff
@@ -156,9 +156,10 @@
 	var/mob/living/carbon/target = hit
 	var/mob/living/carbon/human/T = target
 	var/mob/living/carbon/human/S = user
-	var/tackle_word = isfelinid(user) ? "pounce" : "tackle" ///If cat, "pounce" instead of "tackle".
 
 	var/roll = rollTackle(target)
+	var/tackle_word = isfelinid(user) || iskepori(user) ? "pounce" : "tackle" ///If cat, "pounce" instead of "tackle".
```
Fikou: parentheses here (in the OR statement) would make it less confusing, also the comment could use an update, same with the leap_word change above.

Kapu1178 (approved):

PR #696, opened 2022-01-21
[Fix] Couch Rotation

untrius:
## About The Pull Request

Couches currently do not rotate correctly.  This can be seen whenever a ship lands and takes off.

### Original versions
Constructed in space:
image

Orientation after landing:
image

### Causes
Incorrect handedness on upward-facing sofas:
image

Incorrect directional sprites on sofa corners:
image

### Corrected versions
Constructed in space:
image

Orientation after landing:
image

### Known Issue
The corner sprites do not cover the player as expected.

All of the couches' upward-facing sprites will cover the player sprite when buckled.  This is correct for the left, right, and middle sofa sprites, but incorrect for the corners.

image
image
image
image

The fix for this would want to allow the player sprite to cover only part of the sofa sprite, and would effect 2 directions instead of just up.

I am choosing not to fix this at this time, as it's out of my depth.

## Why It's Good For The Game

You no longer have to rebuild couches whenever you land and take off, just to make them look right.

## Changelog
:cl:
fix: Couches rotate correctly
/:cl:


MarkSuckerberg (approved):
Finally, we can have couches that actually work

PR #721, opened 2022-01-24
Curtain density change

Dimasw99:
## About The Pull Request

Curtains always remain density 0 now.

## Why It's Good For The Game

Material doors already fill the niche of unpowered doors and curtains would fit the perfect role of blocking vision while being walkthroughable. Likewise it makes little sense that just a few lose pieces of fabric would stop you 

## Changelog
:cl:
tweak: Curtains always remain density 0 now.
/:cl:


MarkSuckerberg (approved):
Good idea tbh, it gives them a use in the scheme of things rather than "weak door alternative"

PR #737, opened 2022-01-29
Fixes some bullet-related bugs.

### LINKED ISSUE #343: [BUG] Mechs cannot be hit by any projectiles
## Reproduction:
Spawn mech
Spawn gun
Shoot at mech with gun
Projectile then proceeds to go around the mech, easier to see firing directly straight at the mech

benbot16:
## About The Pull Request

So when bullets try to hit objects, they try to pick a target from any structures or machinery on the tile they're hitting.

The problem with this is that mechs aren't machines or structures. When a bullet tries to hit them, it'll just ignore the mech and keep bumping up against it until the bullet times out.

This PR fixes that by adding mechs to the list of objects that can get hit by a bullet.

I fixed spacepods the same way.

Fixing X-ray lasers was done by adding pass flag handling to turfs, since it apparently didn't have that before despite the fact that closed turf pass flags exist. 

closes #343

## Why It's Good For The Game

Combat mechs shouldn't be immune to bullets.

## Changelog
:cl:
fix: Mechs can get hit by bullets again.
fix: Spacepods can get hit by bullets again.
fix: X-ray lasers can go through walls now.
/:cl:

### DIFF
--- a/code/_globalvars/lists/typecache.dm
+++ b/code/_globalvars/lists/typecache.dm
@@ -9,4 +9,4 @@ GLOBAL_LIST_INIT(typecache_living, typecacheof(/mob/living))
 
 GLOBAL_LIST_INIT(typecache_stack, typecacheof(/obj/item/stack))
 
-GLOBAL_LIST_INIT(typecache_machine_or_structure, typecacheof(list(/obj/machinery, /obj/structure)))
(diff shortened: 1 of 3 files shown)


PrefabQuasar:
spacepods also have this issue, as do walls when hit by x-ray projectiles

benbot16:
> spacepods also have this issue, as do walls when hit by x-ray projectiles

I'll look at that too then.

PR #752, opened 2022-02-01
I'm deleting everything ss13 related from my computer

Pockets-byte:
this is here merely to save it from oblivion
if it breaks so be it, I can't be assed to have anything to do with it anymore


Pockets-byte:
if this breaks close it, I'm literally never coming back to this

ZephyrTFA:
I'm sorry you feel this way.
I, however, will not allowing maps that will not be maintained onto the codebase.
I hope you happiness in your future endeavors and please stay safe.

PR #764, opened 2022-02-04
Minor Map-export Fixes

MarkSuckerberg:
## About The Pull Request
Fixes up map export code a small amount, see changelog for fixes

## Why It's Good For The Game
Feature maintenance

## Changelog
:cl:
fix: Map exporting now no longer crops off the bottom row of tiles
tweak: Map exporting now saves all items a carbon/human has and dumps them on the floor where they were.
tweak: Less overzealous CHECK_TICKing of the map exporter to make it not slow as heck- benchmarking to see if this is worth it or not incomplete
/:cl:


ZephyrTFA:
What's the goal of moving the CHECK_TICK calls? Just to get it to finish faster?

MarkSuckerberg:
Somewhat, previously it checked if was overtiming a lot more than it should have (at least, it's my belief that it did so) and as such very likely ended up sleeping a lot more than needed, at the expense of taking a lot longer to export than it needed to. 

I tested it by exporting an entire z-level and it had no noticeable impact on performance outside of the proc, which, the proc did seem to run faster, but I didn't run a timer or anything to check for certain.

Honestly I just have it in to be tested on an actual round with a TM so I can see if it lags the server or not still.

MarkSuckerberg:
Readded most CHECK_TICKs, I reflected on it and realized that in the end it doesn't really matter how slow the writer is in the end, as it's not used for anything but adminbus currently. Keeping the game running smoothly for everything outside of the exporter is the important part.

I did keep the CHECK_TICKs I believed were obviously unneeded removed, though. They're just looping through some text more or less. (Tested, obviously.)

PR #771, opened 2022-02-05
Ports Austation's port of NSV's Nuclear Reactor, plus some related gases.

benbot16:
## About The Pull Request

Ports the AGCNR, constricted plasma, and nucleium from NSV. This also ports a bunch of AGCNR-related PRs made on Austation.
This includes parts of BeeStation/NSV13#828, austation/austation#2375, austation/austation#2393, austation/austation#2401, austation/austation#2402, austation/austation#2404, austation/austation#2416, austation/austation#2485, and probably some others that I didn't notice.

The reactor itself produces power through a moderator, which is piped in at the top of the reactor. Coolant also has to be pumped through the reactor, which heats up in the reactor based on how high the reactor's power level is. The reactor also produces radiation.
The reactor can be destroyed if it overheats too much, causing a meltdown, or if the pressure gets too high, causing a blowout.
The reactor and fuel rods are obtainable through Cargo.

Constricted plasma is made using plasma constrictors, which can be researched down the industrial engineering tree. This gas is similar to plasma, but it burns hotter and has slightly different byproducts.

Nucleium's produced by a fueled nuclear reactor, and released into the air when a reactor over-pressurizes. It's radioactive when inhaled, and has a very high heat capacity.

I've tested pretty much everything, and it *should* work fine. The fallout event after causing a blowout will probably need to get looked at, though, since it was acting kind of wonky during testing.

## Why It's Good For The Game

This adds another reactor that can be used on ships, as well as some new gases that have special interactions with it. 

## Changelog
:cl:
add: Ports the AGCNR from NSV/Austation
add: Ports Constricted Plasma from NSV13
add: Ports Nucleium from NSV13
/:cl:


rye-rice:
glad to see the screen states i made finally getting used

francinum:
...This engine is cursed and should never have seen the light of day.

put that thing back where it came from, or so help me

francinum:
To be more formal:

* Plasma constriction is arbitrary and does not have a genuine function outside of the AGCR
* Nucleum is strictly a death gas with no productive use outside of that fact
* Plasma constrictor code sucks ass
* This engine is far more static than any other engine in the game, and in a codebase designed so heavily around the dynamic aspects of the codebase, it doesn't feel all that fitting, as it does on NSV.

benbot16:
> To be more formal:
> 
>     * Plasma constriction is arbitrary and does not have a genuine function outside of the AGCR
> 
>     * Nucleum is strictly a death gas with no productive use outside of that fact
> 
>     * Plasma constrictor code sucks ass
> 
>     * This engine is far more static than any other engine in the game, and in a codebase designed so heavily around the dynamic aspects of the codebase, it doesn't feel all that fitting, as it does on NSV.

I can understand why the gases aren't great, since they don't really add anything unless you're an atmos tech trying to do a few different things, so they're probably not a great fit here.
However, I don't really see how the reactor itself is more static than the other large engines we have here. You can't move a reactor, sure, but you also can't really do that with an SM or antimatter reactor, and it's barely ever done with TEG engines. You can also just turn it off and modify the setup if you really want to.

Kapu1178:
The maintainers have decided that we do not wish to pursue the addition NSV's Nuclear Engine. In addition to the issues outlined by Francinum, there are concerns over its actual gameplay being incredibly linear, as well as the thing being essentially a time bomb.

PR #803, opened 2022-02-11
Hopefully fixes 3 harddels, fixes overlay lighting runtimes

Kapu1178:
## About The Pull Request
DCM, Ore Veins, and Firebots should hopefully properly garbage collect.

## Why It's Good For The Game
Hard dels bad.

## Changelog
:cl:
code: Harddel fixes
fix: riding component causing thousands of runtimes due to a bad signal
/:cl:

### DIFF
--- a/whitesands/code/modules/mining/deepcore/ore_vein.dm
+++ b/whitesands/code/modules/mining/deepcore/ore_vein.dm
@@ -4,6 +4,7 @@ GLOBAL_LIST_EMPTY(ore_vein_landmarks)
 	name = "ore vein"
 	var/datum/material/resource
 	var/material_rate = 0
+	var/obj/machinery/deepcore/miner
 
 /obj/effect/landmark/ore_vein/Initialize(mapload, var/datum/material/mat)
 	. = ..()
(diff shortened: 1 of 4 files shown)


SuperNovaa41:
Bots will still harddel after this, they still have refs in data huds
image

whitesands/code/modules/mining/deepcore/ore_vein.dm
```diff
@@ -4,6 +4,7 @@ GLOBAL_LIST_EMPTY(ore_vein_landmarks)
 	name = "ore vein"
 	var/datum/material/resource
 	var/material_rate = 0
+	var/obj/machinery/deepcore/miner
```
SuperNovaa41:
```suggestion
	var/obj/machinery/deepcore/drill/miner
```

francinum:
> Bots will still harddel after this, they still have refs in data huds image

...Is it not getting cleaned up by `remove_from_all_data_huds()` at `living.dm:34?`

SuperNovaa41:
> > Bots will still harddel after this, they still have refs in data huds image
> 
> ...Is it not getting cleaned up by `remove_from_all_data_huds()` at `living.dm:34?`

No, and i'm not sure why. `/mob/dead/observer` is also suffering from the same sort of harddels (this is on voidcrew but i havent seen any fixes related to this so i'm assuming you guys also suffer from them)

francinum:
> 

Can't replicate locally? Unless something is massively fucked with huds that I miss, but that doesn't seem to be the case.

SuperNovaa41:
> > 
> 
> Can't replicate locally? Unless something is massively fucked with huds that I miss, but that doesn't seem to be the case. 

If you turn on ref tracking they pop up sporadically, i'm not sure why only some of them aren't properly clearing

PR #814, opened 2022-02-12
Fixes reflector rotation by shuttles

### LINKED ISSUE #710: [BUG] Reflectors don't rotate when a ship is launched.
## Round ID:
[629](https://shiptest.ga/stats/rounds/629)

## Testmerges:
#230 
#652 
#607 
#530 

## Reproduction:
Put a reflector on a ship, then launch the ship. Notice the reflectors don't rotate, despite the ship rotating when it launches.

benbot16:
## About The Pull Request

This PR fixes reflectors not rotating when a shuttle takes off/lands. 

## Why It's Good For The Game

Having to re-rotate the reflectors in an SM setup every time the ship decides to move is annoying.

closes #710 

## Changelog
:cl:
fix: Reflectors on shuttles rotate properly now.
/:cl:

### DIFF
--- a/code/modules/shuttle/shuttle_rotate.dm
+++ b/code/modules/shuttle/shuttle_rotate.dm
@@ -82,7 +82,7 @@ If ever any of these procs are useful for non-shuttles, rename it to proc/rotate
 
 /obj/structure/reflector/shuttleRotate(rotation, params)
 	. = ..()
-	setAngle(rotation_angle += rotation)
+	setAngle(SIMPLIFY_DEGREES(rotation_angle += rotation))
 
 /************************************Machine rotate procs************************************/
 
--- a/code/modules/shuttle/shuttle_rotate.dm
+++ b/code/modules/shuttle/shuttle_rotate.dm
@@ -80,6 +80,10 @@ If ever any of these procs are useful for non-shuttles, rename it to proc/rotate
 	params &= ~ROTATE_OFFSET
 	return ..()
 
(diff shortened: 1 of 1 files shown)


code/modules/shuttle/shuttle_rotate.dm
```diff
@@ -80,6 +80,10 @@ If ever any of these procs are useful for non-shuttles, rename it to proc/rotate
 	params &= ~ROTATE_OFFSET
 	return ..()
 
+/obj/structure/reflector/shuttleRotate(rotation, params)
+	. = ..()
+	setAngle(SIMPLIFY_DEGREES(rotation_angle += rotation))
```
MarkSuckerberg:
```suggestion
	setAngle(SIMPLIFY_DEGREES(rotation_angle + rotation))
``` 
might only need to be this? I could be wrong

benbot16: setAngle already sets rotation_angle, and I need SIMPLIFY_DEGREES to check for degrees greater than 360. Rotation_angle + rotation is just getting the new angle we want to set.

MarkSuckerberg: Yep, in that case you'll only need + since += adds and then sets the var to it.

benbot16: Thought those two were flipped, evidently I just can't read.

PR #825, opened 2022-02-14
Makes the ak47 and NT-AK not fit on the belt slot

rye-rice:
## About The Pull Request

Makes the ak47 and subtypes fit on the back instead of the belt slot

## Why It's Good For The Game

I fucking missed this and never fixed it until now 

## Changelog
:cl:
fix: ak47 no longer fits in the belt slot, but fits on backs like they are supposed to
/:cl:


Latentish:
They could fit WHERE?!
Yeah probably a good idea.

PR #873, opened 2022-02-24
Removes Quantum Pads from technodes, and two ruins.

TheNeoGamer42:
## About The Pull Request
Removes quantum pads from the research node which contains them, effectively making them unobtainable outside of adminspawn. Also removes them from two ruins, even moreso making them adminspawn only. Why a rockplanet ruin had three quantumpads lined up next to eachother is byond me, but I shot them now.

## Why It's Good For The Game
Quantum Pads currently completely invalidate the idea of ships being far away from eachother, as they can simply slap these little buggers up, and teleport to their buddies ship, even if they are as far away as possible. While this would in theory just result in very easy trading, it just results in figuratively bolting two ships together and zero trading.

## Changelog
:cl:
del: Removed Quantum Pads from research nodes. They are now admin only.
del: Removed Quantum Pads from two ruins, there is no longer three free quantum pads in a certain rockplanet ruin.
/:cl:

### DIFF
--- a/_maps/RandomRuins/RockRuins/rockplanet_clock.dmm
+++ b/_maps/RandomRuins/RockRuins/rockplanet_clock.dmm
@@ -37,10 +37,6 @@
 	},
 /turf/open/floor/bronze,
 /area/overmap_encounter/planetoid/rockplanet/explored)
-"hb" = (
-/obj/machinery/quantumpad,
-/turf/open/floor/bronze,
-/area/overmap_encounter/planetoid/rockplanet/explored)
 "hm" = (
 /obj/effect/decal/cleanable/blood/old,
 /obj/effect/decal/cleanable/dirt/dust,
@@ -125,11 +121,6 @@
 /obj/structure/trap/stun,
 /turf/open/floor/bronze,
 /area/overmap_encounter/planetoid/rockplanet/explored)
-"tg" = (
-/obj/structure/fluff/clockwork/alloy_shards,
-/obj/machinery/quantumpad,
-/turf/open/floor/bronze,
-/area/overmap_encounter/planetoid/rockplanet/explored)
 "tI" = (
 /obj/effect/decal/cleanable/greenglow,
 /obj/effect/decal/cleanable/dirt/dust,
@@ -140,7 +131,6 @@
 /turf/open/floor/bronze,
 /area/overmap_encounter/planetoid/rockplanet/explored)
 "ur" = (
-/obj/machinery/quantumpad,
 /obj/structure/trap/damage,
 /turf/open/floor/bronze,
 /area/overmap_encounter/planetoid/rockplanet/explored)
@@ -533,7 +523,7 @@ Nr
 "}
 (11,1,1) = {"
 Qs
-tg
+lm
 Id
 Ji
 NP
@@ -567,7 +557,7 @@ OB
 "}
 (13,1,1) = {"
 Qs
-hb
+NP
 NP
 jK
 WK
--- a/_maps/RandomRuins/SandRuins/whitesands_surface_medipen_plant.dmm
+++ b/_maps/RandomRuins/SandRuins/whitesands_surface_medipen_plant.dmm
@@ -467,7 +467,6 @@
 /area/whitesands/surface/outdoors)
 "lM" = (
 /obj/effect/turf_decal/industrial/outline/yellow,
-/obj/machinery/quantumpad,
 /turf/open/floor/engine,
 /area/whitesands/surface/outdoors)
 "lO" = (
--- a/code/modules/research/techweb/all_nodes.dm
+++ b/code/modules/research/techweb/all_nodes.dm
@@ -213,7 +213,7 @@
 	display_name = "Bluespace Travel"
 	description = "Application of Bluespace for static teleportation technology."
 	prereq_ids = list("practical_bluespace")
-	design_ids = list("tele_station", "tele_hub", "teleconsole", "quantumpad", "launchpad", "launchpad_console", "bluespace_pod")
+	design_ids = list("tele_station", "tele_hub", "teleconsole", "launchpad", "launchpad_console", "bluespace_pod")
 	research_costs = list(TECHWEB_POINT_TYPE_GENERIC = 5000)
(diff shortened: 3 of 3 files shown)


Dimasw99:
Trading as in its current form consists of basically people coming over and asking for some essential basics like RnD tech, mining stuff and then move on. From my experience ship to ship interactions simply happen too rarely as it requires quite an amount of coordination and time investment so you can maybe get like 3 tops if you are actively seeking it out and don't forget that you also impacting everyone elses experience with it as each transition is just going to be bar rp for everyone. 

Quantum pads solve exactly that problem where you simply just hop over to another ship for various reasons and not only for trading. Even then quantum pads are not a guarantee to happen each round.

Kurgis:
Not sure how ships like the Boyardee or Botanical-class are going to cope with this change, usually they can hopefully chase down a ship to get a connection with, but without that others will just avoid them like a plague in search of simply more omnilathe. With this it's just turning it into a game of "stop touching the quantum inverter pile, label them people!", and "drop the inverter after using it!". This instead splits interaction from these ships to individual people who will place inverters in certain areas of the ship. As an anecdote I had made contact with a ship again, and they wanted to talk about connecting our ships together with a quantum pad, unaware that I already made contact with them previously some hours ago. I've been using their facilities without this crewmember even being aware because I've had a inverter onboard. I'd remove quantum inverters as well as it'd make the issue you're attempting to solve even worse, but personally I'm not finding an issue at all.

TheNeoGamer42:
> Not sure how ships like the Boyardee or Botanical-class are going to cope with this change, usually they can hopefully chase down a ship to get a connection with, but without that others will just avoid them like a plague in search of simply more omnilathe. With this it's just turning it into a game of "stop touching the quantum inverter pile, label them people!", and "drop the inverter after using it!". This instead splits interaction from these ships to individual people who will place inverters in certain areas of the ship. As an anecdote I had made contact with a ship again, and they wanted to talk about connecting our ships together with a quantum pad, unaware that I already made contact with them previously some hours ago. I've been using their facilities without this crewmember even being aware because I've had a inverter onboard. I'd remove quantum inverters as well as it'd make the issue you're attempting to solve even worse, but personally I'm not finding an issue at all.

Removing inverters does seem like a good idea, given they are rather similar to this. As for the issues of ships like the Boyardee being avoided in search of omnilathes, well, I think this is simply an issue that will persist no matter what until the plans for reworking RND are finalized and executed, and this is mostly meant to try to prevent the balance from being complete ass by virtue of essentially ignoring the overmap and such

Ms-Mee:
The raging powercreep and obsession with RnD screwing ships like rhe Boyardee over is not something solved by having or not having teleportation, I'd say they're very different matters

TheNeoGamer42:
> The raging powercreep and obsession with RnD screwing ships like rhe Boyardee over is not something solved by having or not having teleportation, I'd say they're very different matters

It's not solved by removing teleportation, but teleportation still doesn't work well with the intentions from shiptest really from what I can tell, so it will still help in that area.

Ms-Mee:
Indeed.

ZephyrTFA:
If you want to make it so qpads aren't as strong you can make them cost resources to use. People would be more conservative with them if they cost say five bluespace crystals

TallDeekus:
> If you want to make it so qpads aren't as strong you can make them cost resources to use. People would be more conservative with them if they cost say five bluespace crystals

Yes, I am generally more up for the idea of nerfing them instead of outright removing them, plus they are a later game tech advancement that players can aim for, and if their position in the techweb is a problem we can just push then further up the tree and give them their own thing, maybe even lock it behind something in a vain similar to illegal tech? Honestly I feel like we should have more lategame stuff that our players want to aim for, although it does need to actually be a challenge to get it I agree.

Kurgis:
> > If you want to make it so qpads aren't as strong you can make them cost resources to use. People would be more conservative with them if they cost say five bluespace crystals
> 
> Yes, I am generally more up for the idea of nerfing them instead of outright removing them, plus they are a later game tech advancement that players can aim for, and if their position in the techweb is a problem we can just push then further up the tree and give them their own thing, maybe even lock it behind something in a vain similar to illegal tech? Honestly I feel like we should have more lategame stuff that our players want to aim for, although it does need to actually be a challenge to get it I agree.

This just puts a longer timer on reaching the same "issues", if it's the convenience of the quantum pad then perhaps it's just best to kill it completely by limiting it to same z-level. Though this only prevents convenience, players will set-up a Teleporter Network instead, same result with extra steps, just less convenience. Ultimately, more fundamentally, that's what the teleportation of science fiction is, it's convenience, and as with most things convenient players grow very accustomed to it.

Spyroshark:
This seems like a bad change, since the only other way to meet with other ships is to fling yourself across space with whatever you can carry, or try to meet up on a (usually megafauna filled) planet. IMO it would be better to merge this when we actually have a decent way to meet up with ships.

TheNeoGamer42:
> If you want to make it so qpads aren't as strong you can make them cost resources to use. People would be more conservative with them if they cost say five bluespace crystals

Late response, but I don't think that will help really, resources are basically free and infinite. I don't think teleportation really has that much of a place in shiptest outside of being a reward for actually doing something impressive and dangerous, like raiding a rather hard ruin. This PR infact still allows for that, I just removed it from the two ruins they do naturally appear in, because both are EXTREMELY easy to get, one not even having hostile mobs of any kind, and the other having three past some easy to kill simplemobs.

TheNeoGamer42:
> This seems like a bad change, since the only other way to meet with other ships is to fling yourself across space with whatever you can carry, or try to meet up on a (usually megafauna filled) planet. IMO it would be better to merge this when we actually have a decent way to meet up with ships.

You have wideband, and holopads to communicate with others. You can also dock in empty space. I even think the overmap rework PR mark is making, allows for up to FOUR ships to dock in an empty space tile. While a better method for communication may be nice, I think quantum pads and such have really just an unintended effect of basically making any two ships linked one ship, and doesn't REALLY help with trading or whatnot, infact as it is, nobody trades much of anything, other than asking for an RND setup and moving on.

Sylphily:
pain

PR #877, opened 2022-02-25
Kicking around the runtime viewer, again

MarkSuckerberg:
## About The Pull Request
Adds a few `CHECK_TICK`s to the runtime viewer so that it doesn't lock up the entire bloody game. Also removes the 1000 runtime limit since I think that caused us to lose some important ones. *Also* reformats the file a bit since I'm sick of looking at `if (condition)`

## Why It's Good For The Game
Doesn't lag the game for everyone to the point of it hanging, but also allows coders to actually view all runtimes. Hopefully

## Changelog
:cl:
admin: The runtime viewer shouldn't hang the server for like 30 seconds anymore.
/:cl:

### DIFF
--- a/code/modules/error_handler/error_viewer.dm
+++ b/code/modules/error_handler/error_viewer.dm
@@ -51,7 +51,7 @@ GLOBAL_DATUM(error_cache, /datum/error_viewer/error_cache)
 
 	. = ""
 
-	if (istype(back_to))
+	if(istype(back_to))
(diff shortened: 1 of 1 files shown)


MarkSuckerberg:
Works perfectly, could open a runtime group with 2000+ runtimes with no noticeable lag for anyone, myself included. It takes a while to open it, but that's peanuts compared to it effecting literally everyone on the server.

PR #895, opened 2022-03-05
Adds an open air unit test

Putnam3145:
## About The Pull Request

I added it to citadel and it's not working there but neither's the reactions unit test and that one's here and it's working so I'm going to see if the open air unit test works here and if it does I'm going to try to figure out what y'all are doing that we're not.

## Why It's Good For The Game

I mean, it's a useful unit test. Probably. Assuming it works. This is a bit of a DNM if I can't finagle it, but that should be obvious.

## Changelog
:cl:
code: Added an "open air" unit test that tests open air gas movement and reactions (both uses the same system)
/:cl:


ZephyrTFA:
This codebase uses Auxmos. This unit test will only not work but we are also not a playground for your code.

MarkSuckerberg:
You're fine, this was a bit of a miscommunication/misunderstanding

code/modules/unit_tests/open_air.dm
```diff
@@ -0,0 +1,28 @@
+/datum/unit_test/open_air/Run()
+	var/datum/virtual_level/vlevel = mapzone.virtual_levels[1]
+	for(var/turf/open/T in vlevel.get_block())
+		T.Initalize_Atmos(0)
+		T.air.clear()
+	var/turf/open/center_turf = vlevel.get_center()
+	center_turf.air.set_moles(GAS_PLASMA, 32)
+	for(var/i in 1 to 10)
+		SSair.fire()
```
MarkSuckerberg, line 9: Might need a sleep or tick check after each call, so that it doesn't get skipped at all, and possibly calling SSair.process_turfs() directly?

Azarak:
`SSair.fire()` Can early return due to tick checks, and it will early return no matter how many times you'll call it if it's called past the threshold. Any of the 10 tasks SSair does could easily push it past it.

Putnam3145:
Right, dumb of me, ha.

code/modules/unit_tests/open_air.dm
```diff
@@ -0,0 +1,29 @@
...
+		sleep()
```
ZephyrTFA: I don't believe calling sleep will resolve issues where ssAir will early return itself, I might be incorrect in that assumption however

Putnam3145: I might just have a wrong idea of how the internal scheduler works, yeah--sleep(1) it is then

Putnam3145:
Aaand it works. Well, that's a thought.

PR #930, opened 2022-03-23
Fixes AIs not being able to use mech actions while in mechs

benbot16:
## About The Pull Request

This fixes a bug where AIs can't use mech actions while they're in mechs, leading to them getting stuck.

The reason this bug is happening is that AIs have the TRAIT_HANDS_BLOCKED trait (presumably due to their notable lack of hands), and mech actions check for this trait while deciding whether someone can use them.

The fix I made for this just removes the trait from AIs when they go into a mech, and gives it back to them when they leave said mech.

## Why It's Good For The Game

Getting stuck because you wanted to drive a ripley around isn't great.

## Changelog
:cl:
fix: AIs can use mech actions now.
/:cl:

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -808,6 +808,7 @@
 	AI.cancel_camera()
 	AI.controlled_mech = src
 	AI.remote_control = src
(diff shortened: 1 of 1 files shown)


benbot16:
Atmos why

ZephyrTFA:
Guess it's time to take atmos out back again

PR #938, opened 2022-03-25
Fixes a few minor bugs

MarkSuckerberg:
## About The Pull Request
Fixes: #932
Fixes: #931

## Why It's Good For The Game
Bugs bad

## Changelog
:cl:
fix: Ships are no longer blocked from moving until the planet fully deloads
/:cl:

PR #969, opened 2022-04-11
Planet Loading Changes

MarkSuckerberg:
## About The Pull Request
Makes it so that planets are loaded either at roundstart during initialisation, or right after they're re-randomised after all ships leave it. 

Also adds a `QUICK_INIT` define so that devs can continue to have quick init times. This replaces `LOWMEMORYMODE`, and, while it only is used for this one thing, should be used much more often in order to keep init times down for as quick a compile-to-game turnaround time as possible for devs.

Might make `QUICK_INIT` default? Would make it so that beginner devs benefit from it too.

**REQUIRES #938 SO THAT UNDOCKING DOESN'T TAKE AGES**

## Why It's Good For The Game
This allows for two things- one, it takes so much less time for people to dock with planets, and two, it allows for ghost roles to actually gain a foothold on planets.

## Changelog
:cl:
tweak: Planet loading now occurs during init and after departure, so initial docking should be a lot faster, as well as ghost roles will actually appear in time for them to get a foothold on the planet.
/:cl:


Zevotech:
oh my GOD YES IVE BEEN WANTING FOR SOMETHING LIKE THIS

good luck catching me on anyones ship anymore im going to become a ghostrole/silicon main

PR #982, opened 2022-04-18
Schmiedeberg-class Pharmacology Ship

TypeVar:
image
## About The Pull Request

Adds a ghetto chemistry/hydroponic focused ship.

## Why It's Good For The Game

Chem Dispensers and Chem Masters are not as engaging to use to make chems. This ship is primarily focused on making drugs (hence it being a pharmacology ship). Similar to the Geneva-class, this ship encourages the use of hydroponics and chemistry to make chems, but even more so, as this ship does not come with a botanical chem dispenser. This will create a dependence on chemistry and botany, as chemistry is able to create fertilizers for the botanist, and the botany is able to produce certain core chems (like sugar). The bartender also does not have dispensers, and will have to manually mix, create and ferment drinks, which to me seems way more engaging.
The ship also has a comparably small crew size compared to other ships of a similar size.
## Changelog
:cl:
add: Adds a new ship
/:cl:


Vexylius (approved):
good ship

PR #988, opened 2022-04-20
Adds an optional surgery step to cancel with a cautery, if the active step does not require a cautery

### LINKED ISSUE #441: [BUG] Can't Cancel Surgeries

IndusRobot:
## About The Pull Request

If the surgery doesn't call for a cautery and a cautery is used, the surgery can be stopped 

If the surgery step does call for a cautery, the surgery continues as normal 

If the user is holding an incorrect tool, the surgery continues to notify the tool is incorrect

## Why It's Good For The Game

Fixes surgeries being unable to be cancelled, perhaps by misclicking an operation option or not knowing enough about a specific operation 

Fixes #441 

## Changelog
:cl:
add: Adds an optional surgery step to cancel with a cautery, if the active step does not require a cautery
/:cl:

### DIFF
--- a/code/modules/surgery/surgery_helpers.dm
+++ b/code/modules/surgery/surgery_helpers.dm
@@ -93,7 +93,8 @@
 
 	if(S.can_cancel)
 		var/required_tool_type = TOOL_CAUTERY
-		var/obj/item/close_tool = user.get_inactive_held_item()
+		// Historically surgical drapes were used with the cautery in the inactive hand, but these drapes don't seem to exist here
+		var/obj/item/close_tool = user.get_active_held_item()
 		var/is_robotic = S.requires_bodypart_type == BODYTYPE_ROBOTIC
 
 		if(is_robotic)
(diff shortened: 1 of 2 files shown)


Zevotech:
wasn't this already a thing with the cautery being in one hand and using an empty one on the body? or was it bugged? I recall it not always working when needed.

PrefabQuasar:
its yet another deeply cursed whitesands change

Zevotech:
we're better off without surgery drapes. this isnt a station where theres several rooms to get some from. also IPC/ghetto surgery exists which drapes would fuck with.

PrefabQuasar:
bedsheets have always worked as ghetto drapes, and are fairly easy to obtain
dont know if it really matters either way, since this fix pretty much gets the drapeless surgery on par with draped

ZephyrTFA:
This will lead to people accidentally canceling surgeries, add a do_mob check please

code/modules/surgery/surgery_helpers.dm
```diff
@@ -83,6 +83,9 @@
 
 /proc/attempt_cancel_surgery(datum/surgery/S, obj/item/I, mob/living/M, mob/user)
 	var/selected_zone = user.zone_selected
+	to_chat(user, "<span class='notice'>You begin to cancel the [S.name].</span>")
```
ZephyrTFA:
```suggestion
	to_chat(user, "<span class='notice'>You begin to cancel \the [S].</span>")
```

code/modules/surgery/surgery_helpers.dm
```diff
@@ -83,6 +83,9 @@
...
+	if ( !do_mob( user, M, 30 ) )
```
ZephyrTFA:
```suggestion
	if ( !do_mob( user, M, 3 SECONDS ) )
```

PrefabQuasar:
this seems to break when used with synthetic surgeries that require a welder, making it impossible to do the repair machinery surgery

PrefabQuasar:
correction after testing, it seems like the issue is actually that the close_surgery step only occurs when you try to use a cautery tool on someone, requiring that it be in your active hand, but for robotic steps, you need to have a screwdriver in your active hand, making it impossible to use this method to stop a robotic surgery

PrefabQuasar:
correction after testing, it seems like the issue is actually that the close_surgery step only occurs when you try to use a cautery tool on someone, requiring that it be in your active hand, but for robotic steps, you need to have a screwdriver in your active hand, making it impossible to use this method to stop a robotic surgery

PrefabQuasar:
oh hey github thanks for double sending my messages

PR #1018, opened 2022-05-01
Adds Wand Wresting

The-Moon-Itself:
## About The Pull Request
This PR adds Wand Wresting, when using a wand without any charges, there is a small chance (1 in 121) that the wand will fire one last time before disintegrating. That's it.

## Why It's Good For The Game
Rewards anyone with a empty wand if they're dedicated or desperate enough to squeeze out the last charge.

## Changelog
:cl:
tweak: Empty wands have a rare chance to activate one last time when used.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/magic/wand.dm
+++ b/code/modules/projectiles/guns/magic/wand.dm
@@ -1,5 +1,5 @@
-//For use in prob()
-#define WAND_WREST_CHANCE (100/121)
+//For use in prob() to determine if an empty wand will shoot once then break.
+#define WAND_WREST_CHANCE (1/121)
 
 /obj/item/gun/magic/wand
 	name = "wand"
@@ -57,7 +57,7 @@
 	update_icon()
 
 /obj/item/gun/magic/wand/shoot_with_empty_chamber(mob/living/user)
-	if(prob(WAND_WREST_CHANCE))
+	if(prob(100*WAND_WREST_CHANCE))
 		to_chat(user,"<span class='danger'>You manage to activate [src] one last time.</span>")
 		charges++
 		recharge_newshot()
@@ -255,3 +255,4 @@
 	ammo_type = /obj/item/ammo_casing/magic/nothing
 
 
+#undef WAND_WREST_CHANCE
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/guns/magic/wand.dm
```diff
@@ -9,6 +9,7 @@
 	can_charge = FALSE
 	max_charges = 100 //100, 50, 50, 34 (max charge distribution by 25%ths)
 	var/variable_charges = TRUE
+	var/wrest_chance = 0.82644628 //For use in prob(), about a 1 in 121 chance
```
Fikou: for something so specific, it'd be better as a define

The-Moon-Itself: I was thinking a define too initially, but I thought that someone might want the option to change this value for any child classes. (I.E. change it to 0 to disable wresting).
However, if you want me to to just make a define for the number then set the prototype value to that define, I could do that, but it seems a bit redundant as this is the only place this value shows up currently.

Fikou: if someone wants to disable wresting, they can then make it a variable, but i doubt anyone would ever do that. keep it as a define.

The-Moon-Itself: Okay, I've changed it to a define now.

code/modules/projectiles/guns/magic/wand.dm
```diff
@@ -1,3 +1,6 @@
+//For use in prob(), about a 1 in 121 chance
+#define WAND_WREST_CHANCE 0.82644628
```
ZephyrTFA:
```suggestion
#define WAND_WREST_CHANCE (1/121)
```

The-Moon-Itself: I'll change it (100/121) because for some reason byond has the input for prob() be a percentage and not a decimal.

code/modules/projectiles/guns/magic/wand.dm
```diff
@@ -1,3 +1,6 @@
+//For use in prob()
+#define WAND_WREST_CHANCE (100/121)
```
Fikou: #undef this at the end of the file

Fikou: also, 1/121 * 100 might be more readable?
also, the comment should probably state the reference or tell what the prob() is for

The-Moon-Itself: Okay, I’ve added the undef, changed the define for readability, and added more to the comment.

Fikou:
lovely, thank you!

PR #1042, opened 2022-05-07
Even more ship names

Apogee-dev:
## About The Pull Request

Adds a shipload of names to the ship name lists

## Why It's Good For The Game

Fewer repeated names, more minor jokes

## Changelog
:cl:
add: Added even more names for ship generation
/:cl:


strings/ship_names.json
```diff
@@ -465,9 +526,9 @@
 	"WEAPONS": [
 		"Aruval",
 		"Assegai",
-		"Blackjack",
```
Vexylius, line 468: after googling i found out that blackjack is actually a type of weapon, but I'm not objecting to removing it as it's already in general list. Just wanted to put that fun fact out there

Apogee-dev: Yep, that's why it was there to begin with. I'd forgotten it was also in the general list, though, and I want to avoid duplicates.

PR #1044, opened 2022-05-08
Adds speaker option to mining scanners

Hardly3D:
## About The Pull Request

This PR adds a speaker option to the mining scanners so you can change on whenever it should make sounds when pulsing and also disables the speaker by default on the automatic scanners while keeping it on for the manual scanners.

## Why It's Good For The Game

While the sound is kind of neat for immersion, it can and will eventually get annoying after hours of mining, this PR allows you to make the mining scanner shush it whenever you want it to.

## Changelog
:cl:Hardly
add: Mining scanners now have a speaker option.
tweak: Automatic mining scanners have their speakers set to off by default.
/:cl:


Zevotech:
I HEAR IT IN MY DREAMS

PR #1051, opened 2022-05-09
Stops multi-z transparency breaking when copying turfs (fixes plastic floors on shuttles)

Bokkiewokkie:
## About The Pull Request

Makes the turf_z_transparency element actually clean up its signals when detached, sets the default alpha of material data to 255, and removes plastic floors' material datum.

## Why It's Good For The Game

No more random (ship)test failures, no more runtimes.

## Changelog
:cl:
fix: Fixed plastic plating causing crashes.
code: Minor improvement to turf transparency element
code: Tweaked material datum opacity usage
/:cl:

### DIFF
--- a/code/datums/elements/turf_transparency.dm
+++ b/code/datums/elements/turf_transparency.dm
@@ -1,5 +1,6 @@
 /datum/element/turf_z_transparency
 	var/show_bottom_level = FALSE
+	var/turf/our_turf
 
 ///This proc sets up the signals to handle updating viscontents when turfs above/below update. Handle plane and layer here too so that they don't cover other obs/turfs in Dream Maker
 /datum/element/turf_z_transparency/Attach(datum/target, show_bottom_level = TRUE)
@@ -7,7 +8,7 @@
 	if(!isturf(target))
 		return ELEMENT_INCOMPATIBLE
 
-	var/turf/our_turf = target
+	our_turf = target
 
 	src.show_bottom_level = show_bottom_level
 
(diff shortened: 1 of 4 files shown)


ZephyrTFA:
Can you explain why this was causing the issues?

Bokkiewokkie:
> Can you explain why this was causing the issues?

I can, the only reason I have chosen not to is to preserve your sanity

Fikou:
> > Can you explain why this was causing the issues?
> 
> I can, the only reason I have chosen not to is to preserve your sanity

please do so.

Bokkiewokkie:
> > > Can you explain why this was causing the issues?
> > 
> > 
> > I can, the only reason I have chosen not to is to preserve your sanity
> 
> please do so.

So the plastic floor initializes, and the plastic material datum is added to the turf, calling its `on_applied()`
The plastic material datum does not have a set `alpha` value, making it equal to `null` or `""` (an empty text string) as this is the default for some reason.
This means that when `on_applied()` checks if it is transparent by running `if(alpha < 255)` it will always return TRUE, because `null`/`""` is less than `255`/`"255"` :)
Though that won't actually have an impact (our turf was already opaque), the material datum's `on_applied_turf()` is called from the proc to continue handling the turf's changes.
`on_applied_turf` again checks `if(alpha < 255)` (which we just discussed, *is* actually the case) and applies the multi-z turf transparency element to the turf as a result.
So far nothing has actually broken yet, the turf transparency can `Attach()` without issues, because the turf does not have a linked z level below it and doesn't tell anyone it should show what's below it.
The fun part is that running the unit tests runs the docking procs, which means the shuttles get to move.
During shuttle movement a bunch of things happen to the turfs to make sure they're all copied over properly, including the transfer of the very helpful comsignals our turf transparency element has registered on the turf! So when the new turf that will pretend to be our old plastic turf gets around to initializing, it wakes up to find the comsignals this transparency element is trying to register are already present! And overriding these comsignals sends out a `CRASH()`, causing the runtime.

ZephyrTFA (approved):

Fikou:
let's see

PR #558, opened 2021-12-28
speech controllers and hearers_in_view improvements

jupyterkat:
## About The Pull Request
ports
tgstation/tgstation#60219
tgstation/tgstation#63118

## Why It's Good For The Game

says

## Changelog
:cl: Kylerace, CoffeeKat
add: Added a speech controller
code: improved get_hearers_in_view()
/:cl:

PR #1058, opened 2022-05-12
[TM-ONLY] Revert "speech controllers and hearers_in_view improvements"

ZephyrTFA:
Reverts shiptest-ss13/Shiptest#558


Vexylius:
We might need this again actually

Vexylius:
...or not

Issue #1078, opened 2022-05-15
[FEATURE] Hyperspace Retrieval Device


Zevotech:
## Name of feature:
Any method which would allow players to retrieve bodies from the currently inescapable void of hyperspace.

### Additional Info:
We really need a way to allow players to bring in hyperspaced bodies without admin intervention. Whatever it may be, it needs to be able to access bodies outside of player vision range and retrieve them safely without being qdel'd by hyperspace. 
Some possible ideas:
A gun that can lock onto players and use a fulton extraction kit (or other method of retrieval) on them.
A limited use bluespace retriever that can be crafted and used to retrieve something at range. 
A hook-gun of sorts that lets you zoom out and fire a chain that pulls in players like the meathook weapon does.

Zevotech:
This method would obviously have to be quickly accessible by ships, given that you can't pen a body all the way out in hyperspace to keep a body from rotting out there.

IndusRobot:
If cloning still exists in this repo, that's an option. But I dunno how long ago TG removed cloning when whitesands and shiptest branched

Zevotech:
cloning is here, but it is locked behind tech and needs the brain/head of the person being revived to make an actual clone of them that isnt catatonic.

Zevotech:
If possible, I want this to be something that isn't locked behind tech. If someones captain falls into hyperspace and dies I want the crew to be able to retrieve the body without docking. Ideally by making the method of retrieval available in crafting or as a common ship thing.

Vexylius:
why use some ass high-tech device, where a spear on a rope will do. Just throw it at the corpse, embed it, then pull it in with the rope. That we could do. Or maybe some hook instead of a spear

Sylphily:
anything that is not a human pretty much gets instantly delted
or a mob i should say

zestybastard:
Hyperspace retrieval is somewhat of a thing already, I'd rather give players tools to make it easier or more interesting than adding a designated fix tool.

MarkSuckerberg:
Going to be real, deletion or harm at all was a stopgap measure. In the future I want something akin to a space encounter being created when something manages to reach the BORDER of the transit zone. Although maybe that might not be great for ejecting small bits of trash, it still would be pretty cool I think.

Zevotech:
bump

MarkSuckerberg:
imagine

PR #1081, opened 2022-05-16
Ship Keys

ZephyrTFA:
## About The Pull Request

Allows people to lock their helm console using a special key that is created at the helm console.

## Why It's Good For The Game

Allows Captains, or the first person to spawn a ship, to lock the Helm console to minimize the likelyhood of unsavory people doing unsavory things. Also gives tangible power to an individual over the ship.

## Changelog
:cl:
add: Helm Consoles can now be locked with a shipkey
add: Shipkeys, if your ship does not already have one , or was destroyed by an idiot, your helm console can print one
/:cl:


MusselMain:
This seems like it'd also give the captains of vessels more incentive to avoid leaving their ship if they have locked the helm and are holding the keys, as if they die their crew would be unable to leave until they locate their corpse

ZephyrTFA:
That would be a positive side effect, ideally captains are captains and not miners

ilikesaying:
Can you make the console make those “beep-beep sounds when unlocked

code/modules/overmap/ships/controlled_ship_datum.dm
```diff
@@ -238,3 +258,18 @@
 	shuttle_port.name = name
 	for(var/area/shuttle_area as anything in shuttle_port.shuttle_areas)
 		shuttle_area.rename_area("[name] [initial(shuttle_area.name)]")
+
+/obj/item/key/ship
+	name = "Ship Key"
```
Fikou: object names are usually lowercase

Vexylius:
bug:
* ship key name does not get updated when ship name is changed

possible issue:
* commonly appearing emag on syndicate ships is now a key opening all ships. That contributes to the problem where NT, SolGov and Independent ships are weaker than the syndicate
* if someone were to make the ship moving, then lock the console and walked out into hyperspace and died the crew would be doomed without admin intervention or https://github.com/shiptest-ss13/Shiptest/issues/1078

suggestions:
* would be cool if there was a way to let the AI pilot the ship despite console being locked
* could change the icon to something more visible, like a keycard maybe?
* could play /sound/machines/buzz_two.ogg when someone tries to use locked console
* could play /sound/machines/button3.ogg on unlocking and button4.ogg on locking
* if you could scan your captainship medal on the console or insert the key into the medal that would be cooler than to use a plain key
* ability to lock research server/ore silo
* remote locking/unlocking like on some car keys


overall a good feature

ZephyrTFA:
If the key is destroyed, i.e. hyperspace, it automatically unlocks and makes a new key

Vexylius:
it doesn't get destroyed when it's in the inventory of a corpse in hyperspace though, right?

ZephyrTFA:
Touche

ZephyrTFA:
You can now use a multi-tool to unlock a helm console, however doing so alerts the entire v level.
Remote unlocking is now possible by using the key in hand

HanSolo1519:
May be worthwhile adding a sneakier hacking system to futureproof it against antags

IE deconstructing the helm and having a slow do_after of “Hotwiring” the circuit board that doesn't alert the crew, allowing for dastardly turncoats to hijack the ship while leaving their targets stranded planetside, while still leaving the tator very exposed during the whole process (Can't move without interrupting the hack, looks kinda suspicious to be standing over a deconstructed helm).

Not a demand, I'm unsure there ever will be antags in the future.

Sylphily:
Sexy

PR #1170, opened 2022-06-05
Nerfs the oldcode nukie ship map for being a deathtrap.

Zevotech:
## About The Pull Request
Reduces the amount of damage that oldcode sword mobs did from 30 to 20. They already get armor penetration and free space movement alongside this damage, so the change is justified. Also reduces the density of mobs in the ruin to hopefully lower the amount of people this ruin murders.
image

## Why It's Good For The Game
Ruins should not be merciless deathtraps for players, and this ruin definitely fell under that category for me at least. Most players would approach this ruin in a mining hardsuit, look in the first room, and proceed to get mulched (or severely injured) by the oldcode esword mob with 30 melee damage and armor penetration that would lurk in the airlock. In all of my time observing and playing, I have never seen a crew make it past the first two rooms without casualties. Most would die in space as what is essentially a space carp with an esword would chase them into space and murder them for looking at them the wrong way.

If only we could avoid the sin that is what remains of tgs "antag balance" in our antagless codebase. 

## Changelog
:cl:
balance: decreased the oldcode sword mob's damage from 30 to 20
tweak: decreased the amount of mobs in oldcodeops.dmm
/:cl:

### DIFF
--- a/_maps/RandomRuins/SpaceRuins/oldcodeops.dmm
+++ b/_maps/RandomRuins/SpaceRuins/oldcodeops.dmm
@@ -39,8 +39,6 @@
 /obj/item/clothing/mask/gas/syndicate,
 /obj/item/clothing/shoes/sneakers/black,
 /obj/item/storage/backpack,
-/obj/item/clothing/head/helmet/swat,
-/obj/item/clothing/suit/space/swat,
 /obj/item/clothing/under/color/black,
 /obj/item/gun/ballistic/automatic/pistol/deagle,
 /obj/structure/closet{
@@ -48,6 +46,7 @@
 	name = "strange closet"
 	},
 /obj/item/ammo_box/magazine/m50,
+/obj/item/clothing/suit/space/hardsuit/syndi/elite,
 /turf/open/floor/oldshuttle,
 /area/ruin/powered)
 "fO" = (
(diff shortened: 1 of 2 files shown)


Calyxman:
IMO there should be something separating the two syndies in the hall, as they're probably still going to rush at the same time and fuck over any players in there.

Zevotech:
you have a point, I'll change that.

_maps/RandomRuins/SpaceRuins/oldcodeops.dmm
```diff
@@ -321,15 +316,13 @@
 /turf/open/floor/oldshuttle,
 /area/ruin/powered)
 "Xm" = (
-/obj/structure/chair/old{
-	dir = 4
+/obj/item/paper/crumpled/awaymissions/moonoutpost19/hastey_note{
+	info = "<i>Proceed with caution. This place is better than before, but still dangerous."
```
Fikou: let's not

Zevotech: im between the issue of the fact that players are obviously going to metaknowledge and avoid this ruin anyways or walk in and die so this was my best way of telling players "hey, this ruin got changed youll only get mulched slightly less now"

Latentish: This was removed in the following commit.

Latentish:
I agree with these changes. Since it used to look like this.
image
I'd say this should be testmerged.

Latentish (approved):
This hasn't had any problems and it's a simple edit. Margable.

PR #1183, opened 2022-06-07
optimizes icon2html() for icon files known to be in the rsc at compile time

triplezeta:
its just a direct port of https://github.com/tgstation/tgstation/pull/67429

## why its good for deez nuts
read that PR and see why. Also it makes both the context menu and the alt click statpanel menu way better from what I saw in localhost testing.

### DIFF
--- a/code/__DEFINES/flags.dm
+++ b/code/__DEFINES/flags.dm
@@ -41,6 +41,8 @@ GLOBAL_LIST_INIT(bitflags, list(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 204
 #define RAD_PROTECT_CONTENTS_1 (1 << 17)
 /// should this object be allowed to be contaminated
 #define RAD_NO_CONTAMINATE_1 (1 << 18)
+/// Should we use the initial icon for display? Mostly used by overlay only objects
+#define HTML_USE_INITAL_ICON_1 (1<<20)
 
 /// If the thing can reflect light (lasers/energy)
 #define RICOCHET_SHINY (1<<0)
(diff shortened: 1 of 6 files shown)


code/__DEFINES/flags.dm
```diff
@@ -51,6 +51,8 @@ GLOBAL_LIST_INIT(bitflags, list(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 204
 #define NOJAUNT_1 (1<<0)
 /// If a turf can be made dirty at roundstart. This is also used in areas.
 #define CAN_BE_DIRTY_1 (1<<1)
+/// Should we use the initial icon for display? Mostly used by overlay only objects
+#define HTML_USE_INITAL_ICON_1 (1<<20)
```
Fikou: you don't seem to have added this to anything

Fikou: also why is this in turf flags

triplezeta: pasted from TG to make the code compile

ZephyrTFA (approved):
I had asked for this to be ported; it looks like it was delegated to Tetra.
This is approved, but I suggest, and require, it be test merged to ensure it doesn't secretly break everything

triplezeta:
i wouldnt say delegated

code/__DEFINES/flags.dm
Fikou: move this to be where it should
and add it to places where it is on tg

triplezeta: leaving this comment as a reminder to myself to actually do this

triplezeta: ok done

triplezeta:
`HTML_USE_INITAL_ICON_1` moved out of turf flags and added to things that need it.

In localhost, this appears to make the alt-click stat panel menu and context menus both significantly more accurate with icons.

PR #1184, opened 2022-06-08
adjusts overmap encounter docking port placement

tmtmtl30:
## About The Pull Request

adjusts the spacing and placement of docking ports on overmap encounters, particularly empty space encounters. it's a bit difficult to explain in words so i made some pictures to help -- the image represents the full 127x127 virtual level, with a 3-tile border on all sides making a 121x121 tile usable space. the docking port areas are the colored rectangles inside that square. the lines connecting the docking ports are spacings -- if two lines have the same color, they have the same spacing.

docking ports on dynamic encounters used to be placed like this (the upper two docking areas only appear in empty space):
base

now they're placed like this:
new

this shouldn't break or intersect with any ruins, i think

## Why It's Good For The Game

even spacing on each side was my original intention when i wrote the code for placing docking ports on encounters, and i think it's better if the ships in empty space are more closely packed with one another. it lets people get to each other easier (and it also opens up more space for builds on my 0.1 outposts)

## Changelog
:cl:
tweak: Overmap encounter docking port placement has been tweaked to be slightly more consistent.
/:cl:

### DIFF
--- a/code/controllers/subsystem/overmap.dm
+++ b/code/controllers/subsystem/overmap.dm
@@ -288,8 +288,8 @@ SUBSYSTEM_DEF(overmap)
 
 	// locates the first dock in the bottom left, accounting for padding and the border
 	var/turf/primary_docking_turf = locate(
-		vlevel.low_x+RESERVE_DOCK_DEFAULT_PADDING+1 + vlevel.reserved_margin,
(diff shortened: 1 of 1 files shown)


MarkSuckerberg (approved):
Good change, I honestly had just kind of hacked together the additional docks, this is how they should've been implemented in the first place

PR #1191, opened 2022-06-09
Nerfs venus human traps

Zevotech:
## About The Pull Request
Reduces the amount of time a venus human trap's vine grab hardstuns, increases the cooldown on said attack, lowers damage and removes their ability to regain health on hit. Also disables the ability for ghosts to enter them at any time. 

## Why It's Good For The Game
This relic of old tg kudzu was unfortunately overpowered, and needed some adjustments to allow it to be used in ruins without it just fuckin murdering players. We also don't even have kudzu incidents here, so they really don't need to be this powerful.

## Changelog
:cl:
balance: venus human traps have reduced stuns, damage and lose the ability to regain health on hit. the vine attack now has a longer cooldown.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/venus_human_trap.dm
+++ b/code/modules/mob/living/simple_animal/hostile/venus_human_trap.dm
@@ -86,9 +86,10 @@
 	ranged = TRUE
 	harm_intent_damage = 5
 	obj_damage = 60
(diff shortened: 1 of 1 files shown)


MarkSuckerberg (approved):
Seems good to me

PR #1209, opened 2022-06-12
Fixes up /me's formatting

Recoherent:
## About The Pull Request
Manual emotes now remove the separatory space if the /me starts with an apostrophe, and also automatically adds a period to the end of the emote if it doesn't have punctuation already. The code for the space removal was already there and works with commas, but because apostrophes are HELL FORMATTED it didn't work for them.

## Why It's Good For The Game
Joe Tider's head explodes.
seriously HOW has this bug with apostrophes existed this long i hate html formatting i hate html formatting i hate html formatting

## Changelog
:cl:
fix: Emotes now properly remove the separatory space when started with an apostrophe.
tweak: Emotes now auto-add a period at the end if no punctuation is given.
/:cl:

### DIFF
--- a/code/datums/emotes.dm
+++ b/code/datums/emotes.dm
@@ -61,7 +61,13 @@
 		return
 
 	user.log_message(msg, LOG_EMOTE)
-	var/space = should_have_space_before_emote(html_encode(msg)[1]) ? " " : ""
+
+	//Removing space if necessary, and adding punctuation to the end
+	var/space = should_have_space_before_emote(html_encode(msg)) ? " " : ""
+	var/end = copytext(msg, length(msg))
+	if(!(end in list("!", ".", "?", ":", "\"", "-")))
+		msg += "."
+
 	var/dchatmsg = "<b>[user]</b>[space][msg]"
 
 	var/tmp_sound = get_sound(user)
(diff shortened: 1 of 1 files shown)


code/datums/emotes.dm
```diff
@@ -61,7 +61,13 @@
 		return
 
 	user.log_message(msg, LOG_EMOTE)
-	var/space = should_have_space_before_emote(html_encode(msg)[1]) ? " " : ""
+
+	//Removing space if necessary, and adding punctuation to the end
+	var/space = should_have_space_before_emote(html_encode(msg)) ? " " : ""
+	var/end = copytext(msg, length(msg))
+	if(!(end in list("!", ".", "?", ":", "\"", "-")))
```
Vexylius, line 68: would add a ' if we have "

Recoherent: i'd honestly opt against that; it would probably need to be copied to the say code, and in that case typing a message like, say, [that hat is Beans'], you still want a period added at the end. also i don't even think the " not adding a period works for the same reason regex catching ' is such a motherfucker and is the base reason this pr exists. tl;dr withered emoji i'll glance at it i guess

Recoherent: yeah actually i don't think either ' or " work in this particular line just because html encoding ouch

Recoherent: image
i guess it is time to brick everything by trying to refactor the end-punctuation detection for automatically adding a period. please kill me

PR #1211, opened 2022-06-13
Start checking for padding inside parentheses

martinlyra:
## About The Pull Request

Basically:

```DM
if ( hell_world )
    usr << "Ah shit"
```

Bad.

```DM
if (hello_world)
    usr << "Ah hello"
```

Better.

This PR adds an regex expression to the `check_regex.py` tool to check for these spaces inside the parentheses like how they appear in the first example above. It accounts for various "mutations" of the style so it catches more sincere typos.

## Why It's Good For The Game

Makes the code more superficially consistent, which make for more happy maintainers.

## Changelog
:cl:
/:cl:


martinlyra:
The rule is currently set to 0 instances allowed as we might need to make sure if this is actually a good idea.

ZephyrTFA:
> I feel like this is a passive aggressive attack :)

It'll all be over soon... Shhhh... >;)

martinlyra:
Didn't test-compile this. 

Plus the expression does not catch every instance because an actual compilatron slash parser is better suited for this job. As the comments and strings make it much harder to write a "flawless" regex.

As a result, the 22 matches so far are false positives inside comments and strings (usually regex expressions themselves). And there still are more sinners existing in the codebase.

ZephyrTFA (approved):
Alright, so there is no way I'm going to go over all of the files you changed with that regexp; checks are indicating successful code and I trust you to not insert malicious code

ZephyrTFA:
This is essentially only whitespace changes and I'd rather this not get stuck in limbo hell. Merging it in six hours if no other maintainer sees issue

martinlyra:
> This pull request has conflicts, please resolve those before we can evaluate the pull request.

Can you shits just not

I am trying to grill a mean steak right here for god's sake!

martinlyra:
Fixed again

ZephyrTFA:
Sending it before it breaks again

PR #1212, opened 2022-06-14
stops endround BS jumps from spamming the runtime log due to a missing timer flag

tmtmtl30:
## About The Pull Request

the proc that set the final bluespace jump timer on SSshuttle didn't use the TIMER_STOPPABLE flag in its add_timer call, so every time SSstatpanels fired near the end of the round, it'd cause a runtime. this fixes that by adding that flag to that add_timer call

## Why It's Good For The Game

runtimes bad

## Changelog
:cl:
fix: the end-round bluespace jump shouldn't spam the runtime log anymore
/:cl:


MarkSuckerberg (approved):
bloody hell

Nice, I bet this will fix the stat panel timer working too

PR #1218, opened 2022-06-15
Autolathe revamp (Autolathe TGUI port, cassette style disks, more design disks (functionality))

mel-byond:
## About The Pull Request

Since I noticed exporting designs onto design disks was a feature (holy shit, right)
And I had plans for an sort of "ghetto protolathe" that was much unwieldier to use, I discovered that this practically was in the code already. Just non-functional with how the checks were happening. Though, keeping the "import designs" method seems like it can be too easily abused

Generally seeing that autolathes are just weaker and dumber protolathes, hey, why not!

This PR changes the following:

- [x] Autolathe TGUI ported over directly from /tg/
- [x] Added 'eject disk' button to Autolathe UI 
- [x] Cassette style disks for Autolathes, Autolathes can only hold one disk. A disk may contain multiple designs. Remove disks using alt-click or in UI (Removed 'Imported' category)
- [x] Autolathes can now print designs only if it can be built using glass, iron, plastic, gold, silver, uranium or titanium. No bluespace or diamond! (lathe bit can't handle exotic materials :( )
- [x] When transfering Designs to disks, icons now show up!
- [x] Add more designs disk: basic (1 slot), advanced (3 slots), super (5 slots), elite (10 slots) (Please note, all except from the elite design disk can be fabricated by the autolathe itself too)
- [x] Added a small illustration for design disks

## Why It's Good For The Game

I wanted to make it actually feasable to use design disks, since, lets be honest. Nobody uses those fuckers

And mostly from a gameplay perspective, since the amount of protolathes are heavily limited I feel like a middle-ground between a protolathe and an autolathe must exists for those very specific items one might needs at the time when you don't have a protolathe but do have access to (yours? or someone elses?) a RnD console.

Also the general unwieldiness of swapping disks, no exotic material processing and limited designs per disk tries to balace it out from becoming another 'cheaper protolathe'.

Oh yeah, don't forget you can't transfer circuit board designs over to a autolathe.

also tgui is prettier

## Changelog
:cl:
qol: Autolathe interface has now been ported to TGUI
add: Added more design disks, with varying capacity.
add: Autolathes can now fabricate autolathe-friendly designs off a disk, working with more types of materials.
balance: Autolathes no longer import designs, instead it takes one disk at a time with designs on it.
/:cl:


mel-byond:
updated screenshots

tgui/packages/tgui/interfaces/Autolathe.js
```diff
@@ -0,0 +1,214 @@
...
+      <Window.Content scrollable>
+        <Section
+          title="Total Materials"
+          buttons={(
+            <Button
+              icon="eject"
+              content="Eject design disk"
+              disabled={!hasDisk}
+              onClick={() => {
+                  act('diskEject');
```
MarkSuckerberg:
```suggestion
                act('diskEject');
```
This'll fix that error

PR #1222, opened 2022-06-15
Gives a well deserved nerf to Icemoon mining fauna.

Zevotech:
## About The Pull Request
Removes the stupidly powerful health regeneration from Ice whelps, ice demons and polar bears. This was also removed on main TG for similar reasons, go figure.
Removes the armor peircing abilities of the ice whelp, due to it being buggy and applying to the fire breath for some reason. This means if you ever got hit directly by the fire breath attack and took a larger amount of damage than you should have, this is the code to blame.
Decreases the aggro distance of ice whelps, making it so that situations where they break onto your ship and massacre your crew happen even less often.

## Why It's Good For The Game
Ice fauna has been way too powerful for too long with this regen. When a planet becomes known for fauna that murders entire ships due to an unlucky spawn, its time for a nerf. Ice whelps will still be plenty powerful with their combination of burn and brute damage, so don't get too cocky and try to take one on without a proper weapon. 

Example of why this change was necessary: 
image
A single ice whelp murders an entire ship. Every time a crewmember tried to fight it, they would get it to almost no hp, die, and then watch in horror as it regenerated back to full health in under two minutes. They also husk bodies when they hit them with fire breath. Isn't that swell?

## Changelog
:cl:
balance: removed health regeneration from ice mining mobs
balance: removed armor peircing from the ice whelp
balance: lowered the ice whelp's aggro range
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/mining_mobs/ice demon.dm
+++ b/code/modules/mob/living/simple_animal/hostile/mining_mobs/ice demon.dm
@@ -65,12 +65,6 @@
 	SLEEP_CHECK_DEATH(8)
 	return ..()
 
-/mob/living/simple_animal/hostile/asteroid/ice_demon/Life()
-	. = ..()
-	if(!. || target)
-		return
-	adjustHealth(-maxHealth*0.025)
-
 /mob/living/simple_animal/hostile/asteroid/ice_demon/death(gibbed)
 	move_force = MOVE_FORCE_DEFAULT
 	move_resist = MOVE_RESIST_DEFAULT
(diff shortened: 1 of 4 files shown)


rebaia:
thank god, i was the captain of that ship and that was an absolute shitshow. round id 925, though i'm unsure if this server has a replay system

Zevotech:
You arent alone, this has happened to way too many people. It has happened to me three times, only one of which we actually barely recovered from with admin intervention.

Latentish:
A well needed nerf. Though I might have to ask you to remove their ability to push turfs around as well. Otherwise this is great.

MarkSuckerberg (approved):

PR #1225, opened 2022-06-16
Enigma-Class Ruin

ForestCranked:
## About The Pull Request

Adds a ruin version of the Enigma-Class. Now it has good loot and finally has somewhere for the frost miner to spawn. This ruin can be found on ice planets.

## Why It's Good For The Game

The game needs more ruins, and this also gives a place for the Demonic Frost Miner to spawn.

## Changelog
:cl:
add: the Enigma-Class ruin, which has some pretty good(?) loot and a boss in the form of the demonic frost miner.
/:cl:

### DIFF
--- a/_maps/RandomRuins/IceRuins/icemoon_surface_enigmaruins.dmm
+++ b/_maps/RandomRuins/IceRuins/icemoon_surface_enigmaruins.dmm
@@ -1,22 +1,4 @@
 //MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
-"ac" = (
-/obj/machinery/atmospherics/components/unary/vent_pump{
-	dir = 8
-	},
-/turf/open/floor/pod,
-/area/overmap_encounter/planetoid/ice)
-"bm" = (
-/obj/structure/cable{
-	icon_state = "4-8"
-	},
-/obj/machinery/atmospherics/pipe/simple/supply/hidden{
-	dir = 8
-	},
-/obj/machinery/atmospherics/pipe/simple/scrubbers/hidden/layer2{
-	dir = 8
-	},
-/turf/open/floor/plating/snowed,
-/area/overmap_encounter/planetoid/ice)
 "br" = (
 /obj/effect/decal/cleanable/dirt,
 /turf/open/floor/plating/snowed,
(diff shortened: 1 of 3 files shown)


PiperDoots:
You should probably remove some walls and sections of the ship. All in all it looks dirty enough but not less a ruin and more like a parked car as this is just basically a fully intact ship.

PiperDoots:
Also remove some of the piping and wiring whilst you're at it.

ForestCranked:
> 

i will cherish this forever, despite potentially having to close this for another PR.

triplezeta:
remove the `del`s from your changelog btw, you should only be logging things that have changed with the game relative to current master

ForestCranked:
> remove the `del`s from your changelog btw, you should only be logging things that have changed with the game relative to current master

done, thanks for the info

Latentish:
image
This is not a proper generic area (explored and unexplored ruin are acceptable, so is ruin/powered), otherwise I'd ask you to use custom areas.

image
The combat challenge you're using doesn't exactly have much impact on the ruin itself. The good loot is quite a bit away, outside the aggro range of the miner. So there's not much incentive to actually fight the miner. I'd recommend putting the miner closer to the loot, or making a subtype that drops a key into the bridge where the loot is.

Latentish (changes requested):
Like I said, I need to see some change to this level. You have what used to be a ship application, that's had a frost miner slapped into it without much impact on the focal point or likely player focus of the ruin, which seems to be the loot at the bridge. As a stand alone ship ruin, it doesn't have much to salvage, so it really has to fall back on gimmicks, puzzles, combat challenges, and overall design.
I will give you that the environmental story telling is pretty well nailed down.

If you are really attached to this ruin and want to see it put in, I'd ask the following:
-Add some way for the frost miner to have an impact on the level. As is they are just standing in the mostly empty miner's room.
-Distribute the loot a bit more or focus the combat challenge around it.
-Add some more stuff to salvage.
-Add some more to the shape of the ruin as it seems somewhat bland, or bring back more of what it looked like so that it has more flavor and good bits to look at.

A more direct instruction of what I would do would be:
- Move the frost miner to the bridge and add a different monster or two in the cargo area. Any ice planet mob will do. M
- Bring back that interim area at the middle of the ship, show some of those early design flaws to paint the whole ship as an engineering design failure that caused a crash. I'm talking those disposals units, the weird depressurized middle area between glass that had only siphons in it.
- Bring back more of the hull of the ship, as the original shape wasn't bad, even having some remnants of the botany (more than just some machine frames and gear) and bedroom would be good, and bring back some remnants of the thrusters (like one or two circuits for each side). This is a ruin, so we're telling a bit of a story. An engineer designed a ship, it didn't get produced too much, the designs were wonky and dangerous, someone flew it out, it had a serious issue and everyone crashed and died. Maybe add a letter or holotape about it somewhere.
- Add some tiling for what's outside, the ground of the iceplanet around the ship, it'll make it stand out more and not be encased in rock.

Otherwise, and I apologize, but I just don't see this being accepted.

ForestCranked:
Thanks for stating the obvious, dipshit.

PR #1233, opened 2022-06-19
fireblossoms are equippable now

triplezeta:
## About The Pull Request

title

## Why It's Good For The Game

image

## Changelog
:cl:
add: fireblossoms can be worn on your head
/:cl:


zestybastard:
big fan of this but maybe add a way we could alter the blossom to remove its light effect as an alternative variant? a lot of people get annoyed by the weird neon glow these things cause if there's other light sources available.

triplezeta:
yea its called the DNA manipulator

tiramisuapimancer:
ideasguying but can we have it so these raise your body temperature by a bit when theyre worn? would make these a good ghetto wintercoat or what have you

PR #1237, opened 2022-06-20
overmap star refactor

tmtmtl30:
## About The Pull Request

Refactors the code for /datum/overmap/star, making each star type a distinct subtype of /datum/overmap/star instead of having one subtype for each star size. As a result, the system's star is now chosen by picking from subtypes of /datum/overmap/star. The colors are semi-randomized based on the base spectral class for that star type. The behavior for this has been moved into /datum/overmap/star, instead of using subtypes of /obj/overmap, because subtyping /obj/overmap is dumb.

Also, stars rotate automatically now by calling SpinAnimation on the token. I don't think this looks very good, but @triplezeta asked me to do it, and I did, because it was easy.

## Why It's Good For The Game

The old code was bad, using subtypes of overmap icons and doing a lot of picking from lists when it should've been using subtypes. The new code is easier to work with and more modular. Plus, the examine texts are nice.

## Changelog
:cl:
add: Overmap stars now spin automatically.
add: Overmap stars now always have examine text.
add: Yellow giant stars now flicker.
refactor: Refactored overmap star code; star probabilities and colors are altered as a result.
/:cl:

### DIFF
--- a/code/modules/overmap/objects/star.dm
+++ b/code/modules/overmap/objects/star.dm
@@ -8,7 +8,6 @@
 	Rename(gen_star_name())
 	token.desc = token_desc
 	alter_token_appearance()
-	token.SpinAnimation(rand(1 MINUTES, 4 MINUTES))
 
 /datum/overmap/star/proc/gen_star_name()
 	return "[pick(GLOB.star_names)] [pick(GLOB.greek_letters)]"
--- a/code/modules/overmap/objects/star.dm
+++ b/code/modules/overmap/objects/star.dm
@@ -1,136 +1,169 @@
-/datum/overmap/star //TODO: refactor literally all of this this is painful
+/datum/overmap/star
 	name = "Star"
-	token_icon_state = "star1"
-	token_type = /obj/overmap/star
+	var/token_desc = "A star."
+	var/spectral_type = STAR_G
+	var/color_vary = 0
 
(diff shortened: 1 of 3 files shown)


triplezeta:
image
✨🎇✨🎇✨🎇✨🎇✨🎇✨🎇✨🎇✨🎇✨🎇✨🎇✨🎇✨🎇✨🎇✨🎇

code/modules/overmap/objects/star.dm
```diff
@@ -1,136 +1,169 @@
...
-		STARM,
-		STARL,
-		START,
-		STARY,
-		STARD
+/datum/overmap/star/binary/alter_token_appearance()
+	// we don't call the parent proc, since we do colors our own way
+	var/mutable_appearance/star_1
+	var/mutable_appearance/star_2
+	var/list/spectral_types = list(
```
MarkSuckerberg:
```suggestion
	var/static/list/spectral_types = list(
```
I *believe* this will be slightly faster as it won't be re-instantiated every single time it's run

MarkSuckerberg (dismissed):
I think this is all I could find, good job good job I am extremely glad to see this

MarkSuckerberg:
that was meant to be a requested changes but w/e

triplezeta:
okay the spinning was a mistake

Sylphily:
spin me baby right round

PR #1238, opened 2022-06-20
removes deprecated map_generator variable on /area

tmtmtl30:
## About The Pull Request

areas had a map_generator variable, that ran a map_generator datum on them when SSmapping initialized. presumably, this was a holdover from lavaland-type map generation; we don't do things this way anymore, so this variable was useless and did nothing. i removed it, along with 2 other clearly-unused things in the jungle map generator file. i tested it by booting up a local server and generating the planets automatically; didn't run into any runtimes. have a hard time seeing how this would cause errors unless i am monumentally misunderstanding things

## Why It's Good For The Game

dumb, confusing, unused variables shouldn't be allowed to sit around.

## Changelog
:cl:
code: Removed deprecated map_generator variable on /area.
/:cl:

### DIFF
--- a/code/controllers/subsystem/mapping.dm
+++ b/code/controllers/subsystem/mapping.dm
@@ -48,7 +48,6 @@ SUBSYSTEM_DEF(mapping)
 	repopulate_sorted_areas()
 	process_teleport_locs()			//Sets up the wizard teleport locations
 	preloadTemplates()
-	run_map_generation()
 
 	// Add the transit levels
 	init_reserved_levels()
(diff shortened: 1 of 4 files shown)


code/controllers/subsystem/mapping.dm
```diff
@@ -95,10 +94,6 @@ SUBSYSTEM_DEF(mapping)
 
 #define INIT_ANNOUNCE(X) to_chat(world, "<span class='boldannounce'>[X]</span>"); log_world(X)
 
-/datum/controller/subsystem/mapping/proc/run_map_generation()
-	for(var/area/A in world)
```
ZephyrTFA, line 99: This was seriously a var in world loop? 
Those are notorious for causing crashes who coded this

MarkSuckerberg (approved):

PR #1239, opened 2022-06-20
fixes some instances of biotypes as lists

tmtmtl30:
## About The Pull Request

biotypes are bitflags, not lists. IPCs and some simple animals had lists of biotypes instead of bitflags of biotypes, which was causing some runtimes due to the type mismatch. this PR changes the instances of list-biotypes that i could find into flag-biotypes. i may well have missed some but this is a step in the right direction

## Why It's Good For The Game

runtimes bad

## Changelog
:cl:
fix: Fixed some instances in the code where biotypes were assumed to be lists, instead of bitflags, when they're supposed to be bitflags.
/:cl:


MarkSuckerberg (approved):
Huh, neat.

PR #1276, opened 2022-06-30
Advanced Airlock Controller fix: Actually considers airlocks at edge of map as external

mel-byond:
## About The Pull Request

An issue with many maps is that if the airlock is at the edge of the map and turf checks happen to check if an airlock is external or not, the logic for it returns that it's just an internal one. And leaved the AAC broken, which sucks.

So, to fix this there is this small PR to fix this. I have tested this on my own PR ship (Flanger), Shetland and ScrapperB and some others that I don't recall that had broken AACs

## Why It's Good For The Game

Fixes the advanced airlock controller on many ships, such as but not limited to Shetland, ScrapperB, ...

## Changelog
:cl:
fix: Advanced Airlock Controllers now considers airlocks at edge of map as external
/:cl:


ZephyrTFA (approved):
INSERT REVIEW HERE

PR #1295, opened 2022-07-04
Sanitizes _maps and removes old unused items

Yellow-Mushroom:
## About The Pull Request

Removes old and unused .dmm's, goes through a process of checking the map in Strong and then checking if the map is referenced anywhere in code. If it isn't it is removed. Discord was consulted for most maps

**--Away missions in _maps/RandomZLevels are being preserved for the purpose of translating them into ruins at a later point**
_maps/shuttles/infiltrator/infiltrator_advanced has been preserved despite being broken because its a pretty cool ship and can be refurbished at a later point
--the hunter ships, while being completely non-functional/old old are used for a gamemode thing that im unsure if is still used
--same for the pirate ship
--preserves shuttles/ruins, while unused in game, some of these could be salvaged for ruins at a later point
--moves all non shuttle/shiptest folders into a new folder called shuttle/misc

While this isnt a complete sanitation, it does as much as it can within the scope of this PR

## Why It's Good For The Game

sanitization is very nice, makes it easy for newer players to find what theyre looking for and be able to contribute easier.


Yellow-Mushroom:
found a potential bug, moving bag into draft

Latentish:
Well this breaks at least 1 ruin. But it's also a ruin we intend to remove anyway? There's a ruin that relies on one of those shuttles iirc, by using a docking call when the ruin spawns. But uh, it also doesn't work? So yeah I'm pretty content with this change. Good to clean up house now that we're actually maintaining shit.

Latentish (approved):
I'd like to see a testmarg.

PR #1308, opened 2022-07-06
The Grand Limbgrower TGUI update, now two new research nodes, data disks, snazzy ui I modified specifically for this server, way more options, and a QoL feature

Chubbygummibear:
## About The Pull Request
Currently the Limbgrower is turbo garbage and only has a few race options. This is a port of TG's limbgrower update and a lot of shiptest original code written by my own sinful hands. 
The tl;dr is

- Limbgrower has a snazzy new ui
- Limbgrower when first built can print arms, legs and all organs for every roundstart organic race (snowflake limb races rejoice)
- New Tech node right after advanced biology called Xeno-Organ Biology. unlocks data disks which can be uploaded to the limbgrower to add more printing options in the form of non-roundstart xeno specific organs. costs 2500 research.
- Limbgrower will accept raw meat and grind it into synthflesh for fuel. Bear with me for a second for this one. On Yogstation when I first did this port, I added the hugbox meat processing code because people were used to fueling the cloner with meat for biomass instead of synthflesh. the reason I think it's acceptable here is because sometimes resources are very thin, so the ability to create spare limbs and organs without a chemstation could be the difference of life and death for some crew, and meat is abundant on planets with fauna. Yes this can be used to print synthflesh by grinding meat, disassembling the machine, and taking the full beakers out. Let me know if this is too broken. I think another means of getting a vital reagent is just another boon.

what it looks like on the tech tree
image

unupgraded default view
image

upgraded and with all the lesser known race disks installed
image

printed from a protolathe with med access, then you just use them on the limbgrower to unlock more organ designs
image

limbs with mutant colors spawn with a random color. go nuts, make yourself a rainbow if you have the patience to keep rolling limbs
(as of now the colors don't actually stay, sorry to dash your hopes of being a walking pride flag. someday that will get sorted out)
image

## Why It's Good For The Game

The current limbgrower is severely lacking in functionality, and suffers from an old ui on top of that. it only prints limbs for a handful of races and has an awkward reagent mechanic. I wanted to port this rework (which also turned into a lot of original code along the way) to make an old machine actually useful. I think it will benefit a lot of ships to have access to a relatively early means of replacing organs and limbs in their crewmembers. Especially in the case of Kepori in which our friend had to go single armed for over an hour because we didn't have cloning and the prosthetics at the time didn't work. This would provide a much quicker means to rectifying that sort of thing.

In addition to being an early option for healing, it can also be found while salvaging rock planets, making it even more available to ships with less starting amenities. 

I think the hardest sell of what I did was the ability to feed it meat to create synthflesh. This is a holdover from yog where the cloners just took meat, so I added this functionality to allow people a method of operation they were already familiar with. However i think it fits here because meat is easily obtainable through fauna, but chemistry setups can prove difficult to obtain for the more difficult ships. The synthflesh created from the meat does actually stay in the beakers inside the machine, so it does effectively make this a meat -> synthflesh converter, but again in this setting where a chemistry station can be out of the question, being able to synthflesh douse someone and shock them back to life is a very possible scenario. So i would like it to stay

## Changelog
:cl: Melbert, Chubbygummibear
add: new Limbgrower tgui interface
add: way more bodypart designs in the limbgrower, such as kepori and vox limbs to the standard limbgrower
add: new research node called Xeno-Organ Biology, which unlocks limbgrower design disks that allow you to print non-roundstart race limbs and organs
add: new hidden research node that functions like illegal tech and alien tools, Xenomorph Biology. Get out there and kill a xeno, bring it back, and extract some organs to deconstruct to unlock it.
add: Limbgrower accepts raw meat as input and grinds it into synthflesh to use as fuel
del: Limbgrower old ui gone
code: Added some Dwarf stuff (don't worry about it)
code: a lot of snowflake code to make digi legs printable. It is a mess and i'm sorry but it works
/:cl:


triplezeta:
can you merge the xenoorgan biology node with advanced biotech and add a human limb/organ disk

Chubbygummibear:
> can you merge the xenoorgan biology node with advanced biotech and add a human limb/organ disk

human organs are available right from the get-go. unless you mean remove them and require a human disk. The advanced biotech merge is a decent idea. i know research can be a pain to get so even 2500 might be a big ask

triplezeta:
that is what i mean yes

Chubbygummibear:
> that is what i mean yes

i would like to keep the default organ set available right away because the limbgrower boards can be gotten from the rock planet and for a ship that doesn't immediately have research capabilities than can be a godsend. I've been in a position before where my crew member had died on a sand planet after already receiving a heart bypass. so in order to get a replacement heart i had to hunt a legion for a replacement. these kind of dire straits scenarios are why i want the limbgrower to be able to fill a need immediately. in addition, even if you have a research console and server, you might not have a protolathe to print the disks. meaning you still might not have access to organs.

triplezeta:
for the sake of lore i want to avoid having there be a "default" set of organs. as an alternative i would suggest just giving the limb grower all roundstart species limb and organ designs by default instead of separating them out into disks

Chubbygummibear:
> for the sake of lore i want to avoid having there be a "default" set of organs. as an alternative i would suggest just giving the limb grower all roundstart species limb and organ designs by default instead of separating them out into disks

definitely gonna try to talk to some devs about balance. i would like it to be an organ printing option pretty quick, but i'm not sure if it would be too strong having all the designs immediately

Chubbygummibear:
after talking with a project lead all the roundstart race limbs and organs will immediately be available, and the more exotic races will be design disk locked. working on this now

code/game/machinery/limbgrower.dm
```diff
@@ -15,222 +11,317 @@
...
-	if(default_deconstruction_screwdriver(user, "limbgrower_panelopen", "limbgrower_idleoff", O))
-		updateUsrDialog()
+	if(istype(user_item, /obj/item/disk/design_disk/limbs))
+		user.visible_message("<span class='notice'>[user] begins to load \the [user_item] in \the [src]...</span>",
+			"<span class='notice'>You begin to load designs from \the [user_item]...</span>",
+			"<span class='hear'>You hear the clatter of a floppy drive.</span>")
+		busy = TRUE
+		var/obj/item/disk/design_disk/limbs/limb_design_disk = user_item
+		if(do_after(user, 2 SECONDS, target = src))
+			if(!categories.Find(limb_design_disk.species))
```
MarkSuckerberg:
```suggestion
			if(!(limb_design_disk.species in categories))
```

code/game/machinery/limbgrower.dm
```diff
@@ -15,222 +11,317 @@
...
+		for(var/found_category in species_categories)
+			if(found_category in limb_design.category)
+				species_categories[found_category] += limb_design
+
+	for(var/category in species_categories)
+		var/list/category_data = list(
+			name = category,
+			designs = list(),
+		)
+		for(var/datum/design/found_design in species_categories[category])
```
MarkSuckerberg:
```suggestion
		for(var/datum/design/found_design as anything in species_categories[category])
```
Since we know it'll always be a list of purely `/datum/design`s.

MarkSuckerberg (changes requested):
Just realised I didn't actually send these... 😔

MarkSuckerberg (approved):

PR #1312, opened 2022-07-07
Adds Gas Station ruin to Ice planet

zombiemimic:
## About The Pull Request
image
Adds a gas station ruin to the Ice planet.
Contains 4 space pirates, 3 russians, and several toys.

## Why It's Good For The Game
More ice planet ruins, which I hear are in short supply.

## Changelog
:cl:
add: Added gas station ruin to the ice planet
/:cl:


Latentish (changes requested):
This ruin isn't horrible. But it needs some work.
- This ruin is both a death trap and a loot box. These enemies will charge down players and just wipe unexpecting teams. The doors used here won't prevent them from aggroing and leaving to kill players.
- There's waaaayyyy too much loot here. Way too many guns and ammo and just goodies. If this is a gas station I'd like to see more general goods with maybe a gun behind the counter.
I think this ruin would work better if the enemies were thinned out along with the loot overall. And make sure to test it to make sure the iceplanet doesn't outright kill your mobs due to atmos.

PR #1325, opened 2022-07-10
Init Time Reduction for Developers

MarkSuckerberg:
## About The Pull Request
Adds a blank map with size 127x127 (minimum for normal encounters) with only the necessities to replace CentCom. This drastically improves loading times, reducing it to down to around 12 seconds on my PC at least.

Also changes some config options to make dev environments as fast as they can be. Warning

## Changelog
:cl:
config: Default overmap sizes reduced significantly to speed up dev environment spin-up time.
code: <b>IF YOU WANT TO HOST A NON-DEVELOPMENT SERVER, MAKE SURE TO UNCOMMENT THE `#define FULL_INIT` LINE IN `_basemap.dm`!!!</b>
code: Renames the "Bot Comms" subsystem to the "Redbot" subsystem because that's what the file and type are. It should match!
/:cl:


MarkSuckerberg:
Also, already added the flag to our server-side code modifications.

ZephyrTFA (approved):

MarkSuckerberg:
it works it works it works it works it works it works

ZephyrTFA:
This is good to go, and works rather well. I wonder if it would be possible to output "generating XXX out of XXX planets" or something similar though in the future just to let players know how long

PR #1332, opened 2022-07-12
fix rotting stuff causing runtimes when in space

ZephyrTFA:
does that which the title says it does


tmtmtl30:
is this right? space turfs shouldn't have null air, they should have immutable space air. i think /turf/open/space/basic turfs might end up with null air in some cases, but that's a pretty different situation

MarkSuckerberg:
closing this since miasma no longer exists, also I'm swatting zeph for using a branch on the master repo

PR #1337, opened 2022-07-13
Prevent duplicate names for vents and scrubbers

antropod:
Makes scrubbers and vents re-use sequential ids for their names. No longer you will have two vents #9 in the area.

## About The Pull Request

Currently game allocates names for new vents/scrubbers based on how many there are scrubbers in the area. New index for that name will be `names.len + 1`. Assume you have 9 scrubbers in the area. So if you remove, say scrubber #2 and then re-wrench it, it will be named scrubber #9 and have same name as your existing scrubber #9.
On /tg/ this issue was fixed 2 years ago by making scrubbers/vents have random alphanumeric name, which is pain to deal with (original idea was to re-use ids, but then it went south during the conversation in PR)
Here I'm using algorithm for allocating names that was originally proposed in https://github.com/tgstation/tgstation/pull/52398 by myself.

Benefit of this approach is that if you unwrench one vent/scrubber and wrench it back, it will have same name as before.

## Why It's Good For The Game

Makes construction of atmos systems less painful.

## Changelog
:cl:
fix: Prevent duplicate names for vents and scrubbers
/:cl:

### DIFF
--- a/code/__HELPERS/sequential_id.dm
+++ b/code/__HELPERS/sequential_id.dm
@@ -1,0 +1,44 @@
+// Algorithm for allocating names for vents/scrubbers (you may use it for other purposes as long as it has same interface)
+//
+// By "name" here I mean sequential number of the scrubber/vent in the area
+// I keep track of all used names in the area with an array `names`
+// Array index means name, array value means id_tag
+// If value is null, the name is not used
+//
+// # So for example area has vents #1, #2 and #4, then array of names looks like
+// ["tag1", "tag2", null, "tag4"]
+//
+// When you allocate new name for id_tag, you scan array for free slot,
+// assign id_tag to that slot and return its index
+// if all slots occupied, you add value to the end of array
+//
+// When you deallocate name for id_tag, you find id_tag in the array
+// and set value of that element to null (or to some other meaningless value for id_tag)
+
+/**
+	Make new sequential number for this id_tag or reuse one of the free ones
+*/
+/proc/allocate_nameid(list/names, id_tag)
+	for (var/i in 1 to length(names))
+		if (names[i] == id_tag)
+			// this id_tag already has a name
+			return i
+		if (names[i] == null)
+			// found empty slot in the array
+			names[i] = id_tag
+			return i
+	// end of array, add new value to the end
+	names += id_tag
+	return length(names)
+
+/**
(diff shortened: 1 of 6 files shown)


tmtmtl30:
i'm not a huge fan of this, but that's mostly because i hate areas with a violent, searing passion, and see anything that uses them as an assault on all that is good and just. random IDs would avoid my ire, but i'll have to look at this + the original PR

antropod:
This is my experience on working with atmos projects:

Random IDs are annoying, because you have to remember each and every one them and look trough lists to find them.

Sequential ids are much easier to remember, for example you made a burn chamber with 4 scrubbers, and now you just remember that scrubbers 1 to 4 are "burn chamber"

Another problem with random ids is ordering, good luck finding scrubber Xy9sd in the list of 20 scrubbers.

This is why I originally proposed to re-use ids, and PR was going this way, but somewhere in the middle of the conversation it all went south and author just made low-effort random naming

antropod:
@tmtmtl30 can you take a look?

code/__HELPERS/sequential_id.dm
```diff
@@ -0,0 +1,44 @@
...
+// if all slots occupied, you add value to the end of array
+//
+// When you deallocate name for id_tag, you find id_tag in the array
+// and set value of that element to null (or to some other meaningless value for id_tag)
+
+/**
+	Make new sequential number for this id_tag or reuse one of the free ones
+*/
+/proc/allocate_nameid(list/names, id_tag)
+	for (var/i in 1 to names.len)
```
MarkSuckerberg:
```suggestion
	for (var/i in 1 to length(names))
```

code/__HELPERS/sequential_id.dm
```diff
@@ -0,0 +1,44 @@
...
+		if (names[i] == id_tag)
+			// this id_tag already has a name
+			return i
+		if (names[i] == null)
+			// found empty slot in the array
+			names[i] = id_tag
+			return i
+	// end of array, add new value to the end
+	names += id_tag
+	return names.len
```
MarkSuckerberg:
```suggestion
	return length(names)
```

MarkSuckerberg (changes requested):
Sorry for taking so long, here's just some general small optimizations. Seems good

MarkSuckerberg (approved):

PR #1338, opened 2022-07-13
Minor(?) rewrites to lighting code

tmtmtl30:
## About The Pull Request

This PR does three things.

First, it makes the code that determines area luminosity and lighting mode a little bit more comprehensible. I don't think I changed any behavior (especially since we don't use any areas with DYNAMIC_LIGHTING_IFSTARLIGHT anymore, though if we DID, it'd fix a minor bug), I just wanted to make it easier to read.

Second, it stops /turf/change_area() from adding dynamic lighting overlays when the turf isn't supposed to be dynamically lit. I'm not sure if this caused any noticeable problems, but it contributed to some performance inefficiency when generating planets.

Third, it moves one of the code's two starlight updates into /create_all_lighting_objects() (run by SSlighting) and out of /atom/movable/lighting_object/Initialize(). This means there may be slightly more light_source datums created in certain situations and makes init times 1-3 seconds slower locally (depending on whether the starlight config flag is set), but it should also speed up planet generation slightly as ChangeTurf() now calculates starlight only once instead of twice when transitioning from a space turf to a non-space turf. I wouldn't expect a huge change, though; I can do some timing locally if requested.

## Why It's Good For The Game

Hopefully speeds up planet generation and improves code quality somewhat. Fingers crossed.

## Changelog
:cl:
refactor: Minor changes to some area luminosity and starlight code to improve code quality and performance.
/:cl:

### DIFF
--- a/code/controllers/subsystem/lighting.dm
+++ b/code/controllers/subsystem/lighting.dm
@@ -13,12 +13,6 @@ SUBSYSTEM_DEF(lighting)
 
 /datum/controller/subsystem/lighting/Initialize(timeofday)
 	if(!initialized)
-		if (CONFIG_GET(flag/starlight))
-			for(var/I in GLOB.sortedAreas)
-				var/area/A = I
-				if (A.dynamic_lighting == DYNAMIC_LIGHTING_IFSTARLIGHT)
-					A.luminosity = 0
-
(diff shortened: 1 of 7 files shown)


MarkSuckerberg (approved):
Seems good, good catch

PR #1339, opened 2022-07-13
Makes TEG not break when you sneeze on it (fix integrity_falure values)

antropod:
## About The Pull Request

I was playing with TEG the other day and it broke when I accidentally hit it with multitool.
It turns out `integrity_failure` were set incorrectly (it should be a number between 0 and 1)

## Why It's Good For The Game

Fixes are good.

## Changelog
:cl:
fix: Makes TEG not break when you sneeze on it
/:cl:


MarkSuckerberg (approved):
Whew

PR #1351, opened 2022-07-17
this breakey, fixey (test PR, for the synapse breaking branches)

spockye:
## About The Pull Request
synapse json change is forcing itself onto every new branch. this is to check if it can be fixed quickly

## Changelog
:cl:
/:cl:


ZephyrTFA:
Don't put stuff like that in the CL, either have an actual CL or have none.

PR #1360, opened 2022-07-20
Removes Cloning research

Apogee-dev:
## About The Pull Request

Removes Cloning from the tech web. Simple as.

## Why It's Good For The Game

Cloning is extremely prone to abuse when ships get access to it and often ships with science or medical lathes will simply beeline to cloning so they can farm research off of cloned golems. Additionally per dev team discussion cloning is supposed to be extremely rare on the frontier and player access to it should be much more limited. Cloning is still available through a few rare sources, namely ruins and as salvage from wrecked machines on junk planets, so securing access to it should be a much more significant event and not simply a matter of routine.

## Changelog
:cl:
del: Removed cloning from the tech web
/:cl:

### DIFF
--- a/code/modules/research/designs/comp_board_designs.dm
+++ b/code/modules/research/designs/comp_board_designs.dm
@@ -96,6 +96,14 @@
 	category = list("Computer Boards")
 	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL | DEPARTMENTAL_FLAG_SCIENCE
 
+/datum/design/board/cloning
+	name = "Computer Design (Cloning Console)"
+	desc = "Used to clone people and manage DNA."
+	id = "cloning"
+	build_path = /obj/item/circuitboard/computer/cloning
+	category = list("Computer Boards")
+	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL | DEPARTMENTAL_FLAG_SCIENCE
+
 /datum/design/board/comconsole
 	name = "Computer Design (Communications)"
 	desc = "Allows for the construction of circuit boards used to build a communications console."
--- a/code/modules/research/designs/comp_board_designs.dm
+++ b/code/modules/research/designs/comp_board_designs.dm
@@ -96,14 +96,6 @@
 	category = list("Computer Boards")
 	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL | DEPARTMENTAL_FLAG_SCIENCE
 
-/datum/design/board/cloning
-	name = "Computer Design (Cloning Console)"
-	desc = "Used to clone people and manage DNA."
-	id = "cloning"
-	build_path = /obj/item/circuitboard/computer/cloning
-	category = list("Computer Boards")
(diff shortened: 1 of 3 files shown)


rye-rice (approved):

Ketrai:
To be fair this pr doesn't really address cloning being able to be used for infinite research. If anything it's a bit of a bandaid fix. It also means if any other ways to easily get exotic bodies props up, it will still be abused just the same. I do hope maybe some messed up, crude,  more limited form of cloning gets propped up instead later because it can be a lot of fun to mess around with cloning for weird science experiments like doppelgangers. 

Other than that, this seems fine. When dumb stuff becomes meta it either needs to be made harder or temporarily removed

AtlasRules:
god we need disksci so badly

Apogee-dev:
> To be fair this pr doesn't really address cloning being able to be used for infinite research. If anything it's a bit of a bandaid fix. It also means if any other ways to easily get exotic bodies props up, it will still be abused just the same. I do hope maybe some messed up, crude, more limited form of cloning gets propped up instead later because it can be a lot of fun to mess around with cloning for weird science experiments like doppelgangers.
> 
> Other than that, this seems fine. When dumb stuff becomes meta it either needs to be made harder or temporarily removed

It's a bit of a bandaid, but it's still a change that's been talked over a fair bit before. Cloning being easy to abuse for RnD is a separate issue from widely-accessible cloning being at odds with the overall design of shiptest in the first place, after all. they both need fixing but they both have different solutions. This is mostly just to address the second one by making cloning harder to find in the first place, I have some other ideas that would help with the first issue by taking cues from other codebases.

code/modules/research/designs/comp_board_designs.dm
```diff
@@ -96,14 +96,6 @@
...
 	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL | DEPARTMENTAL_FLAG_SCIENCE
 
-/datum/design/board/cloning
-	name = "Computer Design (Cloning Console)"
-	desc = "Used to clone people and manage DNA."
-	id = "cloning"
-	build_path = /obj/item/circuitboard/computer/cloning
-	category = list("Computer Boards")
-	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL | DEPARTMENTAL_FLAG_SCIENCE
-
```
Bokkiewokkie: You might still want to keep the designs to put on to disks

code/modules/research/designs/machine_designs.dm
```diff
@@ -192,14 +192,6 @@
...
 	category = list("Medical Machinery")
 
-/datum/design/board/clonepod
-	name = "Machine Design (Cloning Pod)"
-	desc = "An electronically-lockable pod for growing organic tissue."
-	id = "clonepod"
-	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL | DEPARTMENTAL_FLAG_SCIENCE
-	build_path = /obj/item/circuitboard/machine/clonepod
-	category = list("Medical Machinery")
-
```
Bokkiewokkie: Ditto

MarkSuckerberg (approved):
please go quietly into that good night

PR #1374, opened 2022-07-23
adjusts reaction priorities to stop bugs

tmtmtl30:
## About The Pull Request

lavaland planets have been mass-producing nitryl for a while. turns out auxmos breaks when you have two reactions with the same priority. they overwrite each other. this PR ensures there are no reaction priority collisions

## Why It's Good For The Game

makes reactions act how they should, stops lavaland planets from sometimes ending up all brown and ugly

## Changelog

:cl:
fix: Nitryl now only forms when it should.
/:cl:

### DIFF
--- a/code/modules/atmospherics/gasmixtures/reactions.dm
+++ b/code/modules/atmospherics/gasmixtures/reactions.dm
@@ -343,7 +343,7 @@
 			air.set_temperature((temperature*old_heat_capacity + energy_released)/new_heat_capacity)
 
 /datum/gas_reaction/genericfire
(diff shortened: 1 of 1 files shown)


tiramisuapimancer:
Hold on isn't the nitryl reaction being broken the only reason auxmos works on here.

tmtmtl30:
i think that was some other reaction, and that got fixed anyway. i hope. i'm sure it's fine

MarkSuckerberg:
Honestly just going to assume this works. Seems to be the proper solution according to what I've heard from Putnam/Coffeekat.

PR #1388, opened 2022-07-26
Removes golem mutation toxin, readds the IPC mutation toxin recipe

Zevotech:
## About The Pull Request
Removes golem mutation toxin, which shouldn't exist
Re-adds the IPC mutation toxin recipe, which should exist but got removed in the betterlimbs port for an unspecified reason which was not explicitly stated in the PR or its original port.

## Why It's Good For The Game
The reason I was given for why IPC mutation toxin didn't exist was because IPCs could not process mutation toxins and thus could not be turned back if they had been mutated into an IPC in the first place. This was tested and proven wrong.  IPC mutation toxin would reasonably be the only way to create an IPC in-game without admin intervention, and should exist for that reason.

On the other hand, despite the above reasoning, golem mutation toxin still existed and is the actual mutation that you cannot reverse, as golems cannot process mutation toxins. This removes that and prevents such a situation from ever happening.

## Changelog
:cl:
add: IPC mutation toxin recipe
del: Golem mutation toxin
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -631,14 +631,6 @@
 		return TRUE
 	return ..()
 
-/datum/reagent/mutationtoxin/golem
-	name = "Golem Mutation Toxin"
-	description = "A crystal toxin."
-	color = "#5EFF3B" //RGB: 94, 255, 59
-	race = /datum/species/golem/random
-	process_flags = ORGANIC | SYNTHETIC //WS Edit - IPCs
-	taste_description = "rocks"
-
 /datum/reagent/mutationtoxin/abductor
 	name = "Abductor Mutation Toxin"
 	description = "An alien toxin."
--- a/code/modules/reagents/chemistry/recipes/others.dm
+++ b/code/modules/reagents/chemistry/recipes/others.dm
@@ -627,10 +627,6 @@
 	results = list(/datum/reagent/mutationtoxin/pod = 1)
 	required_reagents  = list(/datum/reagent/mutationtoxin/unstable = 1, /datum/reagent/plantnutriment/eznutriment = 10)
 
-/datum/chemical_reaction/mutationtoxin/golem
-	results = list(/datum/reagent/mutationtoxin/golem = 1)
-	required_reagents  = list(/datum/reagent/mutationtoxin/unstable = 1, /datum/reagent/liquidadamantine = 20)
-
 /datum/chemical_reaction/mutationtoxin/abductor
 	results = list(/datum/reagent/mutationtoxin/abductor = 1)
 	required_reagents  = list(/datum/reagent/mutationtoxin/unstable = 1, /datum/reagent/medicine/morphine = 10, /datum/reagent/toxin/mutetoxin = 10)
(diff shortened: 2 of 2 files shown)


TallDeekus:
Based Logic

Sylphily:
golem mution is easy to reverse they can infact
proccess chems

Ketrai:
> golem mution is easy to reverse they can infact proccess chems

They can, but you cannot easily inject them with chems. Since they don't take needles, I'm not sure about hypospray, but otherwise you might need a custom cryo mix?

Ketrai:
The real reason golem mutation toxin isn't a concern is the fact it needs adamantine. Which you're very unlikely to obtain without xenobiology. With xenobiology you can reverse the transformation anyways via slime extracts. So the average joe experimenting with mutation toxins isn't going to screw themselves over.

Sylphily:
fun fact, pill exist

Ketrai:
> fun fact, pill exist

TRUEEEEEEEEEEEEEEEEEE

Latentish:
I really don't think removing antag and unique species creators is very necessary on shrimptest. But adding the IPC mutation is fine maybe? that's a question for loreheads.
Also if you ingest golem toxin, either it's your own fault, you consented to it, it happened in pvp, or an admin needs to talk to somebody.

Zevotech:
> golem mution is easy to reverse they can infact proccess chems

wrong.
image
they dont process any mutation toxins.

Zevotech:
The reason golem mutation toxin is being removed is because it still existed in spite of the supposed logic that IPC mutation toxin got removed for. I was told IPC mutation toxin got removed because IPCs couldnt process mutation toxins and thus couldn't be turned back. This was incorrect, and only applied to golems, despite having their own mutation toxin. IPC mutation already existed and still was a thing, but its RECIPE was removed for an unlisted reason in the betterlimbs port. If a maintainer or lead wants me to leave golem mutation toxin in I'll change the PR but I will die on the hill of IPC mutation toxin needing a recipe.

Zevotech:
> I really don't think removing antag and unique species creators is very necessary on shrimptest. But adding the IPC mutation is fine maybe? that's a question for loreheads. Also if you ingest golem toxin, either it's your own fault, you consented to it, it happened in pvp, or an admin needs to talk to somebody.

getting turned into a golem and not being able to be turned back is still a stupid and terrible thing, so as stated in the PR description this prevents it from ever happening.

Sylphily:
in that case, please fix the fact that they can't process mutation chems

Sun-Soaked:
you could probably add a PROCESSING_FLAG like the one used to determine which chems synthetics can process, but specifically for golems- would be a little snowflake code-y but ehh

though, can golems really not process any chems? I feel like that hasn't been the case in my experience.

Sylphily:
I have in fact, processed chems as a golem
I know this because I inject my self with medpen and I suddenly start to heal, so I would assume there is for some reason a black list for mutation chems for some reasons?

Ketrai:
There is a difference between blood processing and stomach processing. Maybe golems don't process ingested chems, but they do process blood-bound chems (aka injected chems)? Which is also a bit weird granted golems don't bleed, so it's like the rock itself process the chems.

Zevotech:
oh yeah lol #1443

PR #1401, opened 2022-07-30
Culls the osprey jobslots 

Sylphily:
## About The Pull Request

Culls the osprey jobslots to something more reasonable instead of like 10 some people on this ship.
Code_lrTzuclDyf

## Why It's Good For The Game

Reasonable crew sizes round start.
## Changelog

:cl:
tweak: Culls job slots on the osprey so that they are more reasonable
/:cl:

### DIFF
--- a/_maps/configs/ntsv_osprey.json
+++ b/_maps/configs/ntsv_osprey.json
@@ -24,20 +24,18 @@
 			"officer": true,
 			"slots": 1
 		},
-		"Scientist": 2,
+		"Scientist": 1,
 		"Medical Doctor": 1,
 		"Paramedic": 1,
-		"Engineer": {
+		"Ship Engineer": {
 			"outfit": "/datum/outfit/job/engineer/nt",
 			"slots": 1
 		},
-		"Atmospheric Technician": 1,
 		"Quartermaster": 1,
 		"Cargo Technician": 1,
 		"Shaft Miner": 1,
-		"Cook": 1,
 		"Janitor": 1,
-		"Assistant": 5
+		"Assistant": 2
 	},
 	"cost": 850
 }


Apogee-dev:
While i might be biased as the person that mapped the Osprey in the first place, I don't actually think the job slots are in any way excessive (although the ship itself does badly need some tweaks to give certain players more to do with their time). They take up a lot of space on the join menu due to the unusual diversity in jobs, but the Osprey (prior to this PR) has a designed capacity of 18 crew and a spawn limit of 1. Compare this to other ships of similar size with the same ship cap, such as the Skipper at 17 crew, the Shetland at 17 as well, and the Lagoon at 24 players (even if you don't consider the passengers, it still has 14 crew, which is greater than the reduced count in this PR). The reduced crew complement is actually fewer players than a Riggs or Carina, which are significantly smaller ships. Reducing the crew size just makes a spacious ship like the Osprey feel a bit empty even with a full complement.

Yellow-Mushroom:
I agree with apogee heavily here, especially with there being plenty enough room for two scientists and plenty of assistants. The osprey is THE large Nanotrasen ship as of current, with its main issues being an unavoidable one, if it doesn’t have enough people it really struggles in its niche both RP and game wise.

Sun-Soaked:
I'd have to agree with apogee here, and add that in the future it might be a good idea to have guidelines for crewsize to prevent confusion like this

Apogee-dev:
> I'd have to agree with apogee here, and add that in the future it might be a good idea to have guidelines for crewsize to prevent confusion like this

yeah, i'd agree there, but crew size is hard to put even loose guidelines on since it's ultimately dependent on a number of factors including ship size, crew equipment, and intended feel. It's ultimately one of those things you just need to get a feel for. The most I can say is that even the largest ships typically shouldn't go over 20 players unless it's for specific gimmick reasons like the lagoon's passengers.

MarkSuckerberg:
@Apogee-dev I'm leaving this PR entirely in your hands, merge or close.

Sylphily:
nah

PR #1437, opened 2022-08-07
Ports layershifting

Recoherent:
## About The Pull Request
Ports layershifting verbs from hrzntal/horizon#1202, and changes it so you can't layershift while incapacitated.
This comes with a config option to disable it for the server. This was present in the original PR, the difference being that it's enabled by default here instead.

## Why It's Good For The Game
have you ever tried to get your dude cozy under a blanket and then had your soul invert as you just layered on top of it with no way to fix it? me too, bestie. me too.

## Changelog

:cl:
add: Layershift verbs have been added to the IC tab. Finally, you can get cozy under blankets without relying on the fickle hand of rng!
/:cl:


ZephyrTFA (approved):
I updated the config to start disabled, config values should default to original behavior.

PR #1439, opened 2022-08-08
The Hindenburg(Balloon Alerts)

Sun-Soaked:
## About The Pull Request
Ports the useful by nature and much beloathed by implementation [balloon alert system ](https://github.com/tgstation/tgstation/pull/59130) from /TG/, as well as [visible balloon alerts](https://github.com/tgstation/tgstation/pull/59256).
Also ports elements of the [balloon alert mining implementation](https://github.com/tgstation/tgstation/pull/67816) as a test-drive of using these alerts to help sculpt down chat bloat. 

## Why It's Good For The Game
Provides us a very robust tool for a specific situation:
-Lots of messages in a short time
-No complex information conveyed
If we don't go completely crazy with it like /tg/ this should help cut down on chat spam.

## Changelog

:cl:
add: Balloon Alerts, which appear visually over a character's head and convey a short string of text before disappearing.
add: Area Balloon alerts, the above but visible to anyone nearby.
tweak: Digging into terrain now displays a balloon alert and does not chat log, to prevent spam.
tweak: Guns now do not trigger their "x fires y!" message for the user unless shot at point-blank.
tweak: The Kinetic Accelerator now no longer logs a *click* in chat when fired while on cooldown.
tweak: Reagent containers now balloon alert instead of chat logging when their reagent transfer amount is changed. Their transfer amount can now also be found in their description.
/:cl:

### DIFF
--- a/code/__DEFINES/layers.dm
+++ b/code/__DEFINES/layers.dm
@@ -133,6 +133,9 @@
 
 #define RUNECHAT_PLANE 20
 
+/// Plane for balloon text (text that fades up)
+#define BALLOON_CHAT_PLANE 21
+
 //HUD layer defines
 
 #define FULLSCREEN_PLANE 31
--- a/code/__HELPERS/game.dm
+++ b/code/__HELPERS/game.dm
@@ -331,6 +331,10 @@ block( \
 	O.screen_loc = screen_loc
 	return O
 
+/// Removes an image from a client's `.images`. Useful as a callback.
+/proc/remove_image_from_client(image/image, client/remove_from)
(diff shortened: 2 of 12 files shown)


ZephyrTFA (approved):
I dont have a "I'm Going To Merge This" label but pretend this is it

Apogee-dev:
My only objection is to reagent containers being removed from chat logging, because I'm an idiot and frequently lose track of what I was doing while doing bottlechem, and examining containers to verify transfer amounts is far worse for making chat unreadable.

PR #1457, opened 2022-08-14
Interdiction

ZephyrTFA:
## About The Pull Request

read the actual code. I'm done answering questions for people who can't read

## Why It's Good For The Game

Its something people have wanted forever

## Changelog

:cl:
add: Ship Interdiction
/:cl:


code/modules/overmap/interdiction.dm
```diff
@@ -0,0 +1,180 @@
...
+					c_x--
+				if(NORTHEAST)
+					c_y++
+					c_x++
+				if(SOUTHWEST)
+					c_y--
+					c_x--
+				if(SOUTHEAST)
+					c_y--
+					c_x++
```
MarkSuckerberg:
```suggestion
		if(dir & NORTH)
			c_y++
		if(dir & SOUTH)
			c_y--
		if(dir & EAST)
			c_x++
		if(dir & WEST)
			c_x--
```

code/modules/overmap/interdiction.dm
```diff
@@ -0,0 +1,180 @@
...
+	var/their_speed = target.get_speed()
+	if(their_speed >= 20) // this is gonna hurt, you and me
+		if(tgui_alert(user, "Radars indicitate they are going very quickly. Are you sure you want to do this?", "Interdiction", list("Yes", "No"), 10 SECONDS) != "Yes")
+			interdicting = null
+			target.interdictor = null
+			return
+
+		var/heading = target.get_heading()
+		throw_ship_contents(target, heading, their_speed * SPEED_THROW_STRENGTH_MULT)
+		throw_ship_contents(current_ship, turn(heading, 180), their_speed * SPEED_THROW_STRENGTH_MULT)
```
MarkSuckerberg: Maybe make this speed differential instead of just raw speed? So if you're both going fast it's easier &c

code/modules/overmap/interdiction.dm
```diff
@@ -0,0 +1,180 @@
...
+
+	point.name = "Interdiction Point"
+	point.token.icon_state = "interdiction"
+	target.Dock(point, force=TRUE)
+	tether = point.token.Beam(current_ship.token, time=INFINITY, maxdistance=50)
+	end_timer = addtimer(CALLBACK(src, .proc/end_interdiction), 20 SECONDS, TIMER_STOPPABLE|TIMER_OVERRIDE|TIMER_UNIQUE)
+	say("Interdiction now in effect for 20 seconds. Dismantle or de-powerment of this console will cause an early release.")
+
+	for(var/obj/announce_target as anything in list(target.helms[1], current_ship.helms[1]))
+		priority_announce("Interdiction is now in effect on target \"[target]\"", "Interdiction Tether Launched", 'sound/misc/announce.ogg', "interdiction", "Interdiction ([current_ship])", zlevel=announce_target.virtual_z())
```
MarkSuckerberg: can't you just use the virtual Zs of the two shuttle ports

code/modules/overmap/interdiction.dm
```diff
@@ -0,0 +1,180 @@
...
+	if(panel_open)
+		end_interdiction()
+
+/obj/machinery/computer/interdiction/proc/end_interdiction()
+	deltimer(end_timer)
+	if(!interdicting)
+		return
+	QDEL_NULL(tether)
+	for(var/obj/announce_target as anything in list(interdicting.helms[1], current_ship.helms[1]))
+		priority_announce("Interdiction is no longer in effect on target \"[interdicting]\"", "Interdiction Tether Dissipated", 'sound/misc/announce.ogg', "interdiction", "Interdiction ([current_ship])", zlevel=announce_target.virtual_z())
```
MarkSuckerberg: same here

MarkSuckerberg (changes requested):

Sylphily:
pirates but it actual works???? also the time for it seems a bit low, maybe 1 minute is enough?

Sun-Soaked:
consider adding some kind of gripping mechanic that relies on resources/fine-tuning the interdictor-
if there's a guaranteed short and exact "you will be able to get away" timer you can guarantee that people will respond to being interdicted by sitting on their console and refreshing over and over until the button goes blue

Getting onto someone else's vessel is "doable" (if barely) in 20 sec, but preventing them from jumping in that time is another story

best case scenario you're either 
a: forced to fight a man down as someone stays behind to take your vessel into pursuit
or b: have to fight all in because them jumping out with all your crew onboard _will delete your ship._

RuneQuester:
only 20 seconds is *really* shit. Like, 5 minutes seems better. Or have it scale off something, like engine count(stupid but, an idea is an idea). Smaller vessels will have longer interdiction times, bigger have shorter. 

Great idea though, and Im glad its being added.

Valorium-RB:
> only 20 seconds is _really_ shit. Like, 5 minutes seems better. Or have it scale off something, like engine count(stupid but, an idea is an idea). Smaller vessels will have longer interdiction times, bigger have shorter.
> 
> Great idea though, and Im glad its being added.

Seconded. That way it's easier for pirates to prey on small vessels (as they realistically would) and as they get more resources and can locate ships more easily by having more looted hard suits or jetpacks and whatnot, they can take on bigger ships as a form of natural progression.

genessee-forgot-his-password-again:
whenever i ran pirates and had admin-run interdictions, i usually had a five minute timer

it generally takes some time for attackers to figure out where the other ship actually is (if they're lucky, it takes about a minute) and then it takes another minute or so for everyone to prep up for an attack

20 seconds is, for the most part, nigh impossible unless you're incredibly lucky enough to land directly next to the ship you're attacking - imo, it's one thing to just be disorganized and unable to rally your troops in a timely manner, and it's another where you just get shit rng and are cheated out of a raid because of it

ZephyrTFA:
There is a giant blue beam from their ship to yours.

Sun-Soaked:
> There is a giant blue beam from their ship to yours.

this doesn't change the fact that you're effectively forced to fight without support or retreat- since it's guaranteed that they'll be out of there, likely with your fighters onboard their ship, before any meaningful fighting happens

beyond that, even getting on other people's ships in 20 secs is a *big* if- it'd require that they have an easily-accessible point of entry, and aren't actively taking point and defending the outside of their ship.

the meta you're encouraging is one of rapid high-destructiveness, high-lethality attacks, since that's the only way you'd reasonably be expected to take a decently crewed vessel in that timeframe. 

It also completely demolishes the idea of "surrender ye cargo" piracy rp standoffs that don't happen via. comms beforehand, since the time is so absurdly short that wordlessly attacking is pretty much the only option.

Zevotech:
> It also completely demolishes the idea of "surrender ye cargo" piracy rp standoffs that don't happen via. comms beforehand, since the time is so absurdly short that wordlessly attacking is pretty much the only option.

Pirates SHOULD be doing this, because this is ACTUALLY how pirates work.

ZephyrTFA:
> > It also completely demolishes the idea of "surrender ye cargo" piracy rp standoffs that don't happen via. comms beforehand, since the time is so absurdly short that wordlessly attacking is pretty much the only option.
> 
> Pirates SHOULD be doing this, because this is ACTUALLY how pirates work.

I'm going to have to ask that you actually read the PR contents.
The time has been changed, along with the entire method interdiction works.

ZephyrTFA:
Interdiction Start > Minigame where the ship can try to escape > Depending on how well they escape interdiction will last 30 or less seconds with 30 being the baseline.

ZephyrTFA:
The entire point of this being tested is so that I can adjust it as needed for balance

ZephyrTFA:
Nothing about this is set in stone.

triplezeta:
> > It also completely demolishes the idea of "surrender ye cargo" piracy rp standoffs that don't happen via. comms beforehand, since the time is so absurdly short that wordlessly attacking is pretty much the only option.
> 
> Pirates SHOULD be doing this, because this is ACTUALLY how pirates work.

theres plenty of time to make demands when you're interdicting them. in space. when both bridges are going to be fully staffed and able to attend comms.

Sun-Soaked:
> Nothing about this is set in stone.

Zephyr, we understand that- that's why we're suggesting increasing the timer
having actually looked at the code
here
image
and here
image
and in a few other places
it seems like the max time caps out at 30 seconds with a perfect interdiction, and can be reduced by factors like poor relative vessel speed and higher distance
Is that right?

Because if so it seems like all the stuff being said is still relevant to those conditions, give or take ten seconds of leeway

Might be a good strategy to unbrake it a bit and give vessels overtaking with a speed advantage bonus interdiction time beyond 30
that'd definitely help, but it'd still be mean time to be working in and a half-equipped ghetto tech crew could easily use a variety of methods to stall it out and uncounterably prevent boarding in that period

Sun-Soaked:
examples:
-RCD/girderspam your only two airlocks
-bolt exits
-drop/throw slipping items between your ships
-metal/slippery/dangerous foam spam the space between vessels
-ditto of above for dangerous gas grenades
-lube your front entryways
-Just Shoot At The Guys Coming To Your Ship and force them to take evasive action instead of trying to climb on your mobile starbase

any combo of these could make it functionally impossible to besiege a ship that can escape in sub-minute time without pretty much blasting their bridge open with a tank bomb and firing for 0 survivors

Calyxman:
F

Sun-Soaked:
stale bot kills again

PR #1502, opened 2022-08-27
Kilo Class Mining Ship QOL

Erikafox:
## About The Pull Request
Minorly remaps the Kilo so it's not as bad. It now does things like use functioning airlocks, have working conveyor belts, and fly forwards instead of backwards. 

## Why It's Good For The Game
The Kilo is less unintentionally bad.

dreamseeker_peZMfNOzCr

## Changelog

:cl:
tweak: The Kilo Class now flies to the right. You can now experience hitting bugs with your windows in real time.
add: The Kilo Class now uses dual port airlocks and AACs. 
add: The Kilo Class now has engine shutters
/:cl:


_maps/shuttles/shiptest/whiteship_kilo.dmm
```diff
@@ -1040,251 +1002,226 @@
...
-	},
-/obj/machinery/shower{
-	pixel_y = 18
+/obj/machinery/atmospherics/components/unary/vent_scrubber/on/layer4,
+/mob/living/simple_animal/hostile/netherworld/migo{
+	environment_smash = 0;
+	faction = list("neutral");
+	melee_damage_lower = 5;
+	melee_damage_upper = 10;
+	name = "Maurice"
```
tiramisuapimancer:
```suggestion
	name = "maurice"
```
Please keep his name lowercase, this was an intentional design choice when I added him

Erikafox: Alrighty.

Latentish (changes requested):
To start this is not a valid air cycler. While the doors do actually swap and automatically start cycles, it doesn't have enough air to refill. Moreover airlocks like this require the "high volume" versions of the dualport airvents (otherwise it takes a million years). Just one should do it for this.
The pipes ending this way is also extremely poor practice. If you're going to have the scrubbers network and the dual ports both shoot gas into space, you need to put injectors at the end.
I'd actually like to see you put at least some form of either air recycling or putting a dedicated air supply explicitly for this airlock. Otherwise it just doesn't work.
image

image
You also put an o2 canister connected to the vents network here instead of an air canister. o2 can be toxic as it's just pure oxygen. 

Did you mean for the captain's room door to be bolted? This ship looks like a sty so I assume so.

7. While the kilo doesn't have one to begin with, it wouldn't be bad to give the dedicated mining ship a mining equipment vendor. (make sure to put a real one not a refill and circuit as refills are getting scrapped at some point.)

Erikafox:
I'm putting an air recycler in, putting a gas pump on the injector to space there is, and a to do note to delete this ship in my documents folder. 
>The pipes ending this way is also extremely poor practice. If you're going to have the scrubbers network and the dual ports both shoot gas into space, you need to put injectors at the end.

They go to a layer adapter that formerly went straight to an injector.
>Did you mean for the captain's room door to be bolted? This ship looks like a sty so I assume so.

It's kilo soul.

Erikafox:
Discord_8R22bkRW0j
I agree with Apogee. It's far too small to have one that would work, and I'd prefer to not remodel it for a qol passover.

ghost:
one floater fire alarm on the south suit storage

Erikafox:
fixed, thanks.

Latentish (approved):

Latentish:
They call me maurice.
Marg??????

PR #1531, opened 2022-09-05
renames goon config file; adds ntsv prefix

ghost:
## About The Pull Request

adds the NTSV prefix to the Goon Shuttle since goon shuttle is technically an NT ship. also renames the config file to match with the map file name because easier to track goon shuttle files together then

## Why It's Good For The Game
i am cleaning :)

## Changelog

:cl:
config: goon-class shuttles now properly have their NTSV prefix; the config file name matches the map file name now
/:cl:

### DIFF
--- a/_maps/configs/nanotrasen_goon.json
+++ b/_maps/configs/nanotrasen_goon.json
@@ -1,5 +1,6 @@
 {
 	"$schema": "https://raw.githubusercontent.com/shiptest-ss13/Shiptest/master/_maps/ship_config_schema.json",
+	"prefix": "NTSV",
 	"map_name": "Goon-class Repurposed Emergency Shuttle",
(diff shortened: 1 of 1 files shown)


rye-rice:
you know, i never really saw the goon as a nt ship, thats why i didn't have the nanotrasen outfit subtypes for captain and engi set

nt deserves better ships than the goon

ghost:
> you know, i never really saw the goon as a nt ship, thats why i didn't have the nanotrasen outfit subtypes for captain and engi set
> 
> nt deserves better ships than the goon

fair! i can close this, but for your consideration, what if the goon had those swanky blue floors /jk

Sylphily:
imo it should be removed, its old, and there is pretty much a better ship that fits this niche, but thats my thoughts

PR #1541, opened 2022-09-08
Holy ship: a shepherd remake

PiperDoots:
## About The Pull Request

Remakes the shepherd-class from the ground up to comply with modern mapping standards!
Built from the ground up it's meant to give the shepherd a slightly changed purpose with a new brewery and beekeeping section at the cost of some engineering equipment. I aimed to keep the spirit of the original alive as much as possible so it's still recognizable as the shepherd just with a massive facelift.

## Why It's Good For The Game

The shepherd has been around for a while and had a lot of issues in both construction and balance. 
This remake fixes all of those and more.

image

## Changelog

:cl:
add: Remade the shepherd-class from the ground up
del: Removed the old shepherd-class
fix: Default chaplain is no longer SRM
/:cl:


ghost:
looks really good; only question is if it's possible to rotate the shortwave intercom and the two blast buttons up by the helm so they face towards whoever's at the helm instead of the south wall?

PiperDoots:
> looks really good; only question is if it's possible to rotate the shortwave intercom and the two blast buttons up by the helm so they face towards whoever's at the helm instead of the south wall?

Done! Updated the images to reflect the change.

Yellow-Mushroom (approved):

PR #1548, opened 2022-09-10
DIsables the Boyardee-B

Yellow-Mushroom:
## About The Pull Request

This PR disables (but doesn't delete) the Boyardee-B variant

## Why It's Good For The Game

The ship is a near identical copy of the Boyardee however, worse in most ways as its niche is "boyardee but hydroponics is significantly worse, making the kitchen significantly worse", with the kitchen being like, the main part of the boyardee

It also has a massive ration pack stock pile which, is the antithesis of the boyardee niche. Along with this, it has many errors with perspective, and its niche of boyardee but worse + guns being less than optimal, I feel as if it is better to disable this.

## Changelog

:cl:
del: Disables the Boyardee-B
/:cl:

### DIFF
--- a/_maps/configs/diner_b.json
+++ b/_maps/configs/diner_b.json
@@ -17,6 +17,5 @@
 			"outfit": "/datum/outfit/job/assistant/waiter/syndicate",
 			"slots": 2
 		}
-	},
-	"enabled": true
+	}
 }


Latentish:
Ship alts are generally bad to begin with. This one was on the chopping block for quite a while.

Calyxman:
I dunno I liked the syndicate theming but yeah it's really underwhelming in the only two aspects it has (weapons and food)

Latentish:
merge it before youb reerrgsagsdksd

PR #1556, opened 2022-09-14
Ports "Fix looping timers qdeleted in their callbacks"

tiramisuapimancer:
Co-authored-by: LemonInTheDark <58055496+LemonInTheDark@users.noreply.github.com>

## About The Pull Request
Ports https://github.com/tgstation/tgstation/pull/69879 because I was asked to do this. I don't understand what this PR is supposed to do but I was told it would be cool if I ported it so I did. Game compiles and runs with these changes added, not sure what to look for. Cool!

## Why It's Good For The Game
runtimes bad

## Changelog

:cl:
fix: some sort of timers runtimes?
/:cl:

### DIFF
--- a/code/controllers/subsystem/timer.dm
+++ b/code/controllers/subsystem/timer.dm
@@ -127,7 +127,9 @@ SUBSYSTEM_DEF(timer)
 		ctime_timer.spent = REALTIMEOFDAY
 		callBack.InvokeAsync()
 
-		if(ctime_timer.flags & TIMER_LOOP)
+		if(ctime_timer.flags & TIMER_LOOP) // Re-insert valid looping client timers into the client timer list.
+			if (QDELETED(ctime_timer)) // Don't re-insert timers deleted inside their callbacks.
+				continue
 			ctime_timer.spent = 0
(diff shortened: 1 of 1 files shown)


Yellow-Mushroom:
As of recently, when I let my local run for any extended amount of time (usually between 8-15 minutes or so?) I begin to get periodic runtimes about the timer callback, which I believe this is fixing. 

I’ve never payed close attention to it, but I believe it was at set intervals where the timer callback would break, as it would occur even when no major events were occurring/the join screen was running in the background if I’m remembering correctly

tiramisuapimancer:
I'll leave it up for a little bit and see what happens then.

out-of-phaze:
> As of recently, when I let my local run for any extended amount of time (usually between 8-15 minutes or so?) I begin to get periodic runtimes about the timer callback, which I believe this is fixing.
> 
> I’ve never payed close attention to it, but I believe it was at set intervals where the timer callback would break, as it would occur even when no major events were occurring/the join screen was running in the background if I’m remembering correctly

Uuunfortunately in my local testing this didn't seem to fix it, but maybe I fucked up my port somehow. It's five minutes after a ship despawns iirc.

PR #1558, opened 2022-09-16
Fixes display of suit icons on mobs

Vexylius:
## About The Pull Request

I've noticed that children of /obj/item/clothing/suit don't have any mob_overlay_icon and that caused them to not be displayed on mobs. How/why did that happen I do not know. I don't know how long it has been broken either, but that seems to fix it (tested on a randomgen human with jackets that were not previously displayed, also tested with random other items with mob overlay icons from other files as a sanity check and they are displayed too)

It's nice to see what you're wearing when you're wearing something.

## Changelog

:cl:
fix: Suits are now displayed on mobs(humanoids etc) when worn
/:cl:


triplezeta:
i literally just fixed this in #1557

Vexylius:
> i literally just fixed this in #1557

ah, my bad

PR #1573, opened 2022-10-11
Subships

The-Moon-Itself:
## About The Pull Request
Ports the parts of beestation/beestation-hornet#7152 that adds the framework for ships to land on top of each other and not break everything. A ship can only land on another ship if there's an open docking port on the mothership that's large enough for the subship. Here's a video of it in action on a modified dwayne-class:

https://user-images.githubusercontent.com/51838176/195471361-f9f0d145-d7c9-480e-ad4a-d18705f2590f.mp4

This system should be able to handle just about any orientation of ships on top of each other, such as ships landed across areas, multiple ships landed on a single ship, a single ship landed on multiple ships, a ship that is only partially landed on another ship, a ship that is partially landed on a ship that's partially landed on another ship, and so on. Just make sure that you never try to land a ship on itself.

Something to note for this is that ships remember what's underneath them via baseturfs, and there's a hardcoded check that will cause errors if a baseturf list grows over 10 entries long. Because ship turfs have typically 1-3 baseturfs, after about 3 ships stacked on top of each other things will start to break.

You can also make maps with subships on them, to do this, follow these steps:
1. make the subship as if it were a regular ship in its own map file
2. create a new /datum/map_template/shuttle subtype that points to your subship map, these datums can be found in code/datums/shuttle.dm
3. On your main ship, place a stationary docking port where you want your subship to be
4. Set the stationary docking port's roundstart_template var set to your new map_template/shuttle subtype
5. Set the dwidth, dheight, width, and height of the stationary port to be compatible with your subship, this must be set manually unlike mobile docking ports

You can put the stationary docking port anywhere on your map, as long as it's on the ship. You can have its bounding box hang off the side of your ship, but please try to keep the entirety of its bounding box within the bounding box of map file, otherwise subships landing on your main ship might accidentally clip through structures nearby your mainship, including virtual z level borders.

## Why It's Good For The Game
Subships allow for many more creative designs and interesting dynamics between and within ships, especially when a crew may need or want to split its attention between multiple locations at the same time, or to make interactions between ships easier when you just need to land a smaller vessel inside of the other, cutting out the need to travel through space turfs to get between two ships.

## Changelog

:cl:
add: Subships are now possible
code: Lots of large changes to ship code
/:cl:


code/datums/shuttles.dm
```diff
@@ -102,7 +108,119 @@
...
+			continue
+
+		var/list/sanity = islist(shuttle_turf.baseturfs) ? shuttle_turf.baseturfs.Copy() : list(shuttle_turf.baseturfs)
+		sanity.Insert(shuttle_turf.baseturfs.len + 1 - baseturf_length, /turf/baseturf_skipover/shuttle) //The first two are the "real" baseturfs, place above these but below plating.
+		shuttle_turf.baseturfs = baseturfs_string_list(sanity, shuttle_turf)
+
+	my_port.load(src)
+	cached_map = keep_cached_map ? cached_map : null
+
+//This is the best way I could think of to prevent stationary ports from loading here
```
tmtmtl30, line 158: just wanted to note that you should use a different solution here, as we discussed previously on the discord. i can't let you do this in good conscience

tmtmtl30, line 158: just wanted to note that you should use a different solution here, as we discussed previously on the discord. i can't let you do this in good conscience

code/datums/shuttles.dm
```diff
@@ -102,7 +108,119 @@
...
+/datum/map_template/shuttle/initTemplateBounds(list/bounds, init_atmos = TRUE)
```
tmtmtl30: just wanted to note that you should use a different solution here, as we discussed previously on the discord. i can't let you do this in good conscience

tmtmtl30 (changes requested):

code/datums/shuttles.dm
The-Moon-Itself: Yep, I’m still trying to fix the reported SGTs from the test merge, then I’ll redo my changes to map loading

The-Moon-Itself:
oh god oh fuck I think I really ruined everything with git. I might need to call in the nuclear option and make a whole new PR. Maybe a new fork repo to work off of too.

The-Moon-Itself:
Okay, I tried something, I'm hoping this fixed it

The-Moon-Itself:
nope, this is completely ruined

PR #1581, opened 2022-10-19
Sun-light traits for z-levels

### LINKED ISSUE #1421: [BUG] Solars are unreliable
## Round ID:
Not available.

## Testmerges:
Not available.

## Reproduction:
Solars are very unreliable. They are able to produce some small amounts of power when in space, but not on planets. This isn't the problem. The issue is that while in space, where solars are supposed to work, the solars only produce power about 1 minute per every 10 minutes. And at low amounts of power at that. The primary strength of solars is that while producing small amounts of power, they're placed all over a station and able to produce power consistently throughout the round without much disruption, and without much risk of other engines. This isn't a station based server, and ships should logically be able to just face their solar panels towards the sun. Solars should be made to function like station solars (not always producing power due to needing to change direction) while on planets, but always produce power while flying or docked in open space. Or based on proximity to the star in the sector. They just need a tweak and a bit of a buff in general.
In addition, solars reset and disconnect from the controller every single time the ship it is on docks or undocks.

martinlyra:
## About The Pull Request

This is a "minimal-effort" PR for now. It adds three traits available for z-levels. Amongst them being the one that enables the original asimuth code as default. With two others that favors more a static method: 
- `STATIC_EXPOSED`, exposed by default and does not check for azimuth, but neighbors in its `orange(1, src)` vicinity. 
- `STATIC_OBSCURED` that blocks out the sunlight entirely.

It was intended to fix #1421 for the ships in hard space, but not touching the planets case since the day/night controller seems to be absent from this codebase. A more invasive solution would be needed here.

## Why It's Good For The Game

Makes the solars more useful, and their power is maximized to their highest efficiency possible by default.

## Changelog

:cl:
tweak: Adjusted solar power to make use of sunlight traits for z-levels.
/:cl:

### DIFF
--- a/code/modules/power/solar.dm
+++ b/code/modules/power/solar.dm
@@ -139,7 +139,7 @@
 	obscured = TRUE
 
 	var/sun_type = virtual_level_trait(ZTRAIT_SUN_TYPE)
-	if(sun_type == STATIC_OBSTRUCTED)
+	if(sun_type == STATIC_OBSCURED)
 		return
 		// go straight to jail
 	if(sun_type == STATIC_EXPOSED)
--- a/code/modules/power/solar.dm
+++ b/code/modules/power/solar.dm
@@ -143,14 +143,15 @@
 		return
 		// go straight to jail
 	if(sun_type == STATIC_EXPOSED)
-		var/turf/T = loc
-		T = T && T.above()
-		if (!T)
+		var/turf/loc_turf = loc
+		loc_turf = loc_turf && loc_turf.above()
+		if (!loc_turf)
 			var/count = 0
-			for(var/turf/Tt in orange(1, src))
-				count += !isgroundlessturf(Tt)
-			if (count >= 9) return
-		else if(!isopenturf(T))
+			for(var/turf/other_turf in orange(1, src))
+				count += !isgroundlessturf(other_turf)
+			if (count >= 9)
+				return
+		else if(!isopenturf(loc_turf))
 			return
 		obscured = FALSE
 		return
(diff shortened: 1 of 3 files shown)


code/modules/power/solar.dm
```diff
@@ -138,6 +138,24 @@
...
+		return
+		// go straight to jail
+	if(sun_type == STATIC_EXPOSED)
+		var/turf/T = loc
+		T = T && T.above()
+		if (!T)
+			var/count = 0
+			for(var/turf/Tt in orange(1, src))
+				count += !isgroundlessturf(Tt)
+			if (count >= 9) return
```
Fikou: no single liners go straight to jail

code/modules/power/solar.dm
```diff
@@ -138,6 +138,24 @@
 /obj/machinery/power/solar/proc/occlusion_setup()
 	obscured = TRUE
 
+	var/sun_type = virtual_level_trait(ZTRAIT_SUN_TYPE)
+	if(sun_type == STATIC_OBSCURED)
+		return
+		// go straight to jail
+	if(sun_type == STATIC_EXPOSED)
+		var/turf/T = loc
```
Fikou: you must use descriptive variable names

code/modules/power/solar.dm
```diff
@@ -138,6 +138,24 @@
...
+		T = T && T.above()
+		if (!T)
+			var/count = 0
+			for(var/turf/Tt in orange(1, src))
```
Fikou: not Tt

PR #1604, opened 2022-11-09
Holofield Power Draw Increase + Subtype for mapping

Erikafox:
## About The Pull Request

Increases the power draw of the holofield generator by 10. Currently, holofields draw a pittance of power, 10 watts per tile plus 50 from each generator. For an example field of  4 tiles, that's 140 watts. That's not much, if you're running a single PACMAN at the lowest function you can sustain that. This leads to the fields just kinda staying on, and the ship staying open all the time. 
With a power draw of 500 per tile (plus 50 per generator) the holofields are a much more active draw. The example bay of 4 tiles now takes 2.1kw to run, which is still sustainable with a PACMAN, but not the absentminded turn the shield on and open the bay that it was. The max draw of the holofield is 4.1kw, for an 8 tile shield.

Holofields are maxed out at a length of 8, and the new atmos/strong variant maxes out at 20 (ideally I'd want 40 so it'd fit any ship), and draws a minimum of 2kw to stay running. This variant is intended for mapping ruins where a large airtight hangar is implied. At its max size, it will draw 12kw. 

- [x] I loaded and tested the things, and the balance implications seem fine to me.

## Why It's Good For The Game

More power management and another tool to use while mapping

## Changelog

:cl:
add: Strong Holofield subtype, for mapping large hangar bays
balance: the power draw of holofields has been increased tenfold, take care to not run them 24/7
spellcheck: the holofield circuits now have the same name as the machines
/:cl:

### DIFF
--- a/code/game/objects/items/circuitboards/machine_circuitboards.dm
+++ b/code/game/objects/items/circuitboards/machine_circuitboards.dm
@@ -390,8 +390,19 @@
 	needs_anchored = FALSE
 
 /obj/item/circuitboard/machine/shieldwallgen/atmos
-	name = "Atmospheric Holowall Generator (Machine Board)"
+	name = "Atmospheric Holofield Generator (Machine Board)"
 	build_path = /obj/machinery/power/shieldwallgen/atmos
+
+/obj/item/circuitboard/machine/shieldwallgen/atmos/strong
+	name = "High Power Atmospheric Holofield Generator (Machine Board)"
+	build_path = /obj/machinery/power/shieldwallgen/atmos/strong
+	req_components = list(/obj/item/stock_parts/manipulator/nano = 2,
+		/obj/item/stock_parts/micro_laser/high = 2,
+		/obj/item/stock_parts/capacitor/adv = 2,
+		/obj/item/stack/sheet/plasmaglass = 1,
+		/obj/item/stack/cable_coil = 5
+		)
+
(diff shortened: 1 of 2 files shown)


code/game/objects/items/circuitboards/machine_circuitboards.dm
```diff
@@ -390,9 +390,20 @@
...
 
+/obj/item/circuitboard/machine/shieldwallgen/atmos/strong
+	name = "High Power Atmospheric Holofield Generator (Machine Board)"
+	build_path = /obj/machinery/power/shieldwallgen/atmos/strong
+	req_components = list(/obj/item/stock_parts/manipulator/nano = 2,
+		/obj/item/stock_parts/micro_laser/high = 2,
+		/obj/item/stock_parts/capacitor/adv = 2,
+		/obj/item/stack/sheet/plasmaglass = 1,
+		/obj/item/stack/cable_coil = 5
+		)
```
ZephyrTFA:
```suggestion
	req_components = list(
		/obj/item/stock_parts/manipulator/nano = 2,
		/obj/item/stock_parts/micro_laser/high = 2,
		/obj/item/stock_parts/capacitor/adv = 2,
		/obj/item/stack/sheet/plasmaglass = 1,
		/obj/item/stack/cable_coil = 5,
	)
```

PR #1627, opened 2022-12-13
SubShips attempt 2

The-Moon-Itself:
## About The Pull Request
Accidentally destroyed my old PR for this, #1573, by completely botching a merge from master to the point that it was easier to make a whole new fork than try to save it, so here we are again. Here's the original description:

Ports the parts of beestation/beestation-hornet#7152 that adds the framework for ships to land on top of each other and not break everything. A ship can only land on another ship if there's an open docking port on the mothership that's large enough for the subship. Here's a video of it in action on a modified dwayne-class:

https://user-images.githubusercontent.com/51838176/195471361-f9f0d145-d7c9-480e-ad4a-d18705f2590f.mp4

This system should be able to handle just about any orientation of ships on top of each other, such as ships landed across areas, multiple ships landed on a single ship, a single ship landed on multiple ships, a ship that is only partially landed on another ship, a ship that is partially landed on a ship that's partially landed on another ship, and so on. Just make sure that you never try to land a ship on itself.

Something to note for this is that ships remember what's underneath them via baseturfs, and there's a hardcoded check that will cause errors if a baseturf list grows over 10 entries long. Because ship turfs have typically 1-3 baseturfs, after about 3 ships stacked on top of each other things will start to break.

You can also make maps with subships on them, to do this, follow these steps:
1. make the subship as if it were a regular ship in its own map file
2. create a new /datum/map_template/shuttle subtype that points to your subship map, these datums can be found in code/datums/shuttle.dm
3. On your main ship, place "subship dock" landmark in turf where you want the bottomleft corner of the subship's bounding box to be, you can also use the offset_x and offset_y vars on the landmark to offset this corner if you need to place the landmark somewhere else.
4. Set the "subship_template" var on the landmark to the path of your subship's map_template subtype
5. Optionally change the dir on the landmark to rotate the subship. for reference, NORTH is no rotation, EAST is a 90 degree clockwise rotation, etc.

You can put the stationary docking port anywhere on your map, as long as it's on the ship. You can have its bounding box hang off the side of your ship, but please try to keep the entirety of its bounding box within the bounding box of map file, otherwise subships landing on your main ship might accidentally clip through structures nearby your mainship, including virtual z level borders.

## Why It's Good For The Game
Subships allow for many more creative designs and interesting dynamics between and within ships, especially when a crew may need or want to split its attention between multiple locations at the same time, or to make interactions between ships easier when you just need to land a smaller vessel inside of the other, cutting out the need to travel through space turfs to get between two ships.

## Changelog

:cl:
add: Subships are now possible
code: Lots of large changes to ship code
/:cl:


ZephyrTFA (dismissed):
Subshuttles need to not use docks on the primary ship to function as this will cause issues. Use manual shuttle movement and landmarks on the main shuttle to calculate shuttle area movement

The-Moon-Itself:
> Subshuttles need to not use docks on the primary ship to function as this will cause issues. Use manual shuttle movement and landmarks on the main shuttle to calculate shuttle area movement

I've thought about how to do this, and I'm not really sure how to switch to manual shuttle movement without risking more issues. stationary docking ports are essentially just abstract markers to guide ships to their destination, I'm don't really see what issues arise from using them and how manual movement would not cause more issues. I would need to know exactly what kind of issue I'm trying to fix in order to implement manual movement in a way that actually fixes that issue.

PR #1628, opened 2022-12-14
Buffs the research point values produced by tesla coils

Zevotech:
## About The Pull Request
Increases the minimum research point gain for tesla coils from 1 to 10
Increases the minimum research point gain for the research variant tesla coils (use a screwdriver on the board of a normal tesla coil) from 12 to 50

At a standard 4 coils for a tesla, that is 40 points per shock. With 4 research coils it becomes 200 points per shock.

## Why It's Good For The Game
Players are now properly rewarded for building a working research tesla on a planet or attaching grounding rods and research coils to the sides of their ship and driving through electrical storms like a badass. Before, the research gain was so insignificant that it would take ~10 coils to achieve a research speed even approaching the research nanites. This balances out the risk/reward of playing with Edison's Bane to a nice, speedy research gain.

In testing, which was a pain in the ass because I had to build a tesla at least three times, this was able to accrue ~2500 points in 2-3 minutes with 4 coils. Given the risk of playing with high-voltage shocks, I see this as balanced.

## Changelog

:cl:
balance: power tesla coils now produce a minimum of 10 points per shock when linked to a server
balance: research tesla coils now produce a minimum of 50 points per shock when linked to a server
/:cl:


Sylphily:
You are doing the good work here sir!

ZephyrTFA:
probably too much of an increase, but it's not like anymore reads the changelog

Zevotech:
> probably too much of an increase, but it's not like anymore reads the changelog

as stated in the PR, this requires a massive effort on the part of someone decently versed in engineering to build and harness a tesla, or properly attach and set up tesla coils on platforms outside of the ship to do this. It simply makes it worth the effort. This was tested thoroughly and only really becomes a problem at around 100 points per shock.

PR #1629, opened 2022-12-14
Makes most hostile mining/jungle mobs unable to smash down metal walls in a single hit

Zevotech:
## About The Pull Request
changes the environment_smash of simple_animal/hostile/asteroid to ENVIRONMENT_SMASH_MINERALS, which makes most mining mobs (excluding the fugu of course) only able to break mineable walls in a single hit.
(Also does the same for simple_animal/hostile/jungle, and directly applies SMASH_WALLS to large mobs like the megarachnid, leaper and seedling.)

## Why It's Good For The Game
Mobs that were previously able to break open and depressurize a ship without warning that definitely should have been unable to do so:
- Wandering Hermits (Reason for this PR)
- Watchers
- Basilisks
- Polar Bears
- Wolves
- Any other mob derived from simple_animal/hostile/asteroid without a var for ENVIRONMENT_SMASH_WALLS

The only mobs which would reasonably be able to break walls in at least a few hits would be Goliaths, Ice Whelps and maybe polar bears. I am willing to change those to be able to smash walls in multiple hits if necessary but I think for the sake of sanity they should remain unable to.

## Changelog

:cl:
balance: Changed simple_animal/hostile/asteroid's environment_smash to ENVIRONMENT_SMASH_MINERALS
balance: Changed simple_animal/hostile/jungle's environment_smash to ENVIRONMENT_SMASH_MINERALS
add: ENVIRONMENT_SMASH_WALLS to the leaper, seedling and megarachnid
del: ENVIRONMENT_SMASH_MINERALS on mobs already derived from simple_animal/hostile/asteroid
/:cl:


ZephyrTFA:
interesting commit names

PR #1637, opened 2022-12-20
Mapgen fixes and speedups (ignore the branch name. I'm dumb)

tmtmtl30:
## About The Pull Request

Alters the structure of map/planet generation to squash some bugs and improve performance.

Previously, planet maps were generated by placing the ruin first, and THEN generating the turfs according to the map_generator datum. This has been adjusted -- now, turfs are generated WITHOUT objects such as mobs/flora, the ruin is placed, and THEN the objects are added (turfs are "populated"). In conjunction with the addition of needed AfterChange() calls to update the atmos adjacency of the generated turfs, this ensures that planet atmos acts correctly surrounding ruins.

When deleting reservations (such as the deletion of planets after undocking), all objects on the planet are rounded up in a list and qdeleted. Although this causes a small lag spike, it SHOULD prevent items from hanging out inside the edges of planets. 

There's a feature to change the default baseturf of a virtual level, ZTRAIT_BASETURF, that we now use. This should cut down on the instances where a ruin on a planet is blown up and there's space underneath (might still happen on asteroids, because the baseturf there is still space; I didn't want space turfs without space as their baseturf).

Overmap encounter areas aren't global anymore (they no longer have the flag UNIQUE_AREA). Don't fucking add the flag UNIQUE_AREA to anything that should have weather in it, because if that area gets added anywhere else that _actually respects the flag_ you'll end up with cross-planet weather, because weather code sucks. This didn't cause bugs before, because the flag wasn't respected; it will now.

The biome assoc list has been moved into the map generator datum, and all encounters now generate using a map generator that either uses a biome or replaces everything with a single turf. This prevents duplication of cave generation code and makes dynamic overmap object code slightly easier to understand.

Some systems have been altered to improve performance; many of these changes are rather small, like the changes to turf population (mob placement now uses a stack of recently-created mobs to check if there are any nearby, instead of checking everything within 12 turfs; I've yet to add ruin mobs to these stacks to avoid placing mobs near ruin mobs) or lighting objects (removed a single line that changed the color of the lighting object on init).

Starlight has been altered, so that small turf changes near space turfs don't need to check as many nearby turfs and so that large turf changes can be batched to prevent further recalculation. This is probably responsible for the biggest performance increase.

Smoothing groups are cached before sorting instead of after, to prevent sort calls on many atom inits; /tg/station uses a unit test to avoid needing to sort at runtime ever, but I couldn't figure out how to do that without larger changes or writing a unit test that attempted to instance every atom once, which would be an undertaking of its own.

Gas strings have been similarly altered, and now their interpretation defaults to copying from a cached, immutable version of the mix encoded by the string. This avoids the significant overhead caused by repeated calls to params2list(). Auxmos has a better solution to this, __auxtools_parse_gas_string(), but our current custom build of Auxmos doesn't support it.

There are a few other small changes that I'm probably forgetting about and you should yell at me to read my own fucking code and tell you what else I changed.

I still need to manually check each planet type to make sure they aren't fucked up, I should probably do some proper profiling comparisons.

## Why It's Good For The Game

Fewer weird bugs, things generate faster, better* code.

## Changelog

:cl:
fix: Ruins don't sometimes start in hard vacuum anymore; planet turfs now share atmos correctly.
fix: There hopefully shouldn't be any random stray objects sitting in the edges of planets anymore.
fix: Planets now (hopefully) have the correct baseturfs (more or less). When you bomb a ruin on a planet, it probably won't break through to space anymore.
refactor: Planet generation has been refactored, improving performance somewhat.
/:cl:


code/modules/overmap/objects/dynamic_datum.dm
```diff
@@ -296,31 +299,58 @@
...
 	flags_1 = CAN_BE_DIRTY_1
 	dynamic_lighting = DYNAMIC_LIGHTING_FORCED
 	sound_environment = SOUND_ENVIRONMENT_STONEROOM
 	ambientsounds = RUINS
 	outdoors = TRUE
 	allow_weather = TRUE
 
+/area/overmap_encounter/New(...)
+	if(area_flags & UNIQUE_AREA)
+		CRASH("Area [src.name] ([src.type], REF: [REF(src)]) created with flag UNIQUE_AREA! Don't do this! Weather will break!")
```
tmtmtl30, line 347: should probably be a unit test

Latentish:
let's fUCKING GOOOOOOOOOOOOOOI

ZephyrTFA:
@MarkSuckerberg my brain isnt large enough to review this

code/datums/map_zones.dm
```diff
@@ -416,15 +425,37 @@
...
 	for(var/turf/turf as anything in get_block())
-		//Reset turf
-		turf.empty(RESERVED_TURF_TYPE, RESERVED_TURF_TYPE, null, TRUE)
+		// don't waste time trying to qdelete the lighting object
+		for(var/datum/thing in (turf.contents - turf.lighting_object))
+			qdel(thing)
+			// DO NOT CHECK_TICK HERE. IT CAN CAUSE ITEMS TO GET LEFT BEHIND
+			// THIS IS REALLY IMPORTANT FOR CONSISTENCY. SORRY ABOUT THE LAG SPIKE
+
+	for(var/turf/turf as anything in get_block())
```
MarkSuckerberg: Any reason get_block() is called separately for each loop?

code/modules/overmap/objects/dynamic_datum.dm
MarkSuckerberg: I can probably make a unit test to load all encounters but it'll be slow as hell

MarkSuckerberg (commented):
Lots of much needed cleanup and improvement, this is just something I caught on an initial review and isn't particularly pressing

code/datums/map_zones.dm
tmtmtl30: nah

tmtmtl30:
maptainers want merge so merge it is

PR #1665, opened 2023-01-05
Adds mining hardsuits to the cargo catalogue

Zevotech:
## About The Pull Request
Adds a pack for two indie mining hardsuits and a separate, more expensive pack for one heavy mining hardsuit and a pair of jump boots.

## Why It's Good For The Game
Mining hardsuits really should be orderable from outposts, seeing as its the main thing to do on planets and asteroids. Also the heavy mining hardsuit looks really cool and I wanted it to be obtainable from cargo again. Drip.

## Changelog

:cl:
add: adds the heavy mining hardsuit and independent mining hardsuits to the cargo catalogue
/:cl:

### DIFF
--- a/code/modules/cargo/packs/spacesuit_armor.dm
+++ b/code/modules/cargo/packs/spacesuit_armor.dm
@@ -28,6 +28,15 @@
 					/obj/item/clothing/head/helmet/space/pilot/random)
 	crate_name = "pilot space suit crate"
 
+/datum/supply_pack/spacesuit_armor/mining_hardsuits_indie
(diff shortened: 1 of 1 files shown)


Apogee-dev:
I’ll just paraphrase what I said in gen-dev here for the sake of making it easy to find. Personally I think it would be better to make heavy mining hardsuits available as a ruin loot item and an occasional “elite” roundstart item for certain ships. We could even buff it back to its old stats (or just buff it a little bit) if we did that since we wouldn’t have to worry about balancing around the things being available almost at will and in potentially unlimited amounts.

PR #1680, opened 2023-01-12
Removes the 2 Katanas from the Ninja Shrine And Also Make It Use Properly Lit Up Tile Variants

Erikafox:
## About The Pull Request
It removes the 2 (REAL) Katanas from the Ninja Shrine.
Those are 40 force weapons with a block chance.
It has loot in the form of the dash sword.

Then we're having issues with ruins starting pitch black and the lighting fix should help that out.

## Why It's Good For The Game
Risk-Reward Balance?
Lighting Good

## Changelog

:cl:
balance: the two ninja shrine katanas are now replicas.
/:cl:


Latentish:
I'd much prefer you make one of the swords drop from the megarachnid rather than  removing the primary loot.

Latentish:
Honestly it might be better to make the megarachnid drop the sword rather than leaving it on the pedastal. You put something else on the pedastal instead. Otherwise this is fine. Since it leaves one of them.

Latentish:
I'd like other maptainers to give their take on this.

Apogee-dev:
Yeah I agree with latency here, the main loot should either be dropped by the megarachnid or the ruin should be redesigned such that you can't really get to the primary loot without fighting the megarachnid. At the moment it's still open enough that you can just yoink it.

Zevotech:
leave the MGR reference arm in and ill allow it

PR #1698, opened 2023-01-16
Makes most year dates accurate to the lore primer page.

TheNeoGamer42:
## About The Pull Request

Changes a few lines of code, that being the year lines on newly made records and PDAs, and makes it show 504 FS. 
"The year is 504 FS – the time since the formation of the Solar Confederation. Seven years have passed since the devastating set of conflicts known now as the Corporate Wars have come to a stop." - The Main Page / Primer of the shiptest lore hackmd
image

## Why It's Good For The Game

Date lore, I guess. And it slightly annoyed me.

## Changelog

:cl: 
tweak: Most machines now correctly believe it is 504 FS.
/:cl:

### DIFF
--- a/code/game/machinery/computer/medical.dm
+++ b/code/game/machinery/computer/medical.dm
@@ -472,7 +472,7 @@
 				var/counter = 1
 				while(active2.fields[text("com_[]", counter)])
 					counter++
-				active2.fields[text("com_[]", counter)] = text("Made by [] ([]) on [] [], []<BR>[]", authenticated, rank, station_time_timestamp(), time2text(world.realtime, "MMM DD"), GLOB.year_integer+540, t1)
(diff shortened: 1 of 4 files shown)


TheNeoGamer42:
tested all of them and despite adding notes to records being basically completely and utterly unused they now have the right date. feed me more instances of the _incorrect_ date of 2563 so I can burn them at the stake there's none left that use what these 4 instances used according to my search

Apogee-dev:
Why don't you make this a CURRENT_YEAR constant so it can be changed again in the future if ever necessary (for example, if somebody wants to add another year every time an IRL year passes)?

TheNeoGamer42:
If it needs to be changed, it can be changed manually again. I don't think making it tick up every IRL year matters much if at all, honestly. Plus it'd cause disparity between the document and ingame too.

PR #1699, opened 2023-01-16
fixes Fake Lattices

spockye:
## About The Pull Request
Currently, you cannot walk over fake lattices due to their density being set to "true". This PR changes that

## Why It's Good For The Game
Fake lattices are like lattices, BUT they are used for making things look good, like a decal. There is no real reason for them to be full objects!

## Changelog

:cl:
tweak: changed how fake lattices interact with the world around them  
/:cl:


spockye:
re-opening

PR #1711, opened 2023-01-19
Holographic signs and barriers have a limited lifespan

Apogee-dev:
## About The Pull Request

Puts an expiration timer on all holosigns and barriers, 5 minutes by default. A visible timer appears over the holosign when 60 seconds remain on its lifespan. Also adds a sound effect when holosigns despawn and a warning beep when they have 1 minute of lifespan remaining. Existing maps with pre-placed holosigns (e.g. Metis) had them replaced with the infinite versions as a temporary measure to avoid breaking anything.

## Why It's Good For The Game

Discourages the use of holosigns and especially holofirelocks as a permanent fixture replacing normal airlocks entirely and encourages their intended use as a temporary measure while carrying out repairs.

## Changelog

:cl:
balance: Holosigns and barriers now have a limited lifespan. Countdown timer becomes visible at 60 seconds remaining.
/:cl:


Sun-Soaked:
I have reservations about this, but if we eventually get timerless inflatable walls it should end up working out(inflatables aren't able to be used as ghetto doors due to being sold, so there'll be a use case for both)

ZephyrTFA (approved):

Apogee-dev:
Alright, should be ready to go for later this week. Main thing I’d like to get a good read on are the actual lifespan values, whether they need to be reduced or extended. Mostly for the atmos barrier.

ZephyrTFA (approved):

code/game/objects/structures/holosign.dm
```diff
@@ -8,22 +8,49 @@
...
+
+	if(death_time < world.time)
+		playsound(src, 'sound/effects/empulse.ogg', 50, TRUE)
+		qdel(src)
+
 /obj/structure/holosign/Destroy()
 	if(projector)
 		projector.signs -= src
 		projector = null
+	qdel(countdown)
```
tmtmtl30, line 53: should be QDEL_NULL()

code/game/objects/structures/holosign.dm
```diff
@@ -8,22 +8,49 @@
...
+	var/obj/effect/countdown/holosign/countdown
 
 /obj/structure/holosign/New(loc, source_projector)
 	if(source_projector)
 		projector = source_projector
 		projector.signs += src
 	..()
 
-/obj/structure/holosign/Initialize()
+/obj/structure/holosign/Initialize(mapload, source, new_lifespan)
```
tmtmtl30, line 24: any point to new_lifespan here? doesn't look like it gets passed in

code/game/objects/structures/holosign.dm
```diff
@@ -8,22 +8,49 @@
...
+		if(new_lifespan)
+			lifespan = new_lifespan
+		death_time = world.time + lifespan
+		countdown = new(src)
+		if(countdown_color)
+			countdown.color = countdown_color
+		countdown.start()
+
+/obj/structure/holosign/process()
+	if(death_time < (world.time + 60 SECONDS) && death_time > (world.time + 58 SECONDS)) //there has to be a better way to get this to just trigger once
```
tmtmtl30: any reason not to do something like 
if(countdown.invisibility && death_time < (world.time + 60 SECONDS))
instead?

code/game/objects/structures/holosign.dm
tmtmtl30: actually i'd probably add a Destroy() proc for holosigns here that looks like
/obj/effect/countdown/holosign/Destroy(...)
  if(attached_to)
    var/obj/structure/holosign/H = attached_to
    if(H.countdown)
        H.countdown = null
  return ..()
that way holosigns clean up their ref to the countdown on deletion and the countdown cleans up its ref to the holosign regardless of whether it's deleting because the holosign is (it pays to be safe)

PR #1704, opened 2023-01-18
fixes spline corners always being white

BarteG44:
## About The Pull Request
Fixes the spline corners always being white by switching two words around 

before
image
image

after
image
image

- [X] I affirm that I have tested all of my proposed changes and that any issues found during tested have been addressed.

## Changelog

:cl:
fix: fixed spline corners always being white

/:cl:

PR #1714, opened 2023-01-19
Reworks the box class

BarteG44:
## About The Pull Request

Modernizes the box class by replacing the tiny fans with holofields, makes the rear airlocks actual airlocks and changes the look while keeping the same general layout

image

## Why It's Good For The Game
The document said that the box needed a touch-up so as the dedicated sector doctor i was obligated to do it

## Changelog

:cl:
tweak: Reworked the box class
/:cl:


Erikafox (commented):
image
I would move the scrubber down about 4 tiles, having it on top of other pipes when it doesn't need to be sacrifices a bit of clarity
StrongDMM_kW04x37x5F
The bodybag spawns outside of the thing
dreamseeker_DcUXNv1hi6
use old blood here

Not entirely sold on having the ship start mostly unlight, but another mappy can weigh in on that. It's not a dealbreaker.
Engineering needs more lights/lightmounts. It's fairly dark once you get the one light installed.
I would put lights in the rear airlocks too.
StrongDMM_0UbA3YOF6D
I would keep this as a normal table. The rolly table will get moved around in times of high traffic.

BarteG44:
the bodybag is intentional along with engineering being dark but i can lighten it up

Erikafox:
dreamseeker_T2AMS0QlSg
There's a good line between kinda dark and "I need a flashlight if I want to go into engineering.
Roger on the bodybag

The engines need shutters.

BarteG44:
well engineering needs to have bulbs inserted too to have lights. i went with the empty light sockets becouse the original also required you to insert them and i thought it's a somewhat interesting way to start a round

Erikafox:
Yeah I see those now.
Consider marking them in some other manner than just being there.

BarteG44:
>Consider marking them in some other manner than just being there.

image
will that do?

Erikafox:
the arrows will render under the catwalk. I'd just make them start with lights installed honestly.

BarteG44:
most of it should be lit now but to have full brightness you'll need to put in the rest of the bulbs yourself

BarteG44:
I would appreciate it if #1704 got merged before this so i can fix some of the decals without making a new PR

spockye (changes requested):
image
please pixelshift your wall-lockers! if there isn't space for them here, move them somewhere else. 

image
please move the APC or the poster!

image
i'd recommend pixel shifting the organ storage from the tile above, then setting the density to 0. it looks a bit out of place from the crew area

though this is definitely an improvement!

BarteG44:
>i'd recommend pixel shifting the organ storage from the tile above, then setting the density to 0. it looks a bit out of place from the crew area

image

*will do*

BarteG44:
image
my beloved

Erikafox (approved):

BarteG44:
the splinefix PR broke one of the newer ruins apparently

BarteG44:
>I would keep this as a normal table. The rolly table will get moved around in times of high traffic.

funny thing about these. i removed it from the ship not becouse they were annoying or moving around but rather becouse anything placed on them will stay behind when you undock the ship.

spockye (approved):
looks good!

PR #1715, opened 2023-01-19
Slice blood bag to transfer in other stuff

ritorizo:
## About The Pull Request

You can now slice the blood bag open with something sharp to pour/splash out it's content.
It's impossible to attach a sliced blood bag to an IV drip.
It's impossible to pour reagents *into* the bag, otherwise it would become an easy 200u beaker.

### More technical part 
The current definition of the blood bag has it's own type of reagent container (so `/obj/item/reagent_containers/blood`) which doesn't support pouring.
So what this pr does is change the blood bag from  `/obj/item/reagent_containers/blood` to  `/obj/item/reagent_containers/glass/blood` which support pouring and stuff.
So I had to change the definition of the blood bag everywhere in the code which I did with the following command
 - ```git grep -l 'reagent_containers/blood' | xargs sed -i 's|reagent_containers/blood|reagent_containers/glass/blood|g'```

Yes still work as intended and generate in structures.

## Why It's Good For The Game

It's a feature that a lot of people requested.

## Changelog

:cl:
added: can now slice a bloodbag open to pour stuff
/:cl:


ritorizo:
Update on possible issues, redefining the blood bag will not mess with stuff that is already merged but I wondered if it could mess with some current map pull request.
Turns out all pr use prefilled freezer/fridge for their blood bags so it won't cause any issue. (yes I checked every map PR)

PS: If someone end up still finding an issue caused by this PR I will solve it as quickly as I can

code/modules/reagents/reagent_containers/blood_pack.dm
```diff
@@ -1,21 +1,42 @@
...
+//	'git grep -l 'reagent_containers/blood'
+//		Get the path of all the files that contain the string 'reagent_containers/blood'.
+//
+//	'|'
+//		take the output of the last command to the next
+//
+//	xargs sed -i 's|reagent_containers/blood|reagent_containers/glass/blood|g'
+//		In all the files which path are inputed
+//		replace "reagent_containers/blood"
+//		by "reagent_containers/glass/blood"
```
ZephyrTFA: Get rid of this

Ms-Mee: Fearful react

PR #1716, opened 2023-01-19
Add the angle grinder, a tool to deconstruct structures quickly

ritorizo:
## About The Pull Request

The angle grinder is a powertool which is used to deconstruct things quickly, it's intended for salvaging structures.

It curently works on :
 - walls (not reinforced ones)
 - girders
 - grille
 - catwalks
 - lattices
 - crates/closets
 - tables
 - reinforced windows
 - bed
 - chairs

It require eye protection to be used like welding tools.
It also works as a getho bone saw (at 80%)

It's currently only obtainable in the outpost.

## Sprite (thanks ~~NIRV-0025~~ NIRV-0026)
image

## Why It's Good For The Game

The idea was to add another way of getting resources than mining ores (maybe to make ore processing harder at some point) and just be a building tool.

More feature will be added eventually but it could be cool to test it early for feedback (like what stuff to add, buff, debuff...)

## Changelog

:cl:
added: The angle grinder, a quick deconstruction tool.
added: A sprite for the angle grinder (**made by NIRV-0024**)
/:cl:


TheNeoGamer42:
Could make this work on the broken machines found on wasteworlds as a quicker alternative to the crowbar you have to use on them.

Calyxman:
Will you be adding it into Industrial Engineering or Advanced Tools as an engineering/supply protolathe item?
It'd be nice if you could make one of these, I think

Apogee-dev:
Needs availability in the Engineering and Supply lathes for sure. You should also allow it to cut down airlocks (either destroy the boards/wires or require the user to switch to crowbar/cutters for those), that’s kind of a critical requirement for a salvage/deconstruction tool

Apogee-dev:
> Could make this work on the broken machines found on wasteworlds as a quicker alternative to the crowbar you have to use on them.

Doesn’t really make sense, a grinder would reasonably just destroy whatever valuable components are actually inside the derelict machines.

CenkTheRaptor:
spacr emgineers

Sylphily:
> > Could make this work on the broken machines found on wasteworlds as a quicker alternative to the crowbar you have to use on them.
> 
> Doesn’t really make sense, a grinder would reasonably just destroy whatever valuable components are actually inside the derelict machines.

i would quite argue that, they could just grind around the components that they want.

Apogee-dev:
> > > Could make this work on the broken machines found on wasteworlds as a quicker alternative to the crowbar you have to use on them.
> > 
> > 
> > Doesn’t really make sense, a grinder would reasonably just destroy whatever valuable components are actually inside the derelict machines.
> 
> i would quite argue that, they could just grind around the components that they want.

it’s an industrial angle grinder, not a dremel tool. this thing takes two hands to use and saws through girders, it really doesn’t seem like something capable of precision work in the slightest.

TheNeoGamer42:
That doesn't stop John Space Engineers though, and in a few ways spessmens are capable of even more than them

ritorizo:
I think we could argue both way for the logic of salvaging circuits, I think we should think in term of balance

ritorizo:
I already coded the feature but we could still revert it later

Apogee-dev:
> I think we could argue both way for the logic of salvaging circuits, I think we should think in term of balance

well, my take from a balance and design perspective is that while having a tool that massively simplifies and expedites salvaging is good, it isn’t necessarily engaging to have a single tool capable of doing _everything_ on its own. Part of the original conversation that led to the creation of the grinder was that salvage is too slow and unrewarding compared to mining, as far as raw resources like metal sheets go. That’s true, and the reason is specifically because dismantling walls and airlocks and the like takes a long time and only yields a few sheets of metal each. And airlocks aren’t even remotely worth it because you have to hack them first. Machines, meanwhile, are more of a source of boards and components than they are of raw materials, and they’re higher value for time spent anyway. Plus the broken rockplanet ones are just a single-step dismantle with a crowbar anyway, it’s not a long process.

BarteG44:
> Needs availability in the Engineering and Supply lathes for sure.

IMO it shouldn't be aviable from a lathe at all. Only salvage ships should start with / it should be available at the outpost. Being able to deconstruct pretty much anything is a very powerful ability and requiring only research to get it would let the not specialized ships do salvage just as easily as mining nullifying one of the points of the original conversation

ZephyrTFA:
I don't really like JackOfAllTrade tools; and as such do not really like this

axelzonvolt:
it's not even really a "jack of all trades" tool. it does exactly one thing very well, but doesn't help with anything else.

Unless of course Sylphily's suggestion of making it magically capable of not turning valuable circuit boards into recycled material is included, which WOULD make it broken instead of a fun alternative.

Sun-Soaked:
I think this is fine if for two changes:
-Remove the ability for it to destroy walls, even unreinforced ones. This is a special thing we should reserve for things like breaching hammers and explosives.
-Get rid of the ability to tear down reinforced windows. Those should be a mild challenge to destroy.

otherwise a quick and dirty ruin stripper tool seems like an alright idea, at least for now

Sun-Soaked:
with regards to the "salvaging" thing slyphily said, it could have a custom interaction with salvage objects where it quickly destroys them but only drops a table of raw materials, rather then the expensive internal components.

goober3:
> it's not even really a "jack of all trades" tool. it does exactly one thing very well, but doesn't help with anything else.
> 
> Unless of course Sylphily's suggestion of making it magically capable of not turning valuable circuit boards into recycled material is included, which WOULD make it broken instead of a fun alternative.

could be further cemented as strictly a Salvaging Tool by, instead of providing loot, dropping a Considerable Amount of Metal and Glass and Nothing Else from ruined machinery.

goober3:
> with regards to the "salvaging" thing slyphily said, it could have a custom interaction with salvage objects where it quickly destroys them but only drops a table of raw materials, rather then the expensive internal components.

holy shit we said the Same Thing???

axelzonvolt:
> > it's not even really a "jack of all trades" tool. it does exactly one thing very well, but doesn't help with anything else.
> > Unless of course Sylphily's suggestion of making it magically capable of not turning valuable circuit boards into recycled material is included, which WOULD make it broken instead of a fun alternative.
> 
> could be further cemented as strictly a Salvaging Tool by, instead of providing loot, dropping a Considerable Amount of Metal and Glass and Nothing Else from ruined machinery.

this is... literally the entire point of ritorizo's PR. only a commenter is making the suggestion for it to ignore circuit boards when tearing stuff apart, which doesn't jive with the whole "grind an entire object/wall/window into salvageable materials."

TheNeoGamer42:
I think it should be able tokill unreinforced walls just fine. Look at the fucking plasma cutter, that thing can literally cut the wall down, and then cut the girder down, zero issue. AND the plasma cutter has even more uses. Keeping this being able to kill walls is perfectly reasonable if you want an actually decent salvaging tool, unless we want to suddenly keep it to the plasma cutter, which is basically one of the best mass mining tools around.

Apogee-dev:
> I think this is fine if for two changes: -Remove the ability for it to destroy walls, even unreinforced ones. This is a special thing we should reserve for things like breaching hammers and explosives. -Get rid of the ability to tear down reinforced windows. Those should be a mild challenge to destroy.
> 
> otherwise a quick and dirty ruin stripper tool seems like an alright idea, at least for now

part of the whole point of adding it, at least in the original conversation about it, was that dismantling ruins for raw materials- mostly the walls- takes absolute ages and is a waste of time compared to mining even with improvised picks. If you want to balance it out, just make it take the same time or slightly longer to do so as a jackhammer so the proper breach tools are still the fastest option. It’ll still be way faster and less clunky than dismantling a wall using power tools.

ritorizo:
For the decon wall issue : it's the whole point of the tool to salvage wall and I don't think it could cause issue in pvp since a welder can do about the same job as an angle grinder carrying one is a huge disadvantage (even tough the grinder works as a chainsaw for combat but maybe we should change this) since it can't even fit in a backpack.

For the decon machine issue : we could make it so when you grind a machine it gives the 5 metals and a "material agregate" which would be an item that contain the matter worth of the part of the machine (boards included).

ritorizo:
I think I'll implement all controversial feature (machine salvaging, rnd access...) anyway but put a build option do toggle them

Apogee-dev:
> For the decon wall issue : it's the whole point of the tool to salvage wall and I don't think it could cause issue in pvp since a welder can do about the same job as an angle grinder carrying one is a huge disadvantage (even tough the grinder works as a chainsaw for combat but maybe we should change this) since it can't even fit in a backpack.
> 
> For the decon machine issue : we could make it so when you grind a machine it gives the 5 metals and a "material agregate" which would be an item that contain the matter worth of the part of the machine (boards included).

walls: just make sure the do_after for grinding down a wall is at least as long as the one on jackhammers and the balance problem becomes a non-issue

machines: it’d probably be better to just drop the raw materials (or the better yet, “scrap” materials that were added with the junk planet machines like scrap circuitry etc) rather than adding a new aggregate item just for this purpose. Destroying components in this way should probably mean losing some materials over careful disassembly.

Also i realize this should have been said much sooner but please include a picture of the sprites in your pull request

ritorizo:
sprite 👍

TheNeoGamer42:
sprite... oh yeah this uses the circular saw sounds doesn't it? it should totally use a louder actual grinder sound. you should KNOW when someone is using this.

ritorizo:
I will look into it (even tough I don't know what the ss13 sound standards are)

ritorizo:
I will look into it (even tough I don't know what the ss13 sound standards are).
Also we still need in hand sprites.

ritorizo:
The code is shipped and ready to be tested, important changelog since the last update :
- Added a sound for the grinder
- Ear damage if you don't wear protection (also you can now wear the earmuff on your neck)
- Deconstruct machine **frames** (so you still need to carry a screwdriver and a crowbar but it's easier to recover steel from stuff)
- Can now configue pretty much everything in code/__DEFINE/anglegrinder.dmm (ex: change speed, price, disable feature...)

Apogee-dev:
Looking forward to the Gecko remap after this. Specialization and earmuffs, let's go.

code/__DEFINES/anglegrinder.dm
```diff
@@ -0,0 +1,8 @@
+// Config, comments shits out to disable features
+
+#define ANGLEGRINDER_OUTPOST_PRICE 1000
+// #define ANGLEGRINDER_BROKEN_MACHINE_SALVAGE_TIME 2.5 SECONDS
+#define ANGLEGRINDER_OPEN_SAFE_TIME 90 SECONDS
+#define ANGLEGRINDER_OPEN_CLOSET_TIME 5 SECONDS
+#define ANGLEGRINDER_RAILING_TIME 1 SECONDS
+#define ANGLEGRINDER_MACHINE_FRAME_TIME 3 SECONDS
```
ZephyrTFA, line 9: These should be config entries not defines.

ZephyrTFA (changes requested):

code/__DEFINES/anglegrinder.dm
ritorizo: Honestly I don't know how to do all of this cleanly using config without making some wacky code or hurting performance. I could have also hard coded all the value as it's usually done, but it would make balancing and tweaking of the feature harder.

Erikafox:
does this still need inhands

ritorizo:
I didn't see it get used that much while it was test merged but it shouldn't break tremendously if it end up a having an issue.
We could merge it so mappers can put it in their ship and and give feedback so I can make some improvement (re balancing, feature idea...) pr afterward.

code/modules/mining/equipment/angle_grinder.dm
```diff
@@ -0,0 +1,94 @@
...
+ * code/game/objects/structures/bed_chairs/chair.dm
+*/
+
+/*
+ * Configure features by editing __DEFINES/anglegrinder.dm
+*/
+
+/obj/item/anglegrinder
+	name = "angle grinder"
+	desc = "A powerfull tool to salvage about anything. You better wear some protection tough"
```
Zevotech: minor spelling/grammar mistakes. "A powerful tool to salvage just about anything, you better wear some protection though."

Preferably, I would replace that second part with something that outright states what is protection. "...anything, you should wear earmuffs and welding protection to avoid damage to your ears and eyes." would be my best guess but something longer might be better.

code/modules/mining/abandoned_crates.dm
```diff
@@ -1,4 +1,4 @@
-//Originally coded by ISaidNo, later modified by Kelenius. Ported from Baystation12.
+//riginally coded by ISaidNo, later modified by Kelenius. Ported from Baystation12.
```
Zevotech: I dont know what the O in Originally did to deserve being deleted but you should probably fix it.

Erikafox:
Resolve the merge conflict please.

PR #1722, opened 2023-01-20
Painting fixes

### LINKED ISSUE #1625: [BUG] When unsecuring painting frames, the painting contained within is obliterated
## Reproduction:
Step 1. Spend 40 minutes on a painting
Step 2. Place a painting frame on wall
Step 3. Place painting into frame
Step 4. Try to unwrench the painting frame
Step 5. Suffer, as now not only does the painting frame lose its icon until found and placed on a wall again it obliterates the painting of Capricornus YYZ you did for 40 minutes.

MarkSuckerberg:
## About The Pull Request
Fixes: #1625

Also should make it so that all painting frames save and load even outside of the few libraries that exist.

## Why It's Good For The Game
Paintings take quite a while, and it must suck when they get completely obliterated. This fixes that as well as tweaks it so more paintings are saved.

## Changelog

:cl:
fix: Paintings are no longer deleted when their frames are unwrenched.
tweak: Paintings should save even if they're outside of libraries.
/:cl:

### DIFF
--- a/code/game/objects/structures/signs/_signs.dm
+++ b/code/game/objects/structures/signs/_signs.dm
@@ -61,6 +61,9 @@
 
 /obj/structure/sign/wrench_act(mob/living/user, obj/item/wrench/I)
 	. = ..()
+	//If it's not buildable, just make them hit it with the wrench.
+	if(!buildable_sign)
+		return FALSE
 	user.visible_message(
 		"<span class='notice'>[user] starts removing [src]...</span>", \
 		"<span class='notice'>You start unfastening [src].</span>")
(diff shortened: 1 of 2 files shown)


code/game/objects/structures/signs/_signs.dm
```diff
@@ -61,6 +61,9 @@
 
 /obj/structure/sign/wrench_act(mob/living/user, obj/item/wrench/I)
 	. = ..()
+	//If it's not buildable, just make them hit it with the wrench.
+	if(!buildable_sign)
+		return FALSE
```
MarkSuckerberg, line 66: This is how I was originally going to fix it, but then I realised that there was no way to remove the painting frame without smashing it. That's annoying, so I fixed it on the painting type itself, but I left this in as the comment on this var says that if this var isn't true, you shouldn't be able to unwrench it.

PR #1736, opened 2023-01-25
Adds double filing cabinets

Apogee-dev:
## About The Pull Request
doublecabinets

Adds a couple new filing cabinet types with a double sprite so mappers can stop illegally stacking multiple cabinets on the same tile that break as soon as either cabinet is moved or destroyed.

## Why It's Good For The Game

solves some mapper sins

## Changelog

:cl:
imageadd: added double variants for filing cabinets
/:cl:


BarteG44:
Do a triple too :)

PR #1741, opened 2023-01-30
Remove the message when cutting log (towercap, tree, steelcap...)

ritorizo:
## About The Pull Request

Remove the message when turning wood into plank

## Why It's Good For The Game

It's pretty much useless information and it clog up the chat.

## Changelog

:cl:
del: Removed the message when cutting log into plank.
/:cl:

### DIFF
--- a/code/modules/hydroponics/grown/towercap.dm
+++ b/code/modules/hydroponics/grown/towercap.dm
@@ -53,7 +53,6 @@
 
 /obj/item/grown/log/attackby(obj/item/W, mob/user, params)
 	if(W.get_sharpness())
-		user.show_message("<span class='notice'>You make [plank_name] out of \the [src]!</span>", MSG_VISUAL)
 		var/seed_modifier = 0
 		if(seed)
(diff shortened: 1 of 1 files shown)


Erikafox:
I don't like this

If you really want it out of the chatlog make it a bubble alert or something

ritorizo:
There is two message, the first one that say "You make plank out of the log" and the other one that also indicate how much plank you are making, and how large is the new stack of plank.

The seconds one is really the one that is bothering me since the first one can just stack since it's always the same.

Would it be fine if we just remove the second one and maybe add how much plank you are making in the first one ?

Ms-Mee:
I'm pretty sure the stack one is for consistency with stacks, though a personal annoyance still isn't much a reason to remove something, I do have an idea I could make pretty quickly as to avoid the spam if it really is that much a problem

Sun-Soaked:
this could probably be made into a "chopping"/"chopping wood..." bubble alert, if anything. I don't mind the current logspam since it generally ends up being collapsed into message stacks

PR #1758, opened 2023-02-10
potentially fixes incorrect ruin turf icon_states

tmtmtl30:
## About The Pull Request

ruin turfs (specifically dirt, but also some others) were occasionally generating in ruins with incorrect icon states (either plating or blank tiles). i'm not entirely sure why that happens, but my guess is that it involved an odd race condition where update_icon() was called on turfs that hadn't had Initialize() called yet, causing icon_state and icon_plating to get stuck in "plating" (which, if a turf's icon file doesn't include a "plating" state, would lead to a blank appearance). making icon_plating default to null, and checking to see if it isn't null before swapping to it, should hopefully defuse this bug

## Why It's Good For The Game

ruins that look how they're supposed to look are better than ones that don't

## Changelog

:cl:
fix: hopefully fixed certain incorrect ruin floor turf icons
/:cl:

### DIFF
--- a/code/game/turfs/open/floor.dm
+++ b/code/game/turfs/open/floor.dm
@@ -20,7 +20,8 @@
 	intact = TRUE
 	tiled_dirt = TRUE
 
-	var/icon_plating = "plating"
+	// initiailized as null to stop turfs that update_icon() before Initialize() (in late ruin loading) from getting "stuck" as plating
+	var/icon_plating = null
 	var/broken = FALSE
 	var/burnt = FALSE
 	var/floor_tile = null //tile that this floor drops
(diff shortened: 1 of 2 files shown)


tmtmtl30:
pushed a fix that might fix the stair icon issue -- /turf/open/floor instances were inheriting their dir from a previous instance of /turf/open/floor on the same tile. this is likely to cause some strange behavior with other dir-based floor turfs (if you remove it, then place it back, its dir will have reset, when it'd be sticky before) but i'd like to at least know if it works and then i can think about what a better solution might look like

Erikafox:
This has worked for about a month of uptime, merge please

PR #1760, opened 2023-02-11
Die Of Fate Change

Erikafox:
## About The Pull Request
replaces the die of fate's d20 effect (spawn you as wizard) with spawn wizard clothes and magic mirror under you.

## Why It's Good For The Game
I'm sick of wizards spawning without admin intervention

## Changelog

:cl:
  balance: You can't be turned into a wizard by the die of fate, instead getting a magic mirror and wizard clothes.
/:cl:

### DIFF
--- a/code/modules/awaymissions/mission_code/Academy.dm
+++ b/code/modules/awaymissions/mission_code/Academy.dm
@@ -345,9 +345,15 @@
 			user.physiology.burn_mod *= 0.5
 
 		if(20)
-			//Free wizard!
-			T.visible_message("<span class='userdanger'>Magic flows out of [src] and into [user]!</span>")
-			user.mind.make_Wizard()
+			//Free wizard! //NOT ANY MORE FUCKING CHRIST
+			T.visible_message("<span class='userdanger'>Magic arches out of [src] and into ground under [user]!</span>")
+			new /obj/item/clothing/suit/wizrobe(drop_location())
+			new /obj/item/clothing/head/wizard(drop_location())
+			new /obj/item/clothing/gloves/combat/wizard(drop_location())
+			new /obj/item/clothing/suit/wizrobe/magusblue(drop_location())
+			new /obj/item/clothing/head/wizard/magus(drop_location())
+			new /obj/item/staff(drop_location())
+			new /obj/structure/mirror/magic(drop_location())
 
 /datum/outfit/butler
 	name = "Butler"


Erikafox:
👍 I spawned the die of fate in and it did what I wanted it to do

Sun-Soaked:
:c 
(I rember when we were talking about the D20 being "not good enough")
(All things considered I think it's a fair "best possible award" for an object that can permanently slow you down, reduce your max health, or just instantly permakill you)

Ms-Mee:
> ## About The Pull Request
> 
> replaces the die of fate's d20 effect (spawn you as wizard) with spawn wizard clothes and magic mirror under you.
> ## Why It's Good For The Game
> 
> I'm sick of wizards spawning without admin intervention

Don't see this as too good a reason to change the item in such a manner, if it is that much a problem just remove the sources of it. The die is already a terrible gamble as-is, making the best possible reward a discount lamia makes it pretty worthless. Would also be a decent idea to add a notification to admins as for when someone gets turned into a wizard by the die if there isn't one already.
The item does need a rebalance regardless as it's built around station gameplay, pretty obviously so with the captain id, but this is not a good change as a simple huge nerf to the advertised best reward to the die that you risk getting instantly killed for.
> ## Changelog
> 
> cl tweak: You can't be a wizard with a die of fate anymore /cl

balance: You can't be turned into a wizard by the die of fate, instead getting a magic mirror and wizard clothes.

Erikafox:
>Don't see this as too good a reason to change the item in such a manner

Let me rephrase.
I am tired of the administrative burden of antagonists being created by an item obtainable by players in game. If we're going to remove it, then we'll remove it, and this is a stopgap.

>balance: You can't be turned into a wizard by the die of fate, instead getting a magic mirror and wizard clothes.

👍

Ms-Mee:
And if we aren't? A good stopgap is to remove the few sources of the item until it is either rebalanced or set to be removed, as I stated.

Sun-Soaked:
If I may ask, what's the issue here? Is it the powers themselves, or the antag status? 
If it's the whole "You're a wizard! You're an antag with missions! Kill people and cause chaos!" thing we could pretty easily disable the popup and objectives all together

Erikafox:
I encourage you both to implement your solutions.

ArdentSpark:
as a trial admin, die of fate sucks
there needs to be work done on wizard to make it admin friendly, until then, goodbye
admin intervention is required to get them out of the wizard location, and god forbid they press one of the buttons that either summons ghosts as antagonists, or I don't know
make the entire server psuedo antagonists with objectives and magical items

why would we remove the source of the item that has a chance to be a source of the problem, when we can just remove the problem
If people want to roll dies of fate, they can go nuts

if people want to fix wizard to let it be rolled, then go nuts but this wasted a lot of my time

the problem with wizard are, the buttons that affect the entire server or create additional antagonists, and the fact they need admin intervention to even let them play the game

Apogee-dev:
you're a better person than me. I would have obliterated the DoF entirely.

Ms-Mee:
I am unfortunately incapable of changing maps, meaning if there are still any ruins with the die, which I do recall there being, I can't change those. I could add the admin warning, among changing wizards to get a random assortment of spells, only get to pick spells, etc. and not be sent to the wizard zone, though spells alone still have quite the potential for trouble.

I don't think the die of fate is quite as troublesome as the antimatter engine to warrant instant essentially permanent removal but it just isn't a good thing as is, neither is reworking it a high priority as far as development efforts go, which as I see it means making it unattainable without admin intervention a pretty decent way to deal with it until someone props up to take care of it.
I'll see about changing the nature of fate wizard later today, if nothing else is decided upon by a larger consensus.

ArdentSpark:
what you are saying is exactly why the wizard part is being removed, instead of the ways to obtain it, because there's so many methods and ways

nothing is a permanent removal in an open source codebase

Ms-Mee:
I haven't seen the die other than in one specific ruin and that was nearly a year ago, as far as I'm aware it isn't a particularly accessible item.

ArdentSpark:
> I haven't seen the die other than in one specific ruin and that was nearly a year ago, as far as I'm aware it isn't a particularly accessible item.

there's at least two other sources

Ms-Mee:
As long as this PR isn't the only change that is made permanent and a proper solution to the die of fate comes to be, I don't particularly mind the direction taken to solve the administrative problem, as much as it would be rather disappointing for an actual player. I just doubt a proper rework or removal will be decided upon anytime soon, and this change isn't too great as far as the health of the item itself is concerned.

PR #1767, opened 2023-02-12
removes weird "median" floors

tmtmtl30:
## About The Pull Request

removes some strange WS-era "median" floors that used dirs to function and, by the looks of it, were never really used anywhere. this kind of shit is the job of decals, not floor subtypes

## Why It's Good For The Game

fuck chesterton's fence, why did these exist in the first place? there's no reason for them, and since sticky floor dir is causing bugs, might as well get rid of a piece of code that might use it

## Changelog

:cl:
del: Removed some obsolete floor typepaths
/:cl:


MarkSuckerberg (approved):
bye

PR #1777, opened 2023-02-20
Minor fixes/tweaks to the Aegis-Class

meemofcourse:
## About The Pull Request

One of the SMES was not charging, the Syndicate Jacket in the Captain's closet was invisible, the medihuds in the doctor's closets were for cyborgs and their first aid kits were empty, and the large vials were useless because there wasn't a hypospray that could actually take them. This PR fixes these issues, I think.

It also adds grass without sidings, and changes the tiny fans in the cargo bay with a holofield.

## Why It's Good For The Game

Fix good for videogame. Me like fix. You should like fix too.

## Changelog

:cl:
tweak: Added a deluxe hypospray to the Aegis's Lead Doctor locker, so they can actually use the large vials, and changes its cargo bay's tiny fans with a holofield.
fix: The Syndicate jacket inside the Aegis is no longer invisible, its Medihuds and First-Aid Kits are usable, and the wiring was changed so the first SMES recharges.
/:cl:


meemofcourse:
i would like to note that this is my first pr and i have no idea what i'm doing

Erikafox:
Please replace the tiny fans in the bay with a holoshield generator while you're at it

BarteG44:
While you're at it you could replace the grass behind the glass with a kind that doesn't spread around it's tile

meemofcourse:
i'm not sure if there's grass that has no sidings

BarteG44:
> i'm not sure if there's grass that has no sidings

There definitely is. I don't have access to my pc to tell you exactly which one that is but if i remeber correctly it's further above in the sdmm menu

Erikafox (approved):

PR #1788, opened 2023-02-24
Jupiter-Class Stormrider

Erikafox:
## About The Pull Request
Adds a new pirate ship, the Jupiter-Class Stormrider.
It's designed as an AI centric raiding ship, with the ability to dive into storms to recharge its powerbanks.

https://user-images.githubusercontent.com/94164348/221093940-d3ce5262-8fe0-4877-bafb-e04381c32b44.mp4

## Why It's Good For The Game
Pirate.... good?

## Changelog

:cl:
add: An unknown warship has been spotted lurking in electrical storms. Be cautious, starfarers.
/:cl:


spockye (changes requested):
image
try to only have one large wallmount per wall

image
please put some stuff in the captain's locker. Do also name it!
image

you can't access the cameras, lights, or air tanks. (i also don't think it's a good idea to have that kind of self-destruct on the ship) Perhaps use a bomb instead?

image
this area is a tad bland. maybe give them some beds?

image
please put buttons next to the cargo entrance!

image
Push the arm one out, and add another plating, so you can walk across please 🥺 


cargo bay is lacking an APC. So is bridge and engineering

Please also add air alarms, fire alarms, comms units, and such!

the ai core is a bit too cramped imo. you cant access the self-destruct without moving the ai core, which is a tad hard, seeing the size of the cycling entrance

Fun gimmick. I'd love to see how it would play out ingame!

Erikafox:
>you can't access the cameras, lights, or air tanks. (i also don't think it's a good idea to have that kind of self-destruct on the ship) Perhaps use a bomb instead?

I've swapped it out for a bomb, but the lack of access is intentional, it's supposed to be something only the AI can access, if it's accessed at all.

Rest of review addressed

Erikafox:
okay, balance isn't final, but the datums are in

goober3:
the Bridge Lockdown button isn't linked to the bridge blast doors. the captain could perhaps do with an ID Console for granting access to the Bridge and Armory. that would be neat, I think

Zevotech:
turrets are in the walls and will shoot the adjacent walls when targeting
the batons start without powercells, and require a high capacity at minimum to even use
fuel tank blocks in the welding supplies locker
would be a bit nicer if the long side hallways were 2 wide instead of 1 wide

spockye (changes requested):
image
Please free this innocent locker, Also remove it's access restrictions!

image
use the loaded subtype

image
still not a  huuuuge fan of this area. could definitely use some sprucing, but it's not too bad.

it could definitely use some food aswell. IF this ship is intended for IPC's only, then im sure it's fine if you just ditch the bathroom to expand on the crew area.

spockye (changes requested):
image
Please free this innocent locker, Also remove it's access restrictions!

image
use the loaded subtype

image
still not a  huuuuge fan of this area. could definitely use some sprucing, but it's not too bad.

it could definitely use some food aswell. IF this ship is intended for IPC's only, then im sure it's fine if you just ditch the bathroom to expand on the crew area.

Fikou (dismissed):
you cant copypaste all the code from tesla coils just to remove 1 function of them. edit the original so that you can stop it from doing research, and make this a subtype fo that.

spockye (commented):
image

unconnected pipes

Latentish:
It has my approval personally.

Latentish (approved):
I approve.

spockye (approved):

PR #1792, opened 2023-02-25
Switches away from Prettierx

MarkSuckerberg:
## About The Pull Request
Prettierx hasn't had a release since 2021, and the latest prettier has no issues with how the code is at the moment. I'm just going to make an executive decision and move us to the latest prettier version as I'm truthfully unaware of the circumstances that caused TG to use prettierx instead of prettier main.

## Why It's Good For The Game
Allows devs to actually format the TGUI files in a way that passes ESLint

## Changelog

:cl:
/:cl:


Arturlang:
PrettierX appears to have more options for configuration, not sure if the new versions now have these

PR #1803, opened 2023-03-03
Datumises Planet Types & Adds Ruin Weighting Framework

MarkSuckerberg:
## About The Pull Request
Does what it says on the tin. Moves the planet types out of a massive switch statement, and into a datum for each type. Also puts all ruin types into a singular list of lists with the type as the key instead of adding a new var to the mapping subsystem and hooking it up like that. 

Also adds a system for ruin selection weighting, everything is an even 1 for right now though so it's no real gameplay change yet.

## Why It's Good For The Game
Allows for even easier creation of ruin types and planet types, which I suppose is needed.

## Changelog

:cl:
/:cl:


code/datums/ruins.dm
```diff
@@ -6,9 +6,6 @@
 	How is there a wooden container filled with 18th century coinage in the middle of a lavawracked hellscape? \
 	It is clearly a mystery."
 
-	var/unpickable = FALSE 	 //If TRUE these won't be placed automatically (can still be forced or loaded with another ruin)
-	var/always_place = FALSE //Will skip the whole weighting process and just plop this down, ideally you want the ruins of this kind to have no cost.
-	var/placement_weight = 1 //How often should this ruin appear
```
Erikafox, line 11: if its actually working, I would like to keep this, having some granularity over more loot heavy ruins will eventually be nice

MarkSuckerberg: I was just thinking similar, no, it didn't work in this, but it would be easy to reimplement honestly. I'll add that to the PR

code/modules/overmap/planets/planet_types.dm
```diff
@@ -0,0 +1,113 @@
...
+	desc = "A large asteroid with significant traces of minerals."
+	planet = DYNAMIC_WORLD_ASTEROID
+	icon_state = "asteroid"
+	color = COLOR_GRAY
+	mapgen = /datum/map_generator/single_biome/asteroid
+	// Space, because asteroid maps also include space turfs and the prospect of space turfs
+	// existing without space as their baseturf scares me.
+	default_baseturf = /turf/open/space
+	weather_controller_type = null
+	ruin_type = null // asteroid ruins when
```
Erikafox, line 102: I know what I'm mapping next!

code/datums/ruins.dm
Erikafox: bitchin

MarkSuckerberg: Added!

BarteG44:
Would be awesome if the weight decreased as they get spawned to limit ruin repeating

Latentish:
Ruin weighting!!!!!! Yeah I love ruin weighting!!!!!!!!! The Maptainers are gonna have a field day with this!!!!

tmtmtl30:
it'd be more work but it'd be fairly useful and reasonable if /datum/overmap/dynamic was subtyped to /datum/overmap/dynamic/lava, etc., and then got qdeleted after takeoff and replaced with a new datum instead of being recycled. it'd make a lot more sense and allow for code to spawn temporary encounters of its own, with specific ruins, WITHOUT them getting recycled by the end of it

MarkSuckerberg:
> it'd be more work but it'd be fairly useful and reasonable if /datum/overmap/dynamic was subtyped to /datum/overmap/dynamic/lava, etc., and then got qdeleted after takeoff and replaced with a new datum instead of being recycled. it'd make a lot more sense and allow for code to spawn temporary encounters of its own, with specific ruins, WITHOUT them getting recycled by the end of it

I was thinking of something similar, and it wouldn't be TOO difficult to do. Might give it a look tomorrow or something

code/modules/overmap/objects/dynamic_datum.dm
```diff
@@ -109,163 +109,21 @@
  * Chooses a type of level for the dynamic level to use.
  */
 /datum/overmap/dynamic/proc/choose_level_type(load_now = TRUE) //TODO: This is a awful way of hanlding random planets. If maybe it picked from a list of datums that then would be applied on the dynamic datum, it would be a LOT better.
```
rye-rice, line 111: maybe remove this comment?

PR #1817, opened 2023-03-06
Map Exporter Overhaul

MarkSuckerberg:
## About The Pull Request
Improves the map exporter's saving features, as well as the map uploader. Uploading a custom ship should now *actually work*, so that's also nice.

Also just replaces our debug_variables.dm with [tgstation](https://github.com/tgstation/tgstation)'s, as ours was outdated and infuriating to work with.

...also readds the blue "vacuum" state to lights for when the area is depressurised. Apparently this was a feature, and it had been broken for who knows how long!

...also the map exporter seems to run a little faster after I did a little reordering, but that may just be wishful thinking.

## Why It's Good For The Game
Allows admins to better export player-made creations such as custom ships and colonies, as well as load said things easier.

## Changelog

:cl:
admin: The map exporter now saves even more details than it previously did (decals, APCs, 
admin: Uploading ship maps in-round now actually works.
admin: View variables looks and works a little bit better.
/:cl:


Erikafox:
we could totally merge this since it works

PR #1844, opened 2023-03-25
Blast doors and shutters above windows

TheLeopold:
## About The Pull Request
Changes blast doors and shutters layer to cover windows

## Why It's Good For The Game
Shutters being between window and grille didn't make any sense. Blast doors can be actually used for protection of the ship, rather than a fancy curtain,

## Changelog

:cl:
balance: bast doors and shutters protect windows
/:cl:


goober3:
idk if I like this. I wanna see how it looks with any% walls tm'd but closed window shutters look way too much like shutters placed on walls now.

Erikafox:
This did not look good when we tested it. I recommend porting + adjusting /tg/'s directional shutters as a stopgap to the above suggestion

TheLeopold:
Okay that sounds too hard. Maybe later if I figure out how.

PR #1864, opened 2023-04-02
Tones down giant spiders

thgvr:
## About The Pull Request

See title

## Why It's Good For The Game

Spiders are incredibly strong for no good reason, one bite from a normal hunter spider would do around 15 brute and ~30 toxin. One bite from a viper spider would kill you, as Venom has a very low metabolism rate, adds Histamine with a chance per tick, and they'd bite you several times even though just one is guaranteed to kill you. Tarantulas and Hunter spiders could eat an entire energy gun plus some, which may be more fine for tarantulas but definitely not the base variants.

## Changelog

:cl:
balance: Giant spiders should not be horrendously strong anymore
/:cl:


Erikafox (approved):
I am content with the balance implications of this

PR #1885, opened 2023-04-11
mitigates a class of SGT

tmtmtl30:
## About The Pull Request

turns out that BYOND has a bug where using the return value of a runtiming proc in a try block inside a for(var/A in X to Y) loop will cause that loop to break prematurely. this only happens with loops of that form, and was causing SGTs (roller-table, IPC gib). without fixing those runtimes, this PR hopefully mitigates their impact by stopping them from terminating the loops. this should prevent the runtimes from escalating into full-blown SGTs

note: haven't tested this yet

## Why It's Good For The Game

sgts are BAD. evil!!! kill!!!!!!!!

## Changelog

:cl:
fix: Some SGTs should hopefully be less likely / impactful.
/:cl:

### DIFF
--- a/code/game/objects/items/cigs_lighters.dm
+++ b/code/game/objects/items/cigs_lighters.dm
@@ -806,7 +806,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 		playsound(src, 'sound/magic/fireball.ogg', 10, TRUE)
 
 /obj/item/lighter/enigma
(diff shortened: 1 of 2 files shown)


Erikafox:
praise be to the headcoder

ZephyrTFA:
Sure would be a shame if this was closed

tmtmtl30:
i didnt TEST it... well, hopefully it works :)

PR #1887, opened 2023-04-11
STG band-aid fix for the colossus and talos.

axelzonvolt:
Removes rolling tables and replaces them with normal ones on the talos and colossus integ ships to isolate problematic ship-devouring structures.

## About The Pull Request

There's something inherently screwy with rolling tables and they've started eating ships. I'm not experienced with byond coding enough to determine the cause at this time, but I can at least do this for a temporary solution until a later date.

## Why It's Good For The Game
reduces admin headaches and player screeching. And also...

errors

## Changelog

:cl:

del: removed some rolling tables
add: replaced the empty spots with non-eldritch tables.

/:cl:

### DIFF
--- a/_maps/shuttles/shiptest/inteq_colossus.dmm
+++ b/_maps/shuttles/shiptest/inteq_colossus.dmm
@@ -2491,7 +2491,7 @@
 /obj/item/storage/firstaid/medical{
 	pixel_x = -5
 	},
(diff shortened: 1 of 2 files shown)


Erikafox:
we just merged a pr that should fix the issue this is supposed to remedy (#1885)

PR #1889, opened 2023-04-12
Update vote.dm

FeenieRU:
## About The Pull Request

"Initiate Bluespace Jump" instead of "Initiate Crew Transfer" per round end vote.

## Why It's Good For The Game

Because... WHY CREW TRANSFER?

## Changelog

:cl:
fix: transfer vote now has option "Initiate Bluespace Jump" instead of "Initiate Crew Transfer"
/:cl:

### DIFF
--- a/code/controllers/subsystem/vote.dm
+++ b/code/controllers/subsystem/vote.dm
@@ -84,7 +84,7 @@ SUBSYSTEM_DEF(vote)
 						factor = 1.2
 					else
 						factor = 1.4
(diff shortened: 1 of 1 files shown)


Zevotech:
i think we all got so used to autocorrecting it in our minds that we just kept forgetting to change it lmao

MarkSuckerberg (approved):
this is why stuff should be #defines and not strings, very good catch

PR #1901, opened 2023-04-15
Shuttle creator now properly assigns ship to ship area.

The-Moon-Itself:
## About The Pull Request
The ship area of a newly created custom shuttle didn't have its mobile_port var assigned, which was causing runtimes and SGTs when trying to fly the custom shuttle. This PR properly sets that var by making area/ship/connect_to_shuttle() call link_to_shuttle(), where this was previously handled.

## Why It's Good For The Game
Makes custom shuttles usable again.

## Changelog

:cl:
fix: Fixed a bug causing custom shuttles to runtime and SGT when taking off.
/:cl:


MarkSuckerberg (approved):
Very good catch, hah.

PR #1922, opened 2023-04-23
Squishes the Cricket

Apogee-dev:
## About The Pull Request

Makes the cricket an adminspawn ship, for the time being.

## Why It's Good For The Game

The Cricket is something of a relic, based on one of the very first iterations of a faction that's been reworked at least twice since then and which still does not have any assets that reflect the current lore. On top of that, it's an antiquated design that reflects a fairly early period in Shiptest balance and which is sorely in need of a complete overhaul or outright replacement.

What makes this a particularly pressing issue is the fact that the Cricket is one of vanishingly few ships that has an omni-lathe, and as such it's been getting increasing amounts of play as more players catch onto this. Which is to say, it has Luxembourg Syndrome and is causing players to latch onto an image of Solgov that's completely incorrect on top of it all.

## Changelog

:cl:
del: Disabled the Cricket
/:cl:

### DIFF
--- a/_maps/configs/solgov_cricket.json
+++ b/_maps/configs/solgov_cricket.json
@@ -14,5 +14,5 @@
 		"Cook": 1,
 		"Assistant": 3
 	},
-	"enabled": true
+	"enabled": false
 }


Zevotech:
Just Take The Omnilathe Out

Apogee-dev:
> Just Take The Omnilathe Out

i only wrote like a whole paragraph explaining all of the other problems with the cricket before even mentioning the omnilathe. Ultimately solgov's place on the frontier is still way up in the air right now, and this ship has huge lore compliance issues on top of that.

spockye (approved):

Latentish (approved):

PR #1953, opened 2023-05-12
ORM QOL: input/output directions can now be changed separately

Hibou48888:
## About The Pull Request

Adds the ability to whack an ORM with a multitool while its panel is open to change whether or not alt+clicking it will rotate the input direction, output direction, or both. Default functionality on a fresh or mapped ORM is still both.

## Why It's Good For The Game

Currently ORMs can only output on the tile directly opposite the tile they take in ore from. This worked fine back on a station, but on a ship space is limited, and the crewmember mining might be the crewmember needing those materials immediately afterward.

Now, shiptesters no longer have to fear ORMs spitting their contents into walls or furniture when building one in a corner, dead end, or other spot where having the input/output tiles be directly opposite each other posed logistical problems.

## Changelog

:cl:
code: multitools now change the alt+click behavior of an ORM
tweak: ORMs can now have their input and output tiles onto the same tile, or two tiles at a right angle, as well as the old behavior of having them on opposite sides of the machine
/:cl:

### DIFF
--- a/code/modules/mining/machine_redemption.dm
+++ b/code/modules/mining/machine_redemption.dm
@@ -23,7 +23,7 @@
 	var/datum/techweb/stored_research
 	var/obj/item/disk/design_disk/inserted_disk
 	var/datum/component/remote_materials/materials
-	var/direction_to_edit = 0 //1 is input, 2 is output, any other value is both (default) (should only ever be either 0, 1, or 2, but won't completely break if something weird happens and it becomes >2 or <0)
+	var/direction_to_edit = ORM_BOTH //ORM_INPUT is 1, ORM_OUTPUT is 2, any other value is ORM_BOTH (defaults to 0) (should only ever be either 0, 1, or 2, but won't completely break if something weird happens and it becomes >2 or <0)
 
 /obj/machinery/mineral/ore_redemption/Initialize(mapload)
 	. = ..()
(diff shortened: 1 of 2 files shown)


code/modules/mining/machine_redemption.dm
```diff
@@ -23,6 +23,7 @@
 	var/datum/techweb/stored_research
 	var/obj/item/disk/design_disk/inserted_disk
 	var/datum/component/remote_materials/materials
+	var/direction_to_edit = 0 //1 is input, 2 is output, any other value is both (default) (should only ever be either 0, 1, or 2, but won't completely break if something weird happens and it becomes >2 or <0)
```
MarkSuckerberg: Maybe switch this into a define?
e.g. have this in a defines file:

```dm
#define ORM_BOTH 0
#define ORM_INPUT 1
#define ORM_OUTPUT 2
```

and then have
```dm
var/direction_to_edit = ORM_BOTH
...
if(direction_to_edit == ORM_INPUT)
``` 
etc

other than that, code looks good

Hibou48888: Thanks for the pointer, I had a little voice in the back of my head telling me there was a better way to do this, just didn't know how. Should be implemented properly now, lemme know if I did something weird.

MarkSuckerberg (approved):

PR #1958, opened 2023-05-19
Literally removes RnD from every ship

Apogee-dev:
as_we_speak

## About The Pull Request

Removes RnD from every ship, potentially as we speak.

## Why It's Good For The Game

Ships are too independent and the job of the scientist and medic is far too stale- All they do is stand in one place and click buttons. Removing RnD encourages cooperation and more interaction with the outpost cargo system.

## Changelog

:cl:
balance: Fixes RnD balance
/:cl:

PR #1967, opened 2023-05-22
Doubles down on making RnD unobtanable

mystery3525:
## About The Pull Request

https://github.com/shiptest-ss13/Shiptest/pull/1958 was an actual patch job, no offense. Some ships didn't even get their lathe removed and ruins where completely glossed over. This PR uses one of our built-in scripts (the update paths one) to ensure that **ALL** maps have RnD removed. I also did a once-over and removed RnD board drops from loot and whatnot.

In addition to things that print and the RnD console, this makes Bepis, Destructive Analyzer, Experimentor, RnD Servers, and Exosuit Fabricators unobtainable as I believe these machines are useless without the RnD console.

## Why It's Good For The Game

It's not. I think this is unequivocally a bad idea for this server at its current state, but with https://github.com/shiptest-ss13/Shiptest/pull/1958 merged, we might as well make it so that testers cannot go around these restrictions? I am all for proving that this is a bad idea, and what better way of proving it is than making sure there is no way to go around it.

thanos-infinity

## Changelog

:cl:
del: Makes RnD unobtainable normally (take 2)
/:cl:


Erikafox:
No

PR #1987, opened 2023-05-28
Removes crew manifest from PDA, Modcomp, AI, PAI

thgvr:
## About The Pull Request
I do not think this information should be available to living players.

## Why It's Good For The Game
Having an exact print out of exactly who is on someone's ship is very bad for any conflict scenario and feels a little off. We aren't on a station, every ship doesn't need a print out of other ships. You can call ships on a holopad for a similar effect without being so verbose.

## Changelog

:cl:
del: PDAs, Modcomps, AIs, and pAIs no longer have a crew manifest option.
/:cl:


BarteG44:
For me having the crew manifest was always a great way of determining what ship I'm dealing with. Multiple times I've looked at it in search of a doctor role when im hurt or a miner role when i want to buy some resources not to mention looking at the ship classes. In it's current state it *does* reveal a little too much info but I'd rather have this than have to holopad every single ship asking if maybe this one has what I'm looking for (especially if I'm bleeding out as I'm calling).

TheNeoGamer42:
one singular thing I would like instead of removing them altogether is just limiting the crew manifest to your own ship, so maybe add a ship crew manifest to something on the ship, cryo pod maybe? helm? though maybe just do that in another PR

MarkSuckerberg:
yeah that's something for another PR

MarkSuckerberg (approved):

PR #1989, opened 2023-05-28
Make it so you can't use any item to cancel mechanical surgery

ritorizo:
## About The Pull Request

Right now if you click an ipc surgery with anything that is not the right item it "screw" the shell and force you to restart teh surgery which is a pain. (it applies to EVERY item like even gause).
You can still close the shell with a knife if you don't have a screwdriver.

## Why It's Good For The Game

Said above

## Changelog

:cl:
tweak: Mechanical surgery won't cancel at any wrong move.
/:cl:


Zevotech:
FINALLYYYYYYYYY

CATHARSIS

ritorizo:
wrong pr name lmao

code/modules/surgery/mechanic_steps.dm
```diff
@@ -18,8 +17,6 @@
 /datum/surgery_step/mechanic_open/tool_check(mob/user, obj/item/tool)
 	if(istype(tool))
 		preop_sound = tool.usesound
-	else if(!tool.get_sharpness()) //if its not a tool, then we check if its sharp
```
MarkSuckerberg, line 21: honestly I think if you just made this an `if` instead of `else if` it would make it work as intended (as in, only sharp items work for it)

ritorizo:
Ok here is a better fix that tg use. It allow for usage of anything sharp without making newspaper close the surgery.

code/modules/surgery/mechanic_steps.dm
```diff
@@ -16,11 +16,11 @@
 			"<span class='notice'>[user] begins to unscrew the shell of [target]'s [parse_zone(target_zone)].</span>")
 
 /datum/surgery_step/mechanic_open/tool_check(mob/user, obj/item/tool)
-	if(istype(tool))
-		preop_sound = tool.usesound
-	else if(!tool.get_sharpness()) //if its not a tool, then we check if its sharp
+	if(implement_type == /obj/item && !tool.get_sharpness())
```
MarkSuckerberg, line 19:
```suggestion
	if(istype(tool) && !tool.get_sharpness())
```

code/modules/surgery/mechanic_steps.dm
```diff
@@ -40,11 +40,11 @@
 			"<span class='notice'>[user] begins to screw the shell of [target]'s [parse_zone(target_zone)].</span>")
 
 /datum/surgery_step/mechanic_close/tool_check(mob/user, obj/item/tool)
-	if(istype(tool))
-		preop_sound = tool.usesound
-	else if(!tool.get_sharpness()) //if its not a tool, then we check if its sharp
+	if(implement_type == /obj/item && !tool.get_sharpness())
```
MarkSuckerberg, line 43:
```suggestion
	if(istype(tool) && !tool.get_sharpness())
```

code/modules/surgery/mechanic_steps.dm
```diff
@@ -16,11 +16,11 @@
...
 /datum/surgery_step/mechanic_open/tool_check(mob/user, obj/item/tool)
-	if(istype(tool))
-		preop_sound = tool.usesound
-	else if(!tool.get_sharpness()) //if its not a tool, then we check if its sharp
+	if(implement_type == /obj/item && !tool.get_sharpness())
 		return FALSE
-	. = ..()
+	if(tool.usesound)
+		preop_sound = tool.usesound
+	return TRUE
```
MarkSuckerberg:
```suggestion
	return ..()
```

code/modules/surgery/mechanic_steps.dm
```diff
@@ -40,11 +40,11 @@
...
 /datum/surgery_step/mechanic_close/tool_check(mob/user, obj/item/tool)
-	if(istype(tool))
-		preop_sound = tool.usesound
-	else if(!tool.get_sharpness()) //if its not a tool, then we check if its sharp
+	if(implement_type == /obj/item && !tool.get_sharpness())
 		return FALSE
-	. = ..()
+	if(tool.usesound)
+		preop_sound = tool.usesound
+	return TRUE
```
MarkSuckerberg:
```suggestion
	return ..()
```

MarkSuckerberg (changes requested):

code/modules/surgery/mechanic_steps.dm
ritorizo: implement_type is the type path that was matched in the implement list 
(TOOL_SCREWDRIVER = 100,
 TOOL_SCALPEL = 75,
  /obj/item/kitchen/knife = 50, /obj/item = 10)

ritorizo: Argh it sent it before I finished typing.
So istype() will match anything as long as it's an item. implement_type == /obj/item will only match if we're in the final case were we are looking for any sharp item (and not a knife for instance)

MarkSuckerberg: yes, which is what we're trying to do here— we want any /obj/item that is also sharp

ritorizo: Ok so here `tool`, is whatever object we are using to do the surgery step, so lets say we use a screwdriver it would be an instance of `/obj/item/screwdriver` which is an instance of `/obj/item` so `istype(tool)` would be True, but the screwdriver is not a sharp item so `!tool.get_sharpness()` 

Surgery steps have a dict called `implements` which link either a tool behaviour like TOOL_SCREWDRIVER or TOOL_SCALPEL or a typepath (like /obj/item/kitchen/knife) to a success modifier for the surgery. 
This dictionary is ordered from the best tool to the worst.

in our case the dictionary is:
``` 
implements = list(
	TOOL_SCREWDRIVER		= 100,
	TOOL_SCALPEL 			= 75, // med borgs could try to unscrew shell with scalpel
	/obj/item/kitchen/knife	= 50,
	/obj/item				= 10) // 10% success with any sharp item.
```

 - If we use a screwdriver TOOL_SCREWDRIVER will be detected at the top of the list so it will be our `implement_type` which is not equal to `/obj/item` and therefore no sharpness check will be made and the success modifier will be 100.
 If we use a knife the `implement_type` will be `/obj/item/kitchen/knife` (typepath) which is not strictly equal to `/obj/item` so no sharpness check will be made. We don't care that no check were made since the knife is sharp but if we ever add other item suposed to open shell and that are not sharp it will matter.
 
 Also I tried this change on local it started the surgery but skipped directly to canceling instead of opening.

ritorizo: `istype(tool)` is checking if tool is an item, which is true for everything and therefore always require the tool to be sharp aka, screwdrivers won't work.

`implement_type == /obj/item` is only true when no explicitly declared tool was found (aka false when it's a screwdriver) and therefore won't check for sharpness if it's a screwdriver.

MarkSuckerberg: makes sense

MarkSuckerberg (approved):

PR #2013, opened 2023-06-03
'fixed' the ticket panel for you.

ZephyrTFA:


code/modules/admin/verbs/adminhelp.dm
```diff
@@ -220,6 +231,7 @@ GLOBAL_DATUM_INIT(ahelp_tickets, /datum/admin_help_tickets, new)
 	RemoveActive()
 	GLOB.ahelp_tickets.closed_tickets -= src
 	GLOB.ahelp_tickets.resolved_tickets -= src
+	GLOB.ahelp_tickets.active_tickets -= src
```
Fikou: this is a part of RemoveActive

ZephyrTFA:
```suggestion
```

PR #2034, opened 2023-06-09
Total Remap of the Ranger

Latentish:
## About The Pull Request
The old ranger is just about a year old. This totally remaps this ship and rebalances it for our current standards.
rangerpng

This is good for the game because it replaces a year old ship with something that's more up to par with our standards.
## Why It's Good For The Game
Ship upkeep and maintaining our maps is a good thing. Plus the old ranger had a bunch of old snowflaked code that was a bandaid fix for the ship access system not existing yet. (It unsnowflakes it).

## Changelog

:cl:
tweak: Remaps the Ranger-class
/:cl:


Latentish:
ToDO:

- [x] Fire Alarms
- [x] Cameras
- [x] Holopads
- [x] Access
- [x] Ripping out old code
- [x] Fixing docking ports
- [x] Decals

BarteG44:
the modern railing handrails can be set to have a custom color 
image

Latentish:
rangernewest

The group pissing picture took 8 minutes to take you get these in a comment this time.

Erikafox (dismissed):
Something in here has changed the default captain outfit to be the LP lieutenant outfit.

Latentish:
> Something in here has changed the default captain outfit to be the LP lieutenant outfit.

I don't see anything in the code that would imply this. Could you give me more specifics? Do you mean outfit datum? Outfit as in clothes? Did you see this on tm? Was it a full outfit swap?

Latentish:
> Something in here has changed the default captain outfit to be the LP lieutenant outfit.

We figured this out, and now know that not properly declaring jobs in jsons can cause this. Funny haha moment. I'll make a pr for this l8r.

OvercastAngel:
This remap leaves out the intercoms from the original. While not conventionally useful for the crew, it's very necessary for the AI that the ship comes with!

code/datums/lp_mission_holotapes.dm
```diff
@@ -286,33 +214,6 @@
...
-	desc = "A holodisk containing an aid mission for the LP."
-	preset_image_type = /datum/preset_holoimage/commissioner
-	preset_record_text = {"
-	NAME Commissioner Gorre Donn
-	SAY This mission, should you choose to accept it, revolves around an aid objective. The LP will need to provide aid to local vessels not hostile or syndicate.
-	DELAY 25
-	SAY A well equipped Ripley will help any vessel in hard times. Provide 2 allied or otherwise friendly, non-syndicate vessels with an equipped Ripley exosuit.
-	DELAY 25
-	"}
-
```
Erikafox, line 300: I actually like some of these aid missions. They seem conceptually neat.

code/modules/jobs/job_types/lp_jobs.dm
```diff
@@ -1,20 +1,7 @@
...
-/datum/job/lp/lieutenant
-	access = list()
-	minimal_access = list()
-	outfit = /datum/outfit/job/lp/lieutenant
-
-/datum/job/lp/lieutenant/get_access()
-	return get_lplieu_access()
-
-/datum/outfit/job/lp/lieutenant
+/datum/outfit/job/captain/nt/lp_lieutenant
```
Erikafox: outfit datums should go in relevant job files. It makes it easier to tell what they inherit from.

_maps/configs/nanotrasen_powerrangers.json
Erikafox, line 1: I'd like it if this and the ship file was changed to "nanotrasen_ranger" to fit in with the schema of faction-class

Erikafox (dismissed):
image
Stray pump. This is not enough atmospherics for a ship of this size, especially on free access.
image
This is a very visually noisy room for not much gain. Consider going lighter on the decals while using alternative tiles (try mono tiles under machinery - tables, with a blue trimline pattern.) I would also put a smartfridge for blood instead of a crate, and an anesthetic closet for their anesthetic. The medkits could perhaps go in the cargo bay to encourage set up.
image
As an aside, the perspective in this room does not work. The railings are too tall and do not line up with the staircase, while the desk lamp is at the high end of the table, instead of the low end.
StrongDMM_RXzJXjBCVN
Edge right against the stairs damages it as well.
StrongDMM_Z7eeRr5Vk3
Consider using window firelocks in place of doubled locks
StrongDMM_EaXfh1k0VV
This room feels very dry. There's no indication of *where* you should store the plasma, the cans are just kinda there. No dirt, no decals, just floor and cans.

Sidenote: You can attach accesses to buttons. I encourage experimenting with it a little
StrongDMM_lA22vz74fw
Random middle of the hallway autolathe and power control console
image
This is another monotile room with a weak decal pattern. Try again, breaking it up with other patterns of tiles and alternative decal methods.
StrongDMM_aB4oezI5CY
I like the flooring pattern here. I do not like the equipment.
Power generation, an AI Core, Telecomms, and Robotics should not all be clumped together in one room. 
You can preset telecomms equipment, look at how the tegus, chronicle, talos, and colossus handle it. If you set the "calibrating" var on the messanging server to 0, it will start functioning.
I personally would prefer that the ship only had a relay, set to link to existing NT commnets, and that the rest of the space was used to more clearly define power generation and the robotics.
StrongDMM_2rtLgW36nV
Don't place light switches like this. They look like holofields and that's how we set up holofield switches.
StrongDMM_eHofEL4XG9
This room has the same problem as the other engine room. 
StrongDMM_UT21ekHgAH
This room suffers the same perspective issues that the medical area does, but the decalling is better. I would consider using red plasteel instead of plastitanium, or ditching the edge around the room.
I would also remove the .38 hunting from the room, and give them something such as an E-TAR SMG. If you wanted to be bold. You could refactor it to allow for quick battery cell swapping and give it a suitably NT name.

StrongDMM_AUHg9rjG6z
You can set up brig doors to work with a timer. Look at how the Vela does it in it's brig.
StrongDMM_csy0muVajk
Any airlock this size needs at least two highcap vents.
StrongDMM_6Zwf3HbOTP
This is fine for a quick resupply, but think about how any bulky amount of supplies would be moved.

StrongDMM_bXYAbmIXAN
Don't maintain the decal pattern you use for walkways to under equipment.
StrongDMM_frOSZ97qQy
I honestly don't think this room needs the railings. It looks p good. Also stray wallmount in the bottom of the screenshot.
StrongDMM_M5F1lcmyYe
This will lead to the left (rear) being crushed by some subshuttles. Their dock will land exactly on this dock, and that can cause them to partially land on the hull.
StrongDMM_FrcyFFqMDR
I'd swap this for 1 pump - 1 big scrub, an oxycan, an aircan, and a spare pacman.
StrongDMM_2ukPEzFWEl
Try doing a trek style bridge (Beluga has one)
My advice is to lengthen the ship a little bit and use the extra space to more clearly define your areas.

Latentish:
hokay it's ready for the testering crew to look at its testerability

Sly9sky:
Mainly with what I've seen so far, the Medical Specialist should get a stretcher if they don't have one already, and the bridge doesn't start out with a helm console, to which I think the foremost console should be replaced with one.

Latentish:
this int test is on some shit

Latentish:
image
This is neither from my map, nor from this latest commit. There is no "library" area on the ranger.

Erikafox:
image
the equipment room needs decals, the other room still has the Bad Decal Pattern

spockye (approved):
Que the crazy frog music

PR #2036, opened 2023-06-09
Various limitations on ship spawning

tmtmtl30:
## About The Pull Request

Adds a variety of restrictions to ship spawning, and joining ships as officer.
- Ship limits have been fixed: by default, any ship template can only be spawned by players through the join menu twice per round. This can be configured per-ship in the ship template JSON.
- Admins with R_ADMIN can now disable the ability of players to spawn new ships via the join menu or shuttle creator.
- If the number of ships _with_ a non-null source_template (including ones built via the shuttle creator) that is _not_ part of the "subshuttles" ship category meets or exceeds the MAX_SHUTTLE_COUNT config number (default 20), players may not create new ships via the join menu or shuttle creator.
- Each ship template has a pair of variables defined in the template's JSON: spawn_time_coeff and officer_time_coeff, both defaulting to 1. If the config flag USE_EXP_TRACKING (off by default) is true and there is an active database:
  - A player cannot spawn a ship via the join menu if their tracked Living playtime in minutes is less than the template's spawn_time_coeff times the SHIP_SPAWN_BASE_EXP_MIN config number (default 600, corresponding to 10 hours of playtime). 
  - A player cannot join on a ship as an officer job via the join menu if their tracked Living playtime in minutes is less than the ship's template's officer_time_coeff times the OFFICER_JOIN_BASE_EXP_MIN config number (default 300, corresponding to 5 hours of playtime). This restriction does not apply when spawning a new ship and thus joining it, regardless of whether the first job is an officer job.
  - If the config flag USE_EXP_RESTRICTIONS_EXP_BYPASS (off by default) is true, and a player has the R_ADMIN admin permission, or if a player has an exemption flag set (currently, I don't think there is an accessible way of setting this flag?), that player will be exempt from playtime-based ship and job restrictions.

It's likely that the cap calculation should be made more robust, so that "dead" ships do not count towards the cap. I'd like suggestions from staff on how best to do that.

## Why It's Good For The Game

These features have been requested by administration, to reduce the number of soloships, ease new players into the game, and place an upper bound on planetgeneration lag.

## Changelog

:cl:
tweak: There is now a cap on the number of ships that may be spawned by players in a round.
tweak: Spawning a ship, or joining as an officer on a ship, may now carry a playtime requirement.
fix: Ship limits have been fixed: Many ships can only be spawned a finite number of times per round.
admin: Admins may now disable join menu ship spawning.
/:cl:


MarkSuckerberg (approved):
Seems fine, just need to run prettier on the one file

PR #2039, opened 2023-06-10
FOXALIVE - Disables Nanites

Apogee-dev:
## About The Pull Request

Removes nanite research from the tech web pending a more extensive rework. Does not touch any pre-placed nanite equipment or disks in ruins and on ships.

## Why It's Good For The Game

nanites are kind of a nightmare for balance, frankly. While they offer some great potential, a lot of the more Gamerly programs need to be reworked or axed and that will take time to do.

## Changelog

:cl:
del: Disabled research nodes for nanites
/:cl:

### DIFF
--- a/code/modules/research/techweb/all_nodes.dm
+++ b/code/modules/research/techweb/all_nodes.dm
@@ -1003,6 +1003,9 @@
 	export_price = 5000
 
 /////////////////////////Nanites/////////////////////////
+
+//Disabled pending nanite rework --Apogee-dev
+/*
 /datum/techweb_node/nanite_base
 	id = "nanite_base"
 	display_name = "Basic Nanite Programming"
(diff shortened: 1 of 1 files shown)


TheNeoGamer42:
1, WOO YEAH NANITES ARE GONE
2, I think theres more than just the three nanite techs? but it probably doesn't matter too much as I think you need the three removed to even get them

Apogee-dev:
> 1, WOO YEAH NANITES ARE GONE 2, I think theres more than just the three nanite techs? but it probably doesn't matter too much as I think you need the three removed to even get them

I commented out the entire nanite block in the techweb file. If there's more techs somebody hid them somewhere and that's terrible practice.

TheNeoGamer42:
ok no actually github was fooling me 10 different nanite techs seems about right and when I was looking at the changed files it only showed three despite me trying to click to reveal more of the changes??? either way death of nanites

Arturlang:
Why is your PR title completely unrelated? It should be obvious so people can see how important this change is and about what it is, hiding controversial changes is no good.

Apogee-dev:
> Why is your PR title completely unrelated? It should be obvious so people can see how important this change is and about what it is, hiding controversial changes is no good.

It's not actually completely unrelated, but you're right, it isn't completely transparent either.

PR #2056, opened 2023-06-12
Makes Cryosickness a range of effects

Erikafox:
## About The Pull Request
Cryosickness now applies a range of possible effects when you wake up, leading to a range of possible wake ups.

## Why It's Good For The Game
I think it's more immersive to have a range of bad things happening than Bad Thing Always

## Changelog

:cl:
balance: Cryopods now apply a range of effects to you when you wake up.
/:cl:

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -32,7 +32,7 @@ GLOBAL_LIST_EMPTY(cryopod_computers)
 MAPPING_DIRECTIONAL_HELPERS(/obj/machinery/computer/cryopod, 17)
 
 /obj/machinery/computer/cryopod/retro
-	desc = "An interface between crew and the cryogenic storage oversight systems. This one appears to  be strugggling to catch up with the more modren cryogenic storage system version."
+	desc = "An interface between crew and the cryogenic storage oversight systems. Currently strugggling to catch up with the modern cryogenic storage system."
 	icon_state = "wallconsole_old"
 	icon_screen = "wallconsole_old_cryo"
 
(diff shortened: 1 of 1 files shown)


Arturlang:
Using this method has a good chance of causing stacking effects, consider using a switch, maybe?

Erikafox:
The effects are supposed to stack

Arturlang:
> The effects are supposed to stack

The way the adjust_disgust seems to be randomized again seems to suggest against it, unless you really want to give them a whole *lot* of it

Erikafox:
Yes that is the intent

Arturlang:
> Yes that is the intent

Then why not increase the prob range for the arguments instead of running the process multiple times

Zevotech:
to be entirely honest almost everyone ive talked to either absolutely hates the disgust/vomiting or is very neutral to it. Being hungry seems like enough, leave the rest up to the players.

Erikafox:
It's not changing.

goober3:
why is this discourse still occurring

tmtmtl30:
for the record when i did the math there was like a 10% chance of vomiting or so on this implementation, unless i'm missing something. i could do it again

code/game/machinery/cryopod.dm
```diff
@@ -429,17 +429,45 @@ MAPPING_DIRECTIONAL_HELPERS(/obj/machinery/computer/cryopod/retro, 17)
...
 	sleepyhead.SetSleeping(80)
-	sleepyhead.set_disgust(75) //nasty
-	sleepyhead.set_nutrition(70)
-	to_chat(sleepyhead, "<span class='userdanger'>The low-quality cryo pod ejects you unceremoniously, you feel REALLY sick...</span>")
+	if(prob(90))
+		sleepyhead.apply_effect(rand(5,15), EFFECT_DROWSY)
+	if(prob(75))
+		sleepyhead.blur_eyes(rand(6, 10))
+	if(prob(66))
+		sleepyhead.adjust_disgust(rand(35, 45)) //rand
```
MarkSuckerberg, line 466: thanks for the clarification

MarkSuckerberg (approved):
This should satisfy most people

PR #2064, opened 2023-06-14
Replaces Bar Tables in maps

### LINKED ISSUE #1981: Wrecked Interceptor jungle ruin uses fucking BAR SUBTYPE TABLES which FLING YOU
Issue reported from  Round ID: 1604 (Shiptest)

WHAT THE FUCK

Orchidthederg:
## About The Pull Request

Replaces Bar Tables in all maps with standard Wooden Tables, as requested by #1981 (permission and method suggested by Erika). Bar Tables are still intact in code, and the method used was via the UpdatePaths tool.

## Why It's Good For The Game

Prevents people from getting randomly flung around in ruin maps. Notably due to the fact that the Bar Table isn't widely being used or acknowledged anymore.
fixed #1981 

## Changelog

:cl:
tweak: Replaced Bar Tables in all maps where present
/:cl:

### DIFF
--- a/_maps/RandomRuins/BeachRuins/beach_fishing_hut.dmm
+++ b/_maps/RandomRuins/BeachRuins/beach_fishing_hut.dmm
@@ -11,7 +11,7 @@
 /turf/open/floor/carpet,
 /area/ruin/beach)
 "av" = (
-/obj/structure/table/wood/bar,
+/obj/structure/table/wood,
 /obj/item/reagent_containers/food/snacks/cubancarp{
 	pixel_y = 6;
 	pixel_x = 11
@@ -970,7 +970,7 @@
 /turf/open/water/beach/deep,
 /area/ruin/beach)
 "MT" = (
-/obj/structure/table/wood/bar,
+/obj/structure/table/wood,
 /obj/item/reagent_containers/food/snacks/fishfry{
 	pixel_y = 8;
 	pixel_x = 2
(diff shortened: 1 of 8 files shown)


Erikafox (approved):

Latentish (dismissed):
I happen to like bartables. If they aren't just a subtype of normal tables with specific smoothing then they should be made that way, instead of removing them from all maps.

Orchidthederg:
Well, I believe the main issue is the fact that it randomly flings and kills people in ruins, which is why it's been addressed in the issue tracker. Maybe we should take it out of the ruins but leave it in other maps? It's still present in the code, so it's still placeable by admins.

Latentish:
> Well, I believe the main issue is the fact that it randomly flings and kills people in ruins, which is why it's been addressed in the issue tracker. Maybe we should take it out of the ruins but leave it in other maps? It's still present in the code, so it's still placeable by admins.

This is hilarious but I was under the impression they looked different. They don't and you can ignore my prior comment.

Sun-Soaked:
> Well, I believe the main issue is the fact that it randomly flings and kills people in ruins, which is why it's been addressed in the issue tracker. Maybe we should take it out of the ruins but leave it in other maps? It's still present in the code, so it's still placeable by admins.

with two bar tables at an angle some mobs can build up spin and reach incredibly high speeds

PR #2065, opened 2023-06-14
Makes some cleanbot improvements.

NithaIsTired:
## About The Pull Request
Makes cleanbots target adjacent tiles first.
Adds a 1 tile range so they can clean under tables.
Adds the cleaning animation to them.
Adds more things that they can clean to the list
Changes the window name to a more general use outside of stations.

Mostly a port from https://github.com/tgstation/tgstation/pull/71159

dreamseeker_ZFzcJegZQn

## Why It's Good For The Game

Cleanbots have been very useless, this gives them just some tiny improvements so they can be more effective and possibly used more often.

## Changelog

:cl: Nitha
tweak: Various cleanbot improvements.
/:cl:


code/modules/mob/living/simple_animal/bot/cleanbot.dm
```diff
@@ -289,23 +287,37 @@
...
+		/obj/effect/decal/cleanable/crayon,
+		)
 
 	if(trash)
-		target_types += /obj/item/trash
-		target_types += /obj/item/reagent_containers/food/snacks/deadmouse
+		target_types = list(
+		/obj/item/trash,
+		/obj/item/reagent_containers/food/snacks/deadmouse,
+		)
```
MarkSuckerberg: I thiiiink this might make these options mutually exclusive instead of multiple things you can have? Simple fix, you can keep them as lists but have them once again be `+=` instead of just `=`

NithaIsTired: ah that's true, hold on

MarkSuckerberg (approved):
Cleanbot.

PR #2080, opened 2023-06-18
Organizes lists in surgery files, nerfs ghetto surgery a bit

thgvr:
## About The Pull Request
I'm not a big fan of people just whipping out a toolbox or whatever, and being able to fully revive people. What's the point of medical tools and/or ships if it's effectively fully available, to everyone?
* Removes pen from tend wounds
* Rebalances ghetto surgery success/fail
* Brain / Heart will now decay in around 45 minutes.
* Every other organ will decay in around 30 minutes.

The gist of it is this, in order of effectiveness:
* Scalpel analogue is Kitchen Knife / Sharp Things
* Retractor is Screwdriver
* Hemostat is Wirecutters
* Saw is Fireaxe, Hatchet, Sharpthings
* Cautery is Welder / Hot Things

Other than tool changes, expect the following:
* 40% surgery success chance with the above tools is baseline for most operations.
* Failing a step will do brute damage, break a bone, or other side effects I can come up with
* The chances can still be increased with ethanol, sterilizine, gloves, beds, etc. (I think this is how it works)

## Why It's Good For The Game
A greater importance should be placed on having medical equipment versus just a toolbelt. I would prefer to see all of it go but that would require a lot more work on the medical system.

## Changelog

:cl:
del: You can no longer tend wounds with a pen
balance: Ghetto surgery is less reliable across the board.
balance: Organs will now decay in ~30 minutes, heart and brain in ~45 minutes.
balance: Inducers now work for revival surgery.
/:cl:


code/modules/mob/living/brain/brain_item.dm
```diff
@@ -11,7 +11,7 @@
 	attack_verb = list("attacked", "slapped", "whacked")
 
 	///The brain's organ variables are significantly more different than the other organs, with half the decay rate for balance reasons, and twice the maxHealth
-	decay_factor = STANDARD_ORGAN_DECAY	/ 2		//30 minutes of decaying to result in a fully damaged brain, since a fast decay rate would be unfun gameplay-wise
+	decay_factor = STANDARD_ORGAN_DECAY	/ 3	//~45 minutes to full decay
```
Erikafox: why not make a new define for this and hearts?

code/modules/surgery/healing.dm
```diff
@@ -25,7 +25,9 @@
 
 /datum/surgery_step/heal
 	name = "repair body"
-	implements = list(TOOL_HEMOSTAT = 100, TOOL_SCREWDRIVER = 65, /obj/item/pen = 55)
+	implements = list(
+		TOOL_HEMOSTAT = 100,
+		TOOL_SCREWDRIVER = 20)//something else could be added here - but I would prefer not. Hemostat is not that hard to come by and SHOULD be standard for ship equipment.
```
Erikafox: hemostats are available in medical aid kits. I think it'd be fine to lower screwdriver's chance down even more given that those are Not Uncommon.

perhaps divide it so powered screwdrivers are better at it.

code/modules/surgery/organic_steps.dm
```diff
@@ -142,7 +162,12 @@
 //drill bone
 /datum/surgery_step/drill
 	name = "drill bone"
-	implements = list(TOOL_DRILL = 100, /obj/item/screwdriver/power = 80, /obj/item/pickaxe/drill = 60, TOOL_SCREWDRIVER = 25, /obj/item/kitchen/spoon = 20)
+	implements = list(
+		TOOL_DRILL = 100,
+		/obj/item/screwdriver/power = 30,
+		/obj/item/pickaxe/drill = 25,
+		TOOL_SCREWDRIVER = 20,
+		/obj/item/kitchen/spoon = 10) //spoon? really? this should be awful.
```
Erikafox: make it awful then

code/modules/surgery/organs/appendix.dm
```diff
@@ -5,7 +5,7 @@
 	slot = ORGAN_SLOT_APPENDIX
 
 	healing_factor = STANDARD_ORGAN_HEALING
-	decay_factor = STANDARD_ORGAN_DECAY
+	decay_factor = STANDARD_ORGAN_DECAY / 2 //~30 minutes
```
Erikafox: change the defines

code/modules/surgery/prosthetic_replacement.dm
```diff
@@ -17,7 +17,11 @@
 
 /datum/surgery_step/add_prosthetic
 	name = "add prosthetic"
-	implements = list(/obj/item/bodypart = 100, /obj/item/organ_storage = 100, /obj/item/chainsaw = 100, /obj/item/melee/synthetic_arm_blade = 100)
+	implements = list(
+		/obj/item/bodypart = 100,
+		/obj/item/organ_storage = 100,
+		/obj/item/chainsaw = 100,
```
Erikafox, line 21: I think there's a conversation to be had about just removing chainsaw arms here

code/modules/surgery/revival.dm
```diff
@@ -27,7 +27,10 @@
 
 /datum/surgery_step/revive
 	name = "shock brain"
-	implements = list(/obj/item/shockpaddles = 100, /obj/item/melee/baton = 75, /obj/item/gun/energy = 60)
+	implements = list(
+		/obj/item/shockpaddles = 100, //this is useful for reviving simepeople.
+		/obj/item/melee/baton = 20, //there's no reason to use a defib with this chance being high - so keep it low.
```
Erikafox: blacklist stunprods

thgvr: I don't think I'm able to blacklist things in these lists, could just make only a specific type I suppose

code/modules/surgery/prosthetic_replacement.dm
thgvr: its kinda cool tho

code/modules/surgery/organic_steps.dm
thgvr: 1/10 is pretty awful. I might just take it out.

code/modules/mob/living/brain/brain_item.dm
thgvr: `STANDARD_ORGAN_DECAY` is `0.00222`. I'm scared of tiny numbers.

code/modules/surgery/revival.dm
thgvr: Actually what if I change it to like. an inducer.

Zevotech: stunprods still need a high powered cell so its somewhat earned

code/modules/mob/living/brain/brain_item.dm
Erikafox: I mean, it logically follows that half of it would be 0.00111, and a third would be 0.00074 then

code/modules/surgery/revival.dm
Erikafox: you can get these "high powered" cells from literally anything, including smeses which is a machine in almost every ship and ruin

I think inducers is better

thgvr: oh right i checked prechargers and not smes. yeah every ship has this

code/modules/mob/living/brain/brain_item.dm
thgvr: We'll try it I suppose

code/modules/surgery/revival.dm
BarteG44: I cannot imagine the disappointment of someone who never used ghetto surgery when they use it as a last ditch effort to save their only other crewmember and they find out that they literally cannot use the stun prod which obviously shocks people to shock the brain. Using revival surgery should give you trauma or something so it comes at a cost instead of making it impossible to do in ghetto circumstances because it uses tools that can pretty much only be obtained using rnd at which point it stops being ghetto. I'm a medship main and i obviously would want PR's that make people more dependent on my ship but you simply cannot assume that there will always be a medship around to do the surgery properly

thgvr: You can always buy a surgery kit at the outpost for not that much really

code/modules/surgery/organic_steps.dm
thgvr: i made it worse

Zevotech:
Cutting the success rates largely across the board doesnt seem like a very interesting or fun way to balance ghetto surgery. An argument could be made for how it is already slow and can fail, but imo the best way to handle it would be wrapping an infection mechanic into wounds or hediffmed for using common mechanical tools on an open wound. I'm not saying to not lower the success rate, but the amount you are lowering it by seems excessive to the point of making it useless. Not to mention stunprod revival surgery is the most accessable one handed revival method when lacking a defib.

zestybastard:
I agree, I would be more interested if success rates were maintained but steps had consequences that required additional resources to manage (e.g. ghetto bonesawing actually creating a fracture in the part). While I think removing prods and batons is a good step, inducers are extremely niche and I would support high-power grille impacts acting as a defib a la Adherent pole.

zestybastard:
Also, if our main issue with ghetto surgery is players using it instead of getting medical help, wouldn't we want to make the onus on prep rather than RNG? That way we still allow players to be alley doctors and such while also making it harder for crews with no medical to start doing ghetto surgery as a spur of the moment decision vs calling a medship.

thgvr:
yeah. i'll think about this some more. maybe I'll make it areal project instead of a bandaid

BarteG44:
>yeah. i'll think about this some more. maybe I'll make it areal project instead of a bandaid

holy hell we got the good ending

thgvr:
You fool. That was the plan in the first place.

thgvr:
Might just mess with revival surgery, and change the organ decay and call this a day

code/modules/surgery/advanced/brainwashing.dm
```diff
@@ -27,7 +27,12 @@
 
 /datum/surgery_step/brainwash
 	name = "brainwash"
-	implements = list(TOOL_HEMOSTAT = 85, TOOL_WIRECUTTER = 50, /obj/item/stack/packageWrap = 35, /obj/item/stack/cable_coil = 15)
+	implements = list(
+		TOOL_HEMOSTAT = 85,
+		/obj/item/soap = 60, //haha. brainwashing. get it?
```
MarkSuckerberg, line 32: im gonna wash your brain. how do you like that huh huh

MarkSuckerberg (approved):
gonna be real at this point it's almost a slight buff if you don't count the stuff that obviously would hardly work for ghetto surgery, so I'm just going to guess and say that this is not controversial anymore

PR #2088, opened 2023-06-21
[DRAFT] Reworks some mining stuff 

Zevotech:
## About The Pull Request
Im going to change how some mining stuff works. Right now the only thing on this PR is an increase in the amount of ore mining bags can hold.
Things I hope to do with this PR:
- [ ] Rework the KPA into a sort of crank-charged energy gun with an internal magazine (instead of a generic infinite mining gun) and some neat new modkits
- [ ] Change how the Crusher functions slightly and make it able to be used as a pickaxe
- [ ] Fix mining vendors falsely displaying most of their items as FREE despite retaining their normal price (If anyone knows how to do this PLEASE tell me so I can fix it)
- [ ] Maybe mess with the loot pools for tendrils to be more useful and less powergame-y
- [ ] Probably nuke pulsating tumors from orbit so that elites can have their own maps and arenas instead of being miner fodder

## Why It's Good For The Game
Slaying the ghosts of lavaland balance's past would probably make the game more enjoyable. Mining equipment is simply a pain in the ass to balance effectively.

## Changelog

:cl:
add: I'll do this later.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/bags.dm
+++ b/code/game/objects/items/storage/bags.dm
@@ -116,7 +116,7 @@
 	STR.allow_quick_empty = TRUE
 	STR.set_holdable(list(/obj/item/stack/ore))
 	STR.max_w_class = WEIGHT_CLASS_HUGE
-	STR.max_combined_stack_amount = 50
+	STR.max_combined_stack_amount = 100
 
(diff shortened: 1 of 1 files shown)


Erikafox:
re-open when you. write more than 1 line over like 2 months

PR #2093, opened 2023-06-22
Fixes some prettier issues people had

Arturlang:
## About The Pull Request
Looks like prettier was autoupdating even before my PR, and it introduced changes that caused it to break if it did, not sure how to fix those, but this does fix it, and also replaces a deprecated config in prettier.

## Why It's Good For The Game
People being able to use format on save is pretty good, y'think?

## Changelog

:cl:
fix: Prettier should now work on VSCode again.
/:cl:


MarkSuckerberg (approved):
yeah, it works now

PR #2097, opened 2023-06-23
Stabilizes space_level handling, preventing overlapping planet generation/allocation

tmtmtl30:
## About The Pull Request

During a recent uptime, the bottom-left corner of a z-level became "cursed". New planets would all generate there, regardless of whether a previous planet already had. When any of the planets unloaded themselves, they would unload their reserve with them, deleting the planet unexpectedly. The other planets which had also generated in the same spot were left behind, appearing to be empty space: in truth, they were the gap left behind after a virtual level is cleared.

As I wasn't able to diagnose the exact cause of the bug during the round it occurred, this PR is admittedly a little bit of a shot in the dark. My initial guess was that there was a null living in the list of virtual_levels on that space_level. Although there weren't runtimes to back this up, I've preemptively fixed that potential bug anyway. Another possibility, which I have also addressed, was downstream of a race condition resulting from the presence of a CHECK_TICK in add_new_zlevel: I have eliminated the race condition by removing the sleep.

## Why It's Good For The Game

Planets overwriting other planets, and deleting people's ships, is bad. That shouldn't happen.

## Changelog

:cl:
fix: Different planets should hopefully no longer repeatedly generate in the exact same part of the map, overwriting each other and causing double-dock SGTs.
/:cl:

### DIFF
--- a/code/controllers/subsystem/mapping.dm
+++ b/code/controllers/subsystem/mapping.dm
@@ -251,7 +251,9 @@ SUBSYSTEM_DEF(mapping)
 /// Creates basic physical levels so we dont have to do that during runtime every time, nothing bad will happen if this wont run, as allocation will handle adding new levels
 /datum/controller/subsystem/mapping/proc/init_reserved_levels()
 	add_new_zlevel("Free Allocation Level", allocation_type = ALLOCATION_FREE)
+	CHECK_TICK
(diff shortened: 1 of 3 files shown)


MarkSuckerberg (approved):
ugh, all stuff that looks obvious as heck now that you've found them

PR #2098, opened 2023-06-23
Fixes the Independent Litieguai organ freezer

FlufflesTheDog:
## About The Pull Request
TLDR - The Litieguai's provided organs now decay as you would expect, pausing when in the shut freezer crate.

As to the nature of the bug, just in case there's a more generalized solution anyone could suggest:
Freezer crates work by toggling the decay state of organs, whenever the crate is opened or closed, as well as when the crate is initialized- in order to freeze any organs that are inside when created.

However, currently the organs are just dropped onto the same tile as the crate, and picked up on a 0 second timer (so after the freeze toggle), which leads to the organs having an inverted decay state- only decaying when the box is closed, and stopping their decay when the box is opened and even when picked up or moved to the floor.

## Why It's Good For The Game
Fix bugs! Also turn an interesting snowflake crate into a proper subtype, perhaps encouraging it to be used elsewhere.

## Changelog

:cl:
fix: Independent Litieguai's organ freezer now properly freezes the provided organs
/:cl:

### DIFF
--- a/_maps/shuttles/shiptest/independent_litieguai.dmm
+++ b/_maps/shuttles/shiptest/independent_litieguai.dmm
@@ -225,21 +225,8 @@
 "fn" = (
 /obj/machinery/firealarm/directional/west,
 /obj/machinery/airalarm/directional/south,
-/obj/structure/closet/crate/freezer/surplus_limbs{
-	name = "organ freezer"
-	},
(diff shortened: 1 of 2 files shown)


MarkSuckerberg (approved):
very nice very nice

PR #2108, opened 2023-06-30
Hotspots no longer ignite reinforced floors

SapphicOverload:
## About The Pull Request

Hotspots can no longer cause turf fires on reinforced floors. Flamethrowers and other things that ignite the turf directly will still work, however.

## Why It's Good For The Game

Currently, the way turf fires work creates a lot of strange issues with certain gas reactions. TEG burn chambers can generate large amounts of CO2 from nothing (regardless of whether the combustion reaction used even makes CO2 at all) while consuming a large portion of the oxygen meant for whichever fuel you're using, which means you sometimes get a lot less out of your fuel than you should.

The CO2 from turf fires also causes problems with fusion. Because of how fusion works, adding more of it makes it much more likely to cool down and kill the reaction, making it more difficult to start than it needs to be.

My solution for this is simple: make hotspots unable to ignite reinforced floors, since they're already used mostly for burn chambers and such due to being fireproof. Flamethrowers and anything else that directly creates turf fires still work on them.

## Changelog

:cl:
tweak: reinforced floors can no longer ignited by hotspots
/:cl:

### DIFF
--- a/code/game/turfs/open/_open.dm
+++ b/code/game/turfs/open/_open.dm
@@ -1,3 +1,7 @@
+#define IGNITE_TURF_CHANCE 30
+#define IGNITE_TURF_LOW_POWER 8
+#define IGNITE_TURF_HIGH_POWER 22
+
 /turf/open
 	plane = FLOOR_PLANE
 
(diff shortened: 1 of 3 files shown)


MarkSuckerberg (approved):
Oh, sorry I took so long to review this, I assumed it was making all fires (not just turf fires) or something not able to ignite on reinforced floors. Good idea, they're made out of metal so they shouldn't be burning anyways, good catch.

PR #2117, opened 2023-07-01
Makes `sanitize_hexcolor()` more lax

MarkSuckerberg:
## About The Pull Request
Basically, it by default forced sanitised hexcolours into 3-hex instead of the default 6-hex, reducing the amount of possible colours SUBSTANTIALLY. As far as I can tell, there's nothing that really needs 3-hex from my investigation of every use of it, so I just set it to use 6-hex by default.

## Why It's Good For The Game
Adds 16773120 new colours for use in character creation. Yes.

## Changelog

:cl:
add: Colours in character creation are no longer crunched into three hex digits, allowing for 16773120 new colours to use for mutcolour, haircolour, etcetera.
/:cl:


Anderson-Cooper:
yeah

PR #2133, opened 2023-07-06
SSlag_switch from /tg/

MarkSuckerberg:
## About The Pull Request

Ports:
- tgstation/tgstation#62999
- tgstation/tgstation#59717

and adds a few shiptest-specific toggles as well to better take advantage of the system.

## Why It's Good For The Game

Recently lag has become a bit more of an issue here, and I think I would rather we be able to play than force ourselves to suffer while we slowly work on perf improvements.

## Changelog

:cl:
admin: Adds lag switch toggles from /tg/, ONLY use them when the lag becomes unbearable!
/:cl:


Erikafox:
lag is solved forever

MarkSuckerberg:
so true

Sun-Soaked:
image
no way I just saw this one while looking for atom comsigs

MarkSuckerberg:
we do a little lag mitigation

PR #2134, opened 2023-07-06
Fixes tritium fires

SapphicOverload:
## About The Pull Request

Tritium fires no longer break conservation of mass by consuming far more oxygen than intended. The burn rate has also been changed to scale with oxygen-fuel ratio instead of being a binary fast/slow reaction.

This also fixes some strange interactions with plasma combustion that were caused by it consuming so much extra oxygen.

## Why It's Good For The Game

Fire makes sense now, TEG power output is less inconsistent.

## Changelog

:cl:
tweak: tritium burn rate is based on oxygen-fuel ratio instead of being binary fast/slow
fix: fixed tritium fires breaking conservation of mass and consuming too much oxygen
/:cl:

### DIFF
--- a/code/__DEFINES/reactions.dm
+++ b/code/__DEFINES/reactions.dm
@@ -19,7 +19,7 @@
 
 #define NITRYL_FORMATION_ENERGY 100000
 #define NITROUS_FORMATION_ENERGY 10000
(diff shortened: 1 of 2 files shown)


MarkSuckerberg (approved):
cool cool cool

PR #2139, opened 2023-07-07
Bugfix for filled big mortars

Hibou48888:
## About The Pull Request

Fixes big mortar logic so they now properly check if the item they're dispensing to is full.

## Why It's Good For The Game

Big mortars try to check if the container they're filling is full to decide if they should dispense reagents or not. The only problem is this check wasn't working correctly, and was instead checking if the _mortar_ is full, meaning if you stuffed it full of plants and let it rip, it'd fill up the mortar and then essentially lock all those reagents inside itself since it would think every container you wanted to dispense to was full. Now they properly check if the target container is full, and your precious "medicinal herbs" are accessible once more.

## Changelog

:cl:
fix: big mortars now check the correct container for fullness before refusing to fill said container
/:cl:


MarkSuckerberg (approved):
woooo, finally, thankya

PR #2172, opened 2023-07-16
Nerfs Block Chance + Cooldown

Erikafox:
## About The Pull Request
Weapons with blockchance are mildly degenerate, due to how you can carry multiple, and they allow near nullification of ranged weapons if stacked enough. Thus I am nerfing their chances to fire and adding a cooldown. Shields are exempt from the cooldown because they're shields. 

## Why It's Good For The Game
Combat should be *more* lethal. Harder to cheese. That sort of thing. Wear armor and carry a shield if you want to avoid damage. Or buy a gun.

## Changelog

:cl:
add: blocking something now starts a short timer before you can block something again, unless you're using a shield.
balance: Block Chances have been altered (usually decreased) across the board. In tandem with the cooldown, don't expect to block as much automatic weaponry, while still being alright against slow attacks.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -103,7 +103,7 @@
 	var/block_chance_modifier = round(damage / -3)
 
 	for(var/obj/item/I in held_items)
-		if(!istype(I, /obj/item/clothing))
+		if(!istype(I, /obj/item/clothing) || get_active_held_item(I) || istype(I ,/obj/item/shield))
 			var/final_block_chance = I.block_chance - (clamp((armour_penetration-I.armour_penetration)/2,0,100)) + block_chance_modifier //So armour piercing blades can still be parried by other blades, for example
 			if(I.hit_reaction(src, AM, attack_text, final_block_chance, damage, attack_type))
 				return TRUE
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -103,7 +103,7 @@
 	var/block_chance_modifier = round(damage / -3)
 
 	for(var/obj/item/I in held_items)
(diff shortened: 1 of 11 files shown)


Sun-Soaked:
I'd leave the officer sabres at 50, if only because their whole gimmick is having pitiable damage and great melee block

to prevent 2 saber max block gaming, it might be a good idea to implement a check if the item is in the active hand before blocking (add a flag to shields and exempt them)
(though max block is also less of a problem now due to cooldown)

Erikafox:
I'll take a look at the sabres. Not sure how I'd implement a flag since I haven't done it before, but I'll take a looksie too.
If I can get that set up I'll adjust the balance some more.

Erikafox:
Alright, shields now block consistently, weapons only block if you're actively using them

Valorium-RB:
This enables a fun design space for parrying daggers and the like when we add more mechanics for SUNS duelist stuff, really cool.

Apogee-dev:
You know, regarding block chance, you could make stacking multiple sources have rapidly diminishing returns or just make it so the object with the highest block chance applies and all the others are ignored.

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -103,7 +103,7 @@
 	var/block_chance_modifier = round(damage / -3)
 
 	for(var/obj/item/I in held_items)
-		if(!istype(I, /obj/item/clothing))
+		if(!istype(I, /obj/item/clothing) || get_active_held_item(I) || istype(I ,/obj/item/shield))
```
MarkSuckerberg:
```suggestion
		if(!istype(I, /obj/item/clothing) || I == get_active_held_item() || istype(I ,/obj/item/shield))
```

MarkSuckerberg (changes requested):
just this

MarkSuckerberg (approved):
forgot i have to clear my change request to make this merge

PR #2177, opened 2023-07-18
prefixes now require admin intervention to change

ZephyrTFA:


Randomguy523:
Kind of wondering why changing the prefix should need an admin, especially since some people might want to run X ship without the default prefix, it's not super common but also not that rare. I've never really seen changing prefixes as problematic.

I personally only see this mechanically resolving edge-case false flags but those should probably be handled by admins anyways. Instead this will make any prefix change, even ones with minor impact to the round, force admins to greenlight every time.

If this was to become a thing, maybe to have an option before spawning the ship to change the prefix if the player doesn't want their ship to be the default prefix roundstart, then have this happen afterwards?

tgui/packages/tgui/interfaces/HelmConsole.js
```diff
@@ -51,16 +51,29 @@ const SharedContent = (_props, context) => {
...
-            currentValue={shipInfo.name}
-            disabled={isViewer}
-            onCommit={(_e, value) =>
-              act('rename_ship', {
-                newName: value,
-              })
-            }
-          />
+          <>
+            <Button.Input
```
tmtmtl30: is it worth adding a button like this when it can't be used by players anyway? it feels like clutter. most admins could just VV it anyway; maybe update_name() any time they VV the prefix?

thgvr:
> Kind of wondering why changing the prefix should need an admin, especially since some people might want to run X ship without the default prefix, it's not super common but also not that rare. I've never really seen changing prefixes as problematic.
> 
> I personally only see this mechanically resolving edge-case false flags but those should probably be handled by admins anyways. Instead this will make any prefix change, even ones with minor impact to the round, force admins to greenlight every time.
> 
> If this was to become a thing, maybe to have an option before spawning the ship to change the prefix if the player doesn't want their ship to be the default prefix roundstart, then have this happen afterwards?

There is no real reason for ship prefixes to be editable player side, at all. The ship prefixes exist to give others information on what kind of ship it is, and what faction. Letting it be freely changed, at any point, nullifies that a bit.

I can assure you taking admin action every time someone changes the prefix gets really, really old after a while. It's mainly why I asked for this

Randomguy523:
> There is no real reason for ship prefixes to be editable player side, at all. The ship prefixes exist to give others information on what kind of ship it is, and what faction. Letting it be freely changed, at any point, nullifies that a bit.

Prefixes like ISV, SSV, NTSV, only tell the people the faction, not the function of the ship (Barring maybe some cases like NTMV or similar). Freely changing it is a bit silly yes, rarely do you have a reason to mid-round. This is why I suggested it being a choice as the ship is spawned.

> I can assure you taking admin action every time someone changes the prefix gets really, really old after a while. It's mainly why I asked for this

I'd like more elaboration on this, are there more issues with changing prefixes I'm unaware of?

thgvr:
If the tags are unspecialized it should be considered an unspecialized ship.

We don't and never will support OC player factions/tags with ships. This is what independent is for. The other bit, it's annoying for the admin team to take action and tell players all of the time they can't change their faction tag. We already enforce things like this - restricting it like this is a better solution.

This will not stop players asking if they would like it changed.

Randomguy523:
> We don't and never will support OC player factions/tags with ships. This is what independent is for. The other bit, it's annoying for the admin team to take action and tell players all of the time they can't change their faction tag. We already enforce things like this - restricting it like this is a better solution.

Preventing non-standard tags is fair, but then why not have like a drop-down list during spawning of preset faction prefixes to pick where applicable (ships with large amounts of faction gear would still be set as X faction, no initial alternatives), and after spawning the ship that can't be changed without request later.

thgvr:
A ship's faction is not a fluid thing.

tgui/packages/tgui/interfaces/HelmConsole.js
MarkSuckerberg: Apogee had a good idea, too, if you want it to be a TGUI button, put it on the shuttle manipulator, not the helm

Draggeru:
Now does this separate the ship's name from the prefix? I can imagine it spamming admin logs if you go to change your ships name without altering the prefix

BarteG44:
Good luck trying to find the medship without IMV or ERRV

Martinpachu:
> A ship's faction is not a fluid thing.

i don't understand this,can there not be,say,an independant Skipper? since the creators of a ship aren't nessesarily always going to be the ones to use them,especially NT would seem to be the type to sell it's patterns of ships to non-employees

MarkSuckerberg:
> 
> i don't understand this,can there not be,say,an independant Skipper? since the creators of a ship aren't nessesarily always going to be the ones to use them,especially NT would seem to be the type to sell it's patterns of ships to non-employees
> 

no the skipper was very much intended as an nt only ship

if a ship was intended to be sold for civilian use it would be an ISV ship, there might be wiggle room for factions using civilian designs but not vice versa unless it's a stolen or retrofitted version which, if the riggs variants are anything to go by, just aren't warranted sadly

Sun-Soaked:
Faction options when buying a ship that swap out the prefix and some of the decals...

thgvr:
whar

PR #2183, opened 2023-07-20
fixes inability to spawn new ships on localhost

tmtmtl30:
## About The Pull Request

turns out i forgot to disable playtime checking on the TGUI window itself when the conditions for it activating weren't met (an active database and the USE_EXP_TRACKING config value set). oops! that's fixed now

## Why It's Good For The Game

how can we ship tests if our devs can't even test ships

## Changelog

:cl:
fix: The roundstart join menu now correctly ignores playtime restrictions if USE_EXP_TRACKING is disabled or no database is found.
/:cl:


Sun-Soaked:
I was manually spawning them in using the shuttle manip but this'll make it less scuffed

MarkSuckerberg (approved):
oopsie woopsie

PR #2195, opened 2023-07-24
Make plasma sheets produce 10x more gas when ignited.

ritorizo:
## About The Pull Request

Right now plasma sheet produce 10 mol of gas when ignited, with this pr they will produce 100mol.

### The balance argument
I think it would be cool if people vaporized plasma crystal to power their ship but right now you need about 1040 crystals to fill a tank (see image) to round start level (10398.4 mol).

With this pr 104 crystals will be required which I think is still kind of a lot considering where we want to take mining but still balanced.

**TLDR;** it would allow plasma sheets to be a real source of gas.

### Physics/lore argument
I also think this pr cool have cool implications lore wise since :

With this pr we could fill  a **10kL** tank **of gas**(see image) with ~100 sheets of plasma so **~2kL of crystal**. which would mean lore wise that crystal is a more effective (but perhaps less safe).

On the contrary it means that **without this** pr gas is the superior way of storing plasma being **2x more effective than crystal**, which is weird since most of the time storing matter as a solid is more efficient than storing it as a gas (expect for water because water like to crystalize weird).

**TLDR;** Right now plasma crystal is less dense than plasma gas in tank and that's weird we should change it.
orange2

## Changelog

:cl:
tweak: Plasma produce 10x more plasma gas.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/sheets/mineral.dm
+++ b/code/game/objects/items/stacks/sheets/mineral.dm
@@ -205,7 +205,7 @@ GLOBAL_LIST_INIT(plasma_recipes, list ( \
 		return ..()
 
 /obj/item/stack/sheet/mineral/plasma/fire_act(exposed_temperature, exposed_volume)
-	atmos_spawn_air("plasma=[amount*10];TEMP=[exposed_temperature]")
+	atmos_spawn_air("plasma=[amount*100];TEMP=[exposed_temperature]")
 	qdel(src)
 
 /obj/item/stack/sheet/mineral/plasma/fifty


zestybastard:
How does this affect plasmaflooding via a welder/solid plasma? I like this idea a lot, and have experimented with gaseous plasma conversion for phorid gameplay, but I do wonder..

ritorizo:
It make solid plasma flooding a lot more dangerous, basically you can't open door and go like "oh it's fine it will just burn a big part of the ship and go away" you have to actually space the flood to fix it.

Also not to sound like a psychopath tider but I was disappointed to find that you can't ignite crysal plasma with a laser gun

ritorizo:
And also the flood may resolve itself by breaking the floor a lot of the time

zestybastard:
That's probably fine then. It's fairly easy to set up a vacuum on your ship capable of doing this conversion, so people should be able to utilize the change easily. Possibly worth adding a room to outpost docking areas that would allow you to do this there, and pipe the resulting gas to your ship?

goober3:
do full stacks make 5000 mols now?

ritorizo:
> do full stacks make 5000 mols now?

Yes with this pr they will.

tmtmtl30:
Where are you getting "2kL of crystal" from? _Technically_ speaking, one bar of any material is 2000 cubic centimeters in in-game units, or 2 liters, making a full stack of plasma bars 100 liters (a lot to hold in your backpack, and implying that you physically can't store more than 25 full stacks of materials in a tile (because tiles have an atmos volume of 2500L (implying dimensions 1m x 1m x 2.5m)), but whatever).

There's also the plasma _reagent_ to consider, which will boil when heated beyond 50C. That creates a number of moles equal to the volume of plasma reagent; each bar has 20 plasma reagent inside it, giving 1000 moles of plasma in a stack. Obviously, these numbers weren't in agreement _anyways_, but it feels weird for converting to a reagent to be less effective than converting from the bars directly. (The contradiction inherent in reagents -- the conflation of moles with volume -- also rears its head here. Humans have 560u of blood inside them, giving 1u ~= 1 centiliter, which matches with volumes like 100u in a large beaker, but it causes problems everywhere else. Using the aforementioned units it implies that a plasma sheet gets 5x smaller once you turn it into liquid).

Also, that's just a fucking lot of plasma. The fuel economy is strange, of course, but I don't know if this really helps it much. How long can a ship run a TEG off of a plasma tank like one of those? I know we use typically turfs for TEG plasma gas storage, but I'm curious.

MarkSuckerberg:
We had a discussion on this, we just don't think it's something we want. The argument here is for realism and *maybe* balance, but realism isn't important and the balance is affecting something that's too far gone to worry about and is pending a rework anyways. Plus, I never intended the massive pressure tanks to become the de facto standard, personally I think they're weirdly made and rather uninteractive, not to mention store far too much gas.

ritorizo:
Yes, I just noticed I forgot to reply to tmt but yes you're right, I fucked up my calculation and factored in a 10 in the crystal density thing somehow and what i'm arguing for is actually already the case. Thank you for noticing.

PR #2218, opened 2023-07-30
Blind Runechat Tweaks

Skies-Of-Blue:
## About The Pull Request

Makes runechat (and balloon popups) appear above the blind layer, making them 'visible' to blinded players. This is my first PR without assistance, so please let me know if something has gone horribly wrong in the process.

## Why It's Good For The Game
Blind people should be able to determine where a voice is coming from if it's within earshot! Major QoL, means you won't have to play minesweeper with your mouse to find out where the person who just spoke is in the room at any given time.

## Changelog

:cl:
tweak: runechat and balloon popups are now visible to blinded players
/:cl:


MarkSuckerberg (approved):
I like it! I think I've heard the same idea thrown around a few times, but this is the first time someone actually did it!

PR #2238, opened 2023-08-03
Access restricts the Aegis bridge

Apogee-dev:
## About The Pull Request

Does what it says on the tin- adds access restrictions and finer door controls to the Aegis's bridge. The Aegis still has a larger rework on the horizon when SUNS content is finished, but for now this is just a quick fix to a recurring issue.

## Why It's Good For The Game

Bridges, as a rule, should be access restricted to officer roles on any ship of sufficient size so it doesn't just turn into another assistant den. The Aegis is no exception.

## Changelog

:cl:
tweak: Added access restrictions to the bridge on the Aegis
/:cl:

### DIFF
--- a/_maps/shuttles/shiptest/syndicate_aegis.dmm
+++ b/_maps/shuttles/shiptest/syndicate_aegis.dmm
@@ -1719,6 +1719,7 @@
 /obj/effect/turf_decal/siding/wood{
 	dir = 6
 	},
+/obj/machinery/airalarm/directional/south,
 /turf/open/floor/mineral/plastitanium/red{
 	icon_state = "plastitanium"
 	},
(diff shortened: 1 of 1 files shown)


Erikafox (approved):

Draggeru:
I am very vocally against this, we addressed this in the previous Aegis PR. If you are intending to strongarm this you must change the Aegis's bridge to have windows that connect the main hallway so the crew can still verbally speak with those in the bridge without having to use a radio.

spockye (approved):

spockye:
> I am very vocally against this, we addressed this in the previous Aegis PR. If you are intending to strongarm this you must change the Aegis's bridge to have windows that connect the main hallway so the crew can still verbally speak with those in the bridge without having to use a radio.

A good point. Mabye add a window or two, with privacy shutters?

Apogee-dev:
> > I am very vocally against this, we addressed this in the previous Aegis PR. If you are intending to strongarm this you must change the Aegis's bridge to have windows that connect the main hallway so the crew can still verbally speak with those in the bridge without having to use a radio.
> 
> A good point. Mabye add a window or two, with privacy shutters?

Yeah, it's a fair point. I'll do that.

PR #2243, opened 2023-08-03
Adds the Syndicate gas mask as a loadout option.

Higgin:
## About The Pull Request

Tosses the Syndicate gas mask (the red and black one) on the loadout as an option under accessories.
## Why It's Good For The Game

 It's otherwise lacking on some ships that might have it and is drippy (the emoji meme mask also acts as a gas mask for the same function so I don't consider this out of pocket for a cosmetic.)

## Changelog

:cl:
add: Adds the Syndicate gas mask (the red and black one) to the loadout menu as an accessory.
/:cl:

### DIFF
--- a/code/modules/client/loadout/loadout_accessories.dm
+++ b/code/modules/client/loadout/loadout_accessories.dm
@@ -102,3 +102,8 @@
 	display_name = "face with tears of joy mask"
 	path = /obj/item/clothing/mask/joy
 	slot = ITEM_SLOT_MASK
+
+/datum/gear/accessory/syndiemask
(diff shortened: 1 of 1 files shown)


thgvr:
Factional stuff is not gonna be allowed in our current loadout. In the future if it happens, it'll be tied to factional background or other such things in character creation. Loadouts are to be used for purely cosmetic or very, very small things (like lighters), and I'll be doing an audit for loadout stuff in general pretty soon.

This mask is something that's pretty directly Syndicate right now and definitely shouldn't end up on ships of other factions with how things currently stand.

PR #2246, opened 2023-08-03
LOOC Fixes

MarkSuckerberg:
## About The Pull Request
See changelog

## Why It's Good For The Game
LOOC is used fairly frequently here, so I might as well keep it up to scratch with normal OOC at least.

## Changelog

:cl:
add: You can now configure LOOC and mentorsay in chat tabs
fix: (Un)Muting LOOC now actually works.
admin: You can now change the global LOOC colour. Because, why not?
admin: LOOC now shows up in the OOC tab of the logging panel for players
/:cl:


MarkSuckerberg:
also I just want to say this came to my attention because LOOC checked if you had the bwoink sound muted instead of the LOOC mute toggle enabled. I will not elaborate

thgvr:
it would be really epic if LOOC was a separate section toggle like OOC is for tabs.

BarteG44:
rune chat for LOOC would be really nice too. the most frequent use for it is to talk to newer players but most of them don't look in the chat tab

MarkSuckerberg:
> it would be really epic if LOOC was a separate section toggle like OOC is for tabs.

yeah that might be cool, probably not terribly hard

> rune chat for LOOC would be really nice too. the most frequent use for it is to talk to newer players but most of them don't look in the chat tab

possibly, no idea how I'd show that as LOOC versus IC speaking though, though I'd be down to hear ideas

BarteG44:
i was thinking that it could look something like goon
obraz
maybe give it a different color too

MarkSuckerberg:
Could work, also, I made mentorchat and LOOC both configurable in chat tabs, updating the PR desc

MarkSuckerberg:
Added an attempt at runechat LOOC, untested, since I'm away from my windows PC atm

PR #2253, opened 2023-08-05
Ship Select now clearly and visibly shows what faction a ship is on both purchase and late join.

ZephyrTFA:

PR #2254, opened 2023-08-05
Ship factions are now in their tags.

TheNeoGamer42:
## About The Pull Request
Does what it says on the tin. I had basically finished it before I was told tags shouldn't be used this way, despite being told some time earlier that I should do it by other staff, so I am PRing it anyways, if it gets closed, it gets closed and I won't bother again.

## Why It's Good For The Game
Seeing the ship's faction is generally a good thing given people shouldn't be playing a ship as a faction it wasn't designed for.
## Changelog

:cl:
add: Each currently added ship has it's intended faction listed in its tags.
/:cl:


Erikafox:
We don't want factions in tags.
+ #2253 is a better implementation

PR #2276, opened 2023-08-11
Cleans Melter Anomalies a little

Erikafox:
## About The Pull Request
they should no longer melt their cores

## Why It's Good For The Game
bugfix

## Changelog

:cl:
fix: melters should avoid melting themselves now.
/:cl:


Zevotech:
Could you also add some more balancing to the plasma soul anomaly? It really shouldn't work through walls because it ends up next to a ship or a cave and often kills someone without them even being aware of it. It also kills IPCs REALLY hard for some reason. Unsure if its intended, but the toxin damage they deal is extremely lethal to IPCs.

Erikafox:
bug: they seem to no longer melt anything.

PR #2289, opened 2023-08-16
Kills the Icecropolis Ruin

spockye:
## About The Pull Request
this PR kills the icecropolis ruin.
murder....

## Why It's Good For The Game
why? because we want to move away from having ghost roles, which is the entire purpose of the icecropolis

## Changelog

:cl:
del: Removed the Icecropolis, and the related code
/:cl:


Latentish:
goodbye old friend. rest in fucking pieces

Valorium-RB:
The Icecropolis is the one with the kobolds, right? Couldn't we just remap it without the kobolds and keep it in, considering it's one of our few real "megadungeon" style ruins?

Latentish:
> The Icecropolis is the one with the kobolds, right? Couldn't we just remap it without the kobolds and keep it in, considering it's one of our few real "megadungeon" style ruins?

I thought you said that the lore team wanted to move away from the necropolis from being represented in the way this ruin does. Like in general. Also you could remove them, but the ruin would still be weird and fucked up. Frankly I feel like it really would need to be totally remapped to be any good. and I'm not particularly planning on doing that anytime soon. But I do agree, there aren't really any other megadungeon ruins.

Valorium-RB:
I suppose. Maybe if it were remapped to get rid of the explicit Necropolis theming, but that would probably be more trouble than it's worth.

Latentish:
> I suppose. Maybe if it were remapped to get rid of the explicit Necropolis theming, but that would probably be more trouble than it's worth.

Frankly it's a matter of how much content that doesn't really fit the server in exchange for gameplay, while we don't have something to replace it. I requested this removed because it's aged poorly and doesn't fit the server. Although we are sorely low on ruins.

PR #2306, opened 2023-08-22
Fixes an abundance of mapping icon sprites in flora and mineral walls as well as invisible trees on sandplanet

Zevotech:
# if there is a broken flora or wall sprite you've come to the right PR
## About The Pull Request
A bunch of flora that previously had bugged icons in mapping now actually properly load their icons in SDMM. They still generate perfectly ingame, but now mapping with them should be a lot easier. Most ash flora use the item instead of the plant because the code and icon names for them are complicated and I don't feel like refactoring that code entirely just to fix something that only affects SDMM.
IN SDMM:
image
INGAME:
image

Barren dead trees also used to be invisible ingame due to a subtyping issue! they are now back on sandplanets in all their barky purple glory. Also fixed the whitesands grass icon in SDMM because a typo turned it into a missing texture.
image

UPDATE! I've also fixed the mineral wall icons in SDMM and updated them to their new wall sprites. What was once a mess of outdated icons, missing icons, and the occasional up to date icon is now consistent across the board in SDMM. This means any issues with wall sprites not smoothing or appearing as the wrong type should be reported on this PR so I can fix them.
IN SDMM:
image
INGAME:
image

## Why It's Good For The Game
Mapping planetary environments is now easier, which should hopefully improve ruin quality!
Also, invisible things on planets are bad.

## Changelog

:cl:
fix: flora sprites now properly work in SDMM instead of being missing textures
fix: dead barren trees on whitesands are no longer invisible due to having their icon state overridden
fix: whitesands grass now has its sprite in SDMM thanks to a typo fix
/:cl:


rye-rice (approved):

Zevotech:
this will benefit the 4 or so people that actually map ruins

PR #2335, opened 2023-09-06
cybernetic organs no longer permanently begin to shut down

Skrem7:
## About The Pull Request

Removes the chance of an EMP effect shutting down cybernetic organs based on its intensity aka rng

## Why It's Good For The Game

I would prefer to not force IPCs to have to ahelp for revival if they get hit by one loose ion (gross)

## Changelog

:cl:
balance: Cybernetic organs can no longer require replacement due to EMPs (they do not suffer permanent damage)
/:cl:


meemofcourse:
wtf buffs ipcs even more?!!!!

Martinpachu:
..buff ipcs...

Skrem7:
> wtf buffs ipcs even more?!!!!

is this unironic

Zevotech:
patching bugs with a feature /=/ buffing a feature

regardless we should buff ipcs more (synthahol when)

meemofcourse:
> is this unironic

yes and i think we should make ipcs run on watermils

Skrem7:
> patching bugs with a feature /=/ buffing a feature

technically not a bug and entirely intentional

> regardless we should buff ipcs more (synthahol when)

no they get to ignore several game mechanics

Zevotech:
intentional? im about to be intentionally violent to whoever added that

ArdentSpark:
This not only removes EMP'd organ's from IPC's but also from standard cybernetic organs that organic crewmembers have had implanted as far as I can tell.

If you want to protect IPC's, I think this should work. Give them an "emp_vulnerability = 0" for each _/obj/item/organ/**BLAHBLAHBLAH**/cybernetic/upgraded/ipc_, as they're inheriting the base EMP vulnerability of the Cybernetic organs.

Skrem7:
> This not only removes EMP'd organ's from IPC's but also from standard cybernetic organs that organic crewmembers have had implanted as far as I can tell.

This still implements EMP effects on organs, they just won't begin to permanently shut down if EMP'd very hard. The idea is for it to suck, not for one EMP to force a full organ replacement. They're too easy to get and already slap cybernetics hard enough, why stack more on top.

> If you want to protect IPC's, I think this should work. Give them an "emp_vulnerability = 0" for each _/obj/item/organ/**BLAHBLAHBLAH**/cybernetic/upgraded/ipc_, as they're inheriting the base EMP vulnerability of the Cybernetic organs.

90% sure this doesn't work but emp_act is a goofy proc and I haven't looked at its base call for months

ArdentSpark:
That's true, EMP does do a base amount of damage to cybernetic organs, but the downside of having them, is well, their vulnerability to EMP's, I don't think removing this across the board is that great, unless it's our last option. (IPC's being stuck in revival hell is not good)

From what I can see emp_vulnerability is the hook that ties directly into the code you're deleting, and by making the vulnerability 0, it should have the same effect for IPC's as not deleting that code entirely. Though, I haven't tested it.

Skrem7:
> That's true, EMP does do a base amount of damage to cybernetic organs, but the downside of having them, is well, their vulnerability to EMP's, I don't think removing this across the board is that great, unless it's our last option. (IPC's being stuck in revival hell is not good)

It's not removed.

image
image
image
image
image

I could also post ears, eyes, and implant emp effects, those aren't touched at all.

ArdentSpark:
Apologies for not making it clear, but I did understand that, I was talking in relation to the damage over time to the cybernetics post-EMP which honestly, is the most significant EMP effect, Taking 10 toxic damage isn't really that much in the grand scheme of things. I know you didn't touch the eyes, ears et cetera.

I understand that's the entire point of what you're doing here, is to actively remove this mechanic, which is fine? I just didn't realise at first it was your end goal, as I focused on the [Why this is good for the game section] too much.
But I do see now that my suggestions go against the core goal of the PR instead of resolving the IPC issue a different way, so I'll stop suggesting alternatives

Skrem7:
> Apologies for not making it clear, but I did understand that, I was talking in relation to the damage over time to the cybernetics post-EMP which honestly, is the most significant EMP effect, Taking 10 toxic damage isn't really that much in the grand scheme of things. I know you didn't touch the eyes, ears et cetera.

I'd 100% be down with upping "oh no 10 toxin damage!" effects since they're all p-lame relative to what they COULD feasibly be. But, as you pointed out, it's not the goal of the PR. I'm just curious to see how the EMPs turn out after this change.

ArdentSpark:
Greatest enemy of testers
Scope

Yeah not a bad idea if maintainers say no to the core idea if you need to barter

MarkSuckerberg (approved):
this is fine, I'll merge this, but some thoughts--

I honestly don't want to make the species that revives even more trivially than organics more powerful healthwise, but, like you pointed out to me, it kind of sucks to just be dead permanently because you're on a ship that can't print new synthorgans, especially when ion storms are a common thing

basically, don't want to buff IPCs healthwise but on the other hand, this is a shitty way to hold them back and it's a fundamental brokenness where this doesn't really matter in the end, it'll be fixed later on

Skrem7:
> basically, don't want to buff IPCs healthwise but on the other hand, this is a shitty way to hold them back and it's a fundamental brokenness where this doesn't really matter in the end, it'll be fixed later on

Like I said, the immediate negative effects of EMP could be spiked up.

Would also help if the piss-easy to revive species were more of a wet noodle protection-wise but I don't actually know where their damage coefficients are at.

PR #2374, opened 2023-09-25
Crate shelf

GenericDM:
## About The Pull Request
This PR intends to add a new method of organizing and storing items in a way that allows ships to keep more stuff off the floor.
githubcrate

Todo:

- [x] Add shelves
- [x] Add the ability for shelves to hold crate
- [x] Add the ability for crates to be removed from shelves
- [x] Arbitrarily tall shelves?
- [x] Add a way to craft them ingame
- [x] Add a way to disassemble them ingame
- [x] Add a check on unloading to ensure they can't be stacked forever (Kapu's turf.Enter() changes may be ported for this)
- [x] Fix (or mitigate) the layering problems
- [x] Bring the code up to consistent standards

## Why It's Good For The Game
I believe that adding more options for organization is a good thing.
Specifically, this will ideally make players more likely to sort items into crates, rather than just dumping everything in.
I built this with balance in mind, so once finished, there should be no way to abuse this to get infinitely dense storage, or anything like that.

## Changelog

:cl: Generic DM, PositiveEntropy
add: Shelves! For crates!
fix: Racks can no longer be stacked infinitely.
/:cl:


code/game/objects/structures/crates_lockers/crates.dm
```diff
@@ -45,11 +45,27 @@
...
+	if(!isturf(user.loc) || user.incapacitated() || user.body_position == LYING_DOWN)
+		return // If the user is in a weird state, don't bother trying.
+	if(get_dist(drop_atom, src) != 1 || get_dist(drop_atom, user) != 1)
+		return // Check whether the crate is exactly 1 tile from the shelf and the user.
+	if(istype(drop_atom, /turf/open) && istype(loc, /obj/structure/crate_shelf) && user.Adjacent(drop_atom))
+		var/obj/structure/crate_shelf/shelf = loc
+		return(shelf.unload(src, user, drop_atom)) // If we're being dropped onto a turf, and we're inside of a crate shelf, unload.
+	if(istype(drop_atom, /obj/structure/crate_shelf) && isturf(loc) && user.Adjacent(src))
+		var/obj/structure/crate_shelf/shelf = drop_atom
+		return(shelf.load(src, user)) // If we're being dropped onto a crate shelf, and we're in a turf, load.
```
MarkSuckerberg:
```suggestion
		return shelf.unload(src, user, drop_atom) // If we're being dropped onto a turf, and we're inside of a crate shelf, unload.
	if(istype(drop_atom, /obj/structure/crate_shelf) && isturf(loc) && user.Adjacent(src))
		var/obj/structure/crate_shelf/shelf = drop_atom
		return shelf.load(src, user) // If we're being dropped onto a crate shelf, and we're in a turf, load.
```

code/game/objects/structures/crateshelf.dm
```diff
@@ -0,0 +1,136 @@
...
+				if(crate.open()) // Break some open, cause a little chaos.
+					crate.visible_message("<span class='warning'>[crate]'s lid falls open!</span>")
+				else // If we somehow fail to open the crate, just break it instead!
+					crate.visible_message("<span class='warning'>[crate] falls apart!")
+					crate.deconstruct()
+			if(3) // Break that crate!
+				crate.visible_message("<span class='warning'>[crate] falls apart!")
+				crate.deconstruct()
+	shelf_contents.Cut()
+	return ..()
```
MarkSuckerberg: I'd move this all to `/proc/deconstruct`. The destroy for this should ideally just be `QDEL_LIST(shelf_contents)` since we don't want side effects in Destroy.

code/game/objects/structures/crateshelf.dm
```diff
@@ -0,0 +1,136 @@
...
+
+/obj/structure/crate_shelf/Destroy()
+	var/turf/dump_turf = drop_location()
+	for(var/obj/structure/closet/crate/crate in shelf_contents)
+		crate.layer = initial(crate.layer) // Reset the crates back to default visual state
+		crate.pixel_y = initial(crate.pixel_y)
+		crate.forceMove(dump_turf)
+		step(crate, pick(GLOB.alldirs)) // Shuffle the crates around as though they've fallen down.
+		crate.SpinAnimation(rand(4,7), 1) // Spin the crates around a little as they fall. Randomness is applied so it doesn't look weird.
+		switch(pick(1, 2, 3)) // Randomly pick whether to do nothing, open the crate, or break it open.
```
MarkSuckerberg:
```suggestion
		switch(pick(1, 1, 1, 1, 2, 2, 3)) // Randomly pick whether to do nothing, open the crate, or break it open.
```
Maybe something more like this? I don't know if you'd want breaking and opening to be as common as nothing happening, just wanted to show that it's possible. Up to you

code/game/objects/structures/crateshelf.dm
```diff
@@ -0,0 +1,136 @@
...
+		handle_visuals()
+		return TRUE
+	return FALSE  // If the do_after() is interrupted, return FALSE!
+
+/obj/structure/crate_shelf/deconstruct(disassembled = TRUE)
+	if(!(flags_1&NODECONSTRUCT_1))
+		density = FALSE
+		var/obj/item/rack_parts/shelf/newparts = new(loc)
+		transfer_fingerprints_to(newparts)
+	qdel(src)
```
MarkSuckerberg:
```suggestion
	return ..()
```

MarkSuckerberg (approved):
Perfect!

PR #2441, opened 2023-10-31
Removes deprecated ruin blacklists and other .txt files.

Zevotech:
## About The Pull Request
Removes the deprecated ruin blacklists and some text files that are no longer in use. Changes the comment in the map datum files to point people to the map catalogue instead so they know to add their ruin to it now.

## Why It's Good For The Game
I was asked by latency to remove the deprecated ruin blacklists so people would stop editing them, and mark pointed me to a few other deprecated files in the same area that can go with them.

## Changelog

:cl:
del: Ruin blacklist text files.
del: Misc. deprecated text files.
/:cl:


MarkSuckerberg (approved):
so long, gay bowser

PR #2458, opened 2023-11-07
Shields, in this economy?

ZephyrTFA:
Adds the ability to construct a shield generator for your ship.
This is a blatant precursor for SvS combat, however for now it's main goal is defending you from those nasty overmap events.
I can almost guarantee that they are current over-powered for this because I'm not able to test this locally because I refuse to downgrade byond.

:cl:
add: Ship Shield Generators. Check your local autolathe.
/:cl:


thgvr:
What's up with this? What does it do, how does it work, where does it fit? Haven't heard anything from you.

thgvr:
I swear we'll get to this after mining stuff. It seems good initially reviewing though

ZephyrTFA:
it's been three months

PR #2473, opened 2023-11-11
Stops gas overlays from inheriting turf color

thgvr:
## About The Pull Request
Adds RESET_COLOR appearance_flag to stop gas overlays from inheriting turf color

## Changelog

:cl:
add: Gas overlays should now look normal on turfs with a color var set.
/:cl:


Erikafox:
you should stop cleaning overlays from doing the same thing

PR #2477, opened 2023-11-13
"Fixes" Subshuttle Error

Erikafox:
guard clause :)


MarkSuckerberg (approved):
grr

MarkSuckerberg:
this would be better done by making subshuttle spawns pass in "force" to the creation of the ships but this is fine

PR #2490, opened 2023-11-17
The new [SOMEWHAT NEAT] Nanotrasen Outpost! 

goober3:
## About The Pull Request
todo:
- [x] make hangars
- [x] make the rock walls indestructible
- [x] expand the bar
- [ ] open up more exterior?

Adds a brand new Nanotrasen-controlled outpost situated on an ice planet. 

Features include:
- A neat little bar.
- Two shops.
- A brig, and a security checkpoint.
- An ERT bay.
- Old maintenance tunnels containing The Horrors. (mothroach x1)
- A nice view.

## Images
<details><summary>The Outpost</summary>

a91dad31c20289ea25ea2d1f7739ccdc

image

image

</details>

<details><summary>Hangars</summary>

image
20x20

image
40x20

image
56x20

image
40x40

image
56x40
</details>

## Changelog

🆑
add: A new Nanotrasen outpost, located on an ice planetoid, has been added.
/🆑


Latentish (changes requested):
Please make the bar a bit bigger. Something like this size. The reason I ask is because the outpost can get crowded at times, and the bar serves as a place for gathering. Your current bar comfortably sits like, 2 people and a bartender.
image

Make them rocks indestructible.

Latentish (dismissed):
- Either commit to the facility being aboveground or belowground. Currently the surface is in this weird place of having a stream going through the middle of a blobish cavern full of trees and some oddly evenly spaced bushes and such. The river starts and ends in some rock, not really implying an underground stream or anything else. 
- The outside areas need more detail and greebling. It feels too clean for nature, and it feels out of place in a cavern.
- I adore the idea of these side areas that are only openable by an admin intervention, however I feel that you miss out on having an outside area that players can go walk around in that isn't the padding between walkways.
- The bar is far too small. It currently sits 2 people and a bartender. With the only seating at the bar itself. I'd like to see more seating away from the bar, more seating at the bar, and a good 3 tiles in both north and west directions. Even better would be if it's an odd or non-square shape.
- Rename the magical smes

tmtmtl30:
the turfs in the outpost make space when you remove them. i'm not sure why, and they're indestructible, but they still shouldn't do that. it might be because they're missing the baseturf zflag (which is really important for planetary outposts and needs to be changed) but changing that at runtime didn't fix it, because the turfs had /turf/open/space/basic in their baseturf stack, above /turf/baseturf_bottom. need to look into it more

FalloutFalcon:
> the turfs in the outpost make space when you remove them. i'm not sure why, and they're indestructible, but they still shouldn't do that. it might be because they're missing the baseturf zflag (which is really important for planetary outposts and needs to be changed) but changing that at runtime didn't fix it, because the turfs had /turf/open/space/basic in their baseturf stack, above /turf/baseturf_bottom. need to look into it more

#2526

Latentish:
marge???

PR #2535, opened 2023-11-28
fixes some backend shipjoin stuff

tmtmtl30:
## About The Pull Request

admin + DB flag exemptions weren't respected by the ship joining UI, but _all_ players were assumed to have the requisite playtime to spawn / join a ship by the code that actually handled joining. this fixes both, technically fixing an href exploit i suppose

## Why It's Good For The Game

bugs bad + consistency good. i am bad at web dev

## Changelog

:cl:
fix: Small playtime-based ship join restriction edge cases fixed.
/:cl:

### DIFF
--- a/code/modules/jobs/job_exp.dm
+++ b/code/modules/jobs/job_exp.dm
@@ -45,6 +45,7 @@ GLOBAL_PROTECT(exp_to_update)
 		return FALSE
 	if(CONFIG_GET(flag/use_exp_restrictions_admin_bypass) && check_rights_for(src, R_ADMIN))
 		return FALSE // if admin exemption is enabled, and client is an admin, let them through
+	return TRUE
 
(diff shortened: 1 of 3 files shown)


MarkSuckerberg (commented):
is the playtime restriction limits even checked at all during the actual purchase?

tmtmtl30:
i believe so; the chain is something like attemptlatejoin -> isjobeligible (or isjobavailable? or something?) -> has_job_playtime. dunno if those are the exact names since i can't search the codebase easily right now

really, i should probably move the relevant calculations into ui_static_data() and just send them to the UI so that it's all centralized. but

tmtmtl30:
turns out that shipowners can infinitely open slots on their ships, because joining a ship decrements the slot counter. this is bad. need to fix it

MarkSuckerberg (approved):
I'll let you get that in a different PR, this is pretty important to have

PR #2555, opened 2023-12-04
stops outposts placing on top of space

tmtmtl30:
## About The Pull Request

planetary outposts currently have /turf/open/space/basic underneath all their turfs even if the baseturf ztrait is set correctly. this fixes that, by making outpost map templates (elevator / hangar templates included, though that shouldn't make a difference due to the way they're used) no longer spawn "on top of" turfs: now they replace them instead

## Why It's Good For The Game

randomly spaced turfs are bad

## Changelog

:cl:
fix: fixes planetary outposts from having space underneath all the turfs
/:cl:

### DIFF
--- a/code/modules/overmap/objects/outpost/outpost_types.dm
+++ b/code/modules/overmap/objects/outpost/outpost_types.dm
@@ -2,6 +2,13 @@
 	Map templates
 */
 
+/datum/map_template/outpost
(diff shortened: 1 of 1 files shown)


MarkSuckerberg (approved):
interesting, makes sense

PR #2556, opened 2023-12-04
Removes the (Independent) Lieutenant job

Zevotech:
## About The Pull Request
Removes the wildly inconsistent and unmaintained whitesands relic that is the Lieutenant job (formerly known as the Solgov Representative. Yes, that one). Pieces of its clothing were shoved into various ships or outfits, and it was only ever used on the Shetland which is admittedly very dated and in need of modernizing. This should not touch the LP Lieutenant job or the Twinkleshine Lieutenant job.
The Shetland now has a First Officer instead of a Lieutenant, and now has two separate lockers for the FO and Captain instead of just one for the Lieutenant. That poor, poor captain, locker-less for so long..

## Why It's Good For The Game
image
This job has visible rust from its existence in the whitesands era and tries and fails to fill in at what the First Officer job does better. The independent captain/command resprite also broke its jumpsuit sprites due to the type paths overlapping.

## Changelog

:cl:
del: Removes the Lieutenant job datum, locker and clothing
add: The Shetland now has a First Officer and lockers for both them and the captain now
/:cl:


_maps/configs/independent_shetland.json
```diff
@@ -22,8 +22,8 @@
 			"officer": true,
 			"slots": 1
 		},
-		"Lieutenant": {
-			"outfit": "/datum/outfit/job/lieutenant",
+		"First Officer": {
```
MarkSuckerberg: wait why does the shetland have both a lieutenant/first officer and a first mate anyways
just yeet one of them

Zevotech: Oh I didn't see the first mate slot when I was doing this. Whoops.

PR #2560, opened 2023-12-05
Fixes tackles causing semi-permanent slowdown, and not being able to trigger grabs

### LINKED ISSUE #2340: [BUG] Kepori Post Pounce Movement Slow
## Testmerges:

 Local commit: 62a1b3ecbb81e85cddba26f36f27928545a4a715
TGS version: 5.13.7
DMAPI version: 5.6.1

## Reproduction:
As a Kepori, pounce on someone. It may cause the character to become permanently slow despite being fully healthy. First experienced as FBP. Aheal may or may not fix.

STEPS
1: As Kepori, prepare a pounce. Nothing in hand, at run speed.
2: Pounce a player.
3: Observe speed loss.

This so far has only been noted in one round, but I will be attempting to recreate this bug every round I can.
I did note a drop in mood rapidly, but it may be unrelated.

Attached his med scanner results.

This is my first report, so any tips on future reports would be greatly appreciated.

SomeguyManperson:
## About The Pull Request

Tackles allow setGrabState() to be called without confirming a grab exists, this causes usually grab-related stuff to trigger, mainly the slowdown for having someone grabbed, without actually having grabbed anyone. As far as I can tell, this would maybe only occur if you succeeded a tackle roll that caused a grab when stamcritting yourself

The channel for standing up, however, was still being applied after the tackle purges the attacker's knockdown (and you cannot, unfortunately, grab people while prone), preventing either grab roll from ever causing a grab. Now, you stand up instantly if your tackle would result in a grab.

image

## Why It's Good For The Game

fixes #2340 

## Changelog

🆑 
fix: Tackles no longer have a chance to give permanent slowdown until you grab someone
fix: Tackles which would result in a grab now actually result in a grab
/🆑

### DIFF
--- a/code/datums/components/tackle.dm
+++ b/code/datums/components/tackle.dm
@@ -208,26 +208,30 @@
 			to_chat(target, "<span class='userdanger'>[user] lands an expert tackle on you, knocking you down hard and maintaining a passive grab!</span>")
 
 			user.SetKnockdown(0)
+			user.get_up(TRUE)
 			user.forceMove(get_turf(target))
 			target.adjustStaminaLoss(40)
 			target.Paralyze(5)
 			target.Knockdown(30)
 			if(ishuman(target) && ishuman(user))
 				INVOKE_ASYNC(S.dna.species, TYPE_PROC_REF(/datum/species, grab), S, T)
-				S.setGrabState(GRAB_PASSIVE)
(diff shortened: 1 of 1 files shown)


MarkSuckerberg (approved):
oh, neat

PR #2590, opened 2023-12-17
Nerfs legion core implanting to not be an aheal

Zevotech:
## About The Pull Request
The legion core implant now gives you a potent heal of all four damage types (-100 brute -100 burn -50 tox -50 oxy) instead of a literal aheal. It now also deals 10 clone damage as a drawback to organics/FBPS (IPCS excluded because they cant use clone damage medicine).
Due to how adjustBruteLoss and adjustBurnLoss work, the implanted core no longer heals mechanical bodyparts, making it mostly useless for IPCs and FBPs only healing oxygen and toxin damage.

## Why It's Good For The Game
1. Player accessible aheals are not good and encourage exploiting to cure ailments or gain an advantage.
2. Synthetics could use self-surgery to implant legion cores on the go for a safety net heal. While not necessarily bad, it was insanely powerful as an aheal and negated the requirement of stabilizing the core and getting another person to put it in you.
3. It had literally no drawbacks. A strong consumable healing ability is cool, but it should come with a cost.

## Changelog

:cl:
fix: legion core implanting no longer aheals you on use
add: legion core implant now just does a potent organic heal with minor clone damage when used
/:cl:

### DIFF
--- a/code/modules/mining/equipment/regenerative_core.dm
+++ b/code/modules/mining/equipment/regenerative_core.dm
@@ -68,8 +68,6 @@
 		owner.adjustToxLoss(-50)
 		if(owner.dna.species.id != SPECIES_IPC)
 			owner.adjustCloneLoss(10) //dont abuse it or take cloneloss (organic only)
-		else
-			owner.adjustCloneLoss(0)
 	qdel(src)
 
 /obj/item/organ/regenerative_core/on_life()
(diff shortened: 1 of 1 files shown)


code/modules/mining/equipment/regenerative_core.dm
```diff
@@ -62,7 +62,14 @@
...
 	else
-		owner.revive(full_heal = TRUE, admin_revive = FALSE)
+		owner.adjustBruteLoss(-100) //previously heal proc
+		owner.adjustFireLoss(-100)
+		owner.adjustOxyLoss(-50)
+		owner.adjustToxLoss(-50)
+		if(owner.dna.species.id != SPECIES_IPC)
+			owner.adjustCloneLoss(10) //dont abuse it or take cloneloss (organic only)
+		else
+			owner.adjustCloneLoss(0)
```
MarkSuckerberg:
```suggestion
```

PR #970, opened 2022-04-11
Syndicate and IRMG maid outfits

Ebin-Halcyon:
## About The Pull Request
You've waited for it. And here it is, maid outfits for the Syndicate and the IRMG. (+ a new coat for the IRMG)
image
image
Featuring 'tactical' skirts and combat sleeves with insulation against shocks.

## Why It's Good For The Game

People like maid outfits, people like playing the Syndicate. I like the IRMG, I also like maid outfits. What's not to love.

## Changelog
:cl:
imageadd: Syndicate and IRMG maid sprites
/:cl:

PR #2594, opened 2023-12-17
Adds a variety maid outfit crate

JimKil3:
## About The Pull Request
Adds a 2500 credit crate with a normal maid outfit and the syndicate/inteq tactical maid outfits. Also touches up the normal maid outfit box's description/name to be more grammatically correct.

## Why It's Good For The Game
The tactical maid outfits aren't currently obtainable outside of some ruins or adminbus, despite being added in PR #970 (well over a year and a half ago!). They'll still be uncommon with the fairly steep price tag, and there shouldn't be an issue with purchasing them just for insulated gloves, as the crate costs more than the normal insulated gloves crate while containing the same number of gloves. 

## Changelog

:cl:
add: Adds the variety maid outfit crate!
tweak: Touches up the grammar on the maid outfit box.
/🆑

### DIFF
--- a/code/game/objects/items/storage/boxes.dm
+++ b/code/game/objects/items/storage/boxes.dm
@@ -1361,8 +1361,8 @@
 
 // because i have no idea where the fuck to put this
 /obj/item/storage/box/maid
-	name = "Maid box"
-	desc = "Contains a maid outfit"
+	name = "Maid outfit"
+	desc = "Contains a normal maid outfit."
 
 /obj/item/storage/box/maid/PopulateContents()
 	var/static/items_inside = list(
--- a/code/modules/cargo/packs/costumes_toys.dm
+++ b/code/modules/cargo/packs/costumes_toys.dm
@@ -231,3 +231,12 @@
 					/obj/item/clothing/under/dress/rilena,
 					/obj/item/gun/energy/buster)
 	crate_name = "collectable merchandise crate"
+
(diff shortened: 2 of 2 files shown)


meemofcourse:
i disagree with allowing inteq and syndicate maid uniforms being able to be bought. you should not be able to get them if you're not in the faction.

in a more opinionated comment, their rarity does give them a lot of fucking comedic value when you do eventually find them, be it in ruins, maintloot adminbus, or ships that have their bundles laying around.

thgvr:
Not a fan of this. They're not supposed to be commonly available whatsoever. They're more of a joke item than a uniform piece as well.

JimKil3:
I could probably manage to split out the uniforms into different crates locked by faction, but that wouldn't address the concerns with rarity. If they were only available with a hacked console, in addition to the faction-locked crates, would that work? (Just asking before I put in that work for a PR that's going to be closed.)

goober3:
i don't think maid outfits should be available through cargo at all. rare ruin loot, at most.

thgvr:
> I could probably manage to split out the uniforms into different crates locked by faction, but that wouldn't address the concerns with rarity. If they were only available with a hacked console, in addition to the faction-locked crates, would that work? (Just asking before I put in that work for a PR that's going to be closed.)

I'm happy with keeping them unobtainable normally and allowing mappers to put them as a funny item in ruins.

PR #2598, opened 2023-12-18
(probably) actually fixes outpost baseturfs, removes holodecks

tmtmtl30:
## About The Pull Request

changes the behavior downstream of /datum/map_template's should_place_on_top (the particular proc in question is build_coordinate, and its placeOnTop variable) so that, when false, the baseturfs of the resulting turf are set to the baseturfs of the type being placed, instead of the previous behavior, which preserved the old baseturfs entirely. 

this odd behavior seemed to be specialized for holodecks, and i mistakenly assumed it'd work for outposts given the variable name and an experiment on a testing map. the behavior has been changed to alter the baseturfs according to the type (so that placing a template with should_place_on_top = FALSE results in the same baseturfs list as you'd get from a changeturf(type, initial(type.baseturfs) call), so that outposts should now place with the correct baseturfs. this enables planetary outposts with a special baseturf defined in a ztrait; before, they would be stuck with a space turf hiding in the baseturf list, making their baseturfs basically always broken. 

in order to accommodate this change without breaking holodeck code, holodeck code has been largely removed, with a few exceptions where it'd be a slight pain to remove certain turfs / objects that were originally used in holodecks but which had been since used in other maps. there's also a bit of admin functionality upstream of some holodeck procs that i didn't want to touch

## Why It's Good For The Game

outposts spacing themselves when people pry up tiles is bad, and a fix like this is necessary for planetary outposts to ever have the right baseturfs.

## Changelog

:cl:
del: Holodeck code has been removed.
fix: Outposts should end up erroneously spaced less frequently.
/:cl:


MarkSuckerberg (approved):
seems legit

PR #2603, opened 2023-12-19
"fixes" material tiles

### LINKED ISSUE #2528: Material tiles are broken
Issue reported from  Round ID: 2033 (Shiptest)

Upon further investigation, All material tiles are broken
image

FalloutFalcon:
## About The Pull Request

No longer will you have to see a massive fix me when you accidentally create material tiles instead of the ones you actually wanted!

Fixes #2528
image

## Why It's Good For The Game

I would argue for scraping them entirely but that fix me sprite was REALLY annoying me, We have had it for 2 years based on the commit history

## Changelog

:cl:
fix: Material tile sprites
/:cl:


thgvr:
It's fine if you don't want to, but it would be better to just remove these as they are often confusing when making normal floor tiles and the sprites aren't being maintained.

FalloutFalcon:
> It's fine if you don't want to, but it would be better to just remove these as they are often confusing when making normal floor tiles and the sprites aren't being maintained.

To be fair it is sprite, singular, currently but i do agree that material tiles are gross and only ever created by accident. I will put fully removing them somewhere in my list of things to do but I will do it in a separate pr if i do

thgvr:
That's fine.

PR #2608, opened 2023-12-21
Fixes the decompressed katana bricking itself

### LINKED ISSUE #2501: Bug with ogre blade/sheath
Issue reported from  Round ID: 2017 (Shiptest)

Was messing around with the ogre blade and ran into an issue. Clicked on a void tile while doing the stance thing with the sheath. The blade went into my other hand, but it didn't do the dash. It also kept the sheath in the stance position. Every time I tried to dash with it like this, it said that the sheath was empty. But every time I tried to put the blade back in, it said that the sheath was locked.

SomeguyManperson:
## About The Pull Request

code pulled the blade out of the sheath before checking if it was viably targetted meaning you could remove the blade without clearing the lock on the sheath

## Why It's Good For The Game

fixes #2501

## Changelog

:cl:
fix: the weebstick is no longer held back from its true purpose if you click on a tile you can't see
/:cl:

### DIFF
--- a/code/game/objects/items/melee/misc.dm
+++ b/code/game/objects/items/melee/misc.dm
@@ -763,11 +763,11 @@
 /obj/item/storage/belt/weebstick/afterattack(atom/A, mob/living/user, proximity_flag, params)
 	. = ..()
 	if(primed && length(contents))
+		if(!(A in view(user.client.view, user)))
+			return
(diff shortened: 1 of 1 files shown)


thgvr (approved):
Nice catch

thgvr:
(also it's a reference to Decompressed )

PR #2609, opened 2023-12-21
make world status use config values

ZephyrTFA:


code/game/world.dm
```diff
@@ -310,14 +310,23 @@ GLOBAL_VAR(restart_counter)
...
 
+	var/discord_url
+	var/github_url
+	if(isnull(config))
+		discord_url = "https://shiptest.net/discord"
+		github_url = "https://github.com/shiptest-ss13/Shiptest"
+	else
+		discord_url = CONFIG_GET(string/discordurl)
+		github_url = CONFIG_GET(string/githuburl)
+
```
MarkSuckerberg, line 321: Can't you just add this all to the `if(config)` check above?

MarkSuckerberg: :(

PR #2611, opened 2023-12-24
typingIndicator

GenericDM:
## About The Pull Request
makes the typing indicator properly disappear when exiting the input box, instead of not doing that

## Why It's Good For The Game
typingindicator

## Changelog

:cl:
fix: typing indicator now properly disappears when exiting the input box
/:cl:

### DIFF
--- a/code/modules/mob/mob_say.dm
+++ b/code/modules/mob/mob_say.dm
@@ -1,11 +1,12 @@
 //Speech verbs.
 
 ///Say verb
-/mob/verb/say_verb(message as text)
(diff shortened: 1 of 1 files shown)


thgvr:
oh nice

thgvr:
What's up with this

GenericDM:
> What's up with this

It breaks using it through the verb bar. I've thought about splitting it up into multiple verbs to do this, or some other better solution, but I ran up against the brick wall of "motivation", unfortunately.

PR #2632, opened 2024-01-09
Idle & Active power rework and defines

Thera-Pissed:
## About The Pull Request
Implements new defines set in power.dm for nearly all machines using idle/active_power_usage
Kills auto_use_power() in favor of useStaticPower() procs
Standardizes power draws for most things using said defines
Decreases power used for lighting by half
Generally increases power used by machinery, especially those that are only active for short periods.
SSUs now consume power when decontaminating, ORMs and vending machines now take idle power.
image
If you have any feedback, feel free to share it! Most of these numbers are based on their original values, so definitely could be changed.

## Why It's Good For The Game
This PR standardizes the draw of most machinery, while making it easier to change power use across the board.
Additionally, lighting takes less power, while actual machines have been tweaked to fill this gap.

## Changelog

:cl:
tweak: SSUs draw power to decontaminate
balance: lights no longer take 60% of a ships power, machines use more to account for this.
code: added standardized defines for power usage
/:cl:


code/game/machinery/_machinery.dm
```diff
@@ -180,8 +180,10 @@ Class Procs:
 /obj/machinery/proc/locate_machinery()
 	return
 
-/obj/machinery/process()//If you dont use process or power why are you here
-	return PROCESS_KILL
+/obj/machinery/process()
+	if(!use_power) //machines without process() set would bug and not draw idle, this checks if they draw power
+		return PROCESS_KILL
+	return
```
Kapu1178: this is doing nothing but wasting CPU power. If a machine isn't doing anything in process(), it should be returning PROCESS_KILL.

Thera-Pissed: everything uses static power now, thanks for the advice!

Thera-Pissed:
I forgot that machines can move between areas, whoops

thgvr:
Sorry, your pr got kind of buried over and over and I haven't heard anything of it in discord, I'll check this out this week.

Thera-Pissed:
finally got around to fixing the merge conflict, should be good to go again

code/game/machinery/_machinery.dm
```diff
@@ -653,3 +669,24 @@ Class Procs:
...
+
+/obj/machinery/powertest/CtrlClick(mob/user)
+	if(can_interact(user))
+		on = !on
+		to_chat(usr, "<span class=\"alert\">[src] has changed setting!</span>")
+		if(on)
+			set_active_power()
+		else
+			set_idle_power()
+	return ..()
```
FalloutFalcon: I think this is fine to remove, Should be other ways to test this.

FalloutFalcon (approved):
Everything looks fine but with some minor balances to power consumption def need some test merges

code/game/machinery/_machinery.dm
Thera-Pissed: yep! just had it in to test moving machines between areas and such. Will remove it now

PR #2650, opened 2024-01-20
fixes crusher loot for icemoon ruin dragon not dropping a key

SomeguyManperson:
## About The Pull Request

adds the key to the crusher loot list since the default loot list is overridden by that

## Why It's Good For The Game

Dragon lair could be locked off without admin intervention if the dragon is killed with a crusher

## Changelog

:cl:
fix: you can no longer lock yourself out of the icemoon dragon lair by killing the dragon with a crusher
/:cl:


Zevotech (approved):
Looks alright to me. Wish this ruin would get replaced with something grander and more fitting for a megafauna though.

PR #2663, opened 2024-01-23
Talos Burn Chamber Fix

Luftkommando:
## About The Pull Request

Removes the wall which shares space with a window to the burn chamber.

Current
StrongDMM_k5sWTG8M5g

New
StrongDMM_1caLPXiSRb

## Why It's Good For The Game

Walls and windows share intrinsically linked purposes. Both serve the ship in retaining atmosphere and forming the hull, yet despite this they are often seen as opposites because one is opaque and the other transparent. 

Even with these philosophical similarities they shouldn't share the same tile of space as one another though.

## Changelog

:cl:
fix: Fixed the accidental wall-window combination on the Talos-class burn chamber view window.
/:cl:


Latentish:
1 Tile change lgtm

PR #2680, opened 2024-01-27
Ship faction display take 2

MarkSuckerberg:
## About The Pull Request

Allows specifying a custom faction name in the ship JSON, as well as makes it so the faction name is only inferred from prefix once per round instead of every time the names are moused over.

Also makes ships sort into categories based on faction instead of just being lumped into the shiptest category, so it's a bit easier to find the ship you want to spawn.

## Why It's Good For The Game

Maptainer request

## Changelog

:cl:
admin: Ships should now be categorised by faction in the shuttle manipulator
/:cl:

### DIFF
--- a/_maps/ship_config_schema.json
+++ b/_maps/ship_config_schema.json
@@ -34,6 +34,11 @@
 			"description": "The prefix of the ship class, appended to randomly generated names when they're first purchased.",
 			"maxLength": 5
 		},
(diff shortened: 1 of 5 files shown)


Apogee-dev:
linters :shrivel:

MarkSuckerberg:
we get a little silly

PR #2692, opened 2024-02-03
Increases the Damage of the Illestren Rifle

Cadunkus:
## About The Pull Request

Increases the damage of the Illestren rifle from 30 to 50 per shot.

## Why It's Good For The Game

The Illestren Rifle is very underwhelming. The slow firing speed of a bolt action, small clip size of 5 rounds, unprintable ammunition you have to purchase from cargo, and bulky size would balance out a higher damage value but it does very low damage making it one of the worse guns in the game. Considering how it was adapted from the Mosin Nagant rifle which originally did much more damage this seems to be unintentional.

## Changelog

:cl:
balance: Increased the damage of the 7.62x54mmR bullet (Illestren Rifle Ammunition) from 30 to 50.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/bullets/rifle.dm
+++ b/code/modules/projectiles/projectile/bullets/rifle.dm
@@ -10,7 +10,7 @@
 /obj/projectile/bullet/a762_54
 	name = "7.62x54mmR bullet"
 	speed = 0.3
-	damage = 30
+	damage = 50
 	armour_penetration = 40
 
 // .300 Magnum (Smile Rifle)


Apogee-dev:
Intended. All rifle calibers had nerfed damage in exchange for increased AP. 7.62x54mmR is balanced identically to .308 and 8x58mm Caseless (the latter of which, by the way, is also used in a bolt-action rifle, albeit one with a scope). Their intended role is defeating higher-quality armor. If you want to buff it (and, yeah, that's fair, they ARE very slow compared to the DMRs because of the bolt action operation), both 7.62x54mmR and 8x58mm could get 35 damage, but any more than that would be excessive given their AP value.

Damage, by the way, is based more on projectile weight and class. The other rifle cartridges (.308, 8x57mm) deal 30 damage as well. So does the .357 revolver. 40 damage is what shotgun slugs and the .300 WIN-MAG used by SRM's actual sniper rifle deals (and is what potential future magnum rifle cartridges like .338 Lapua might deal), and the single most powerful normal ballistic cartridge, .45-70, deals 45 damage. The only bullets that deal more damage than that are things like gauss slugs, gyrojets, and 40mm grenades.

Ballistics, as a rule, should take 3-5 shots to kill a human target. The only exceptions to this are buckshot, hollow point cartridges for more powerful casings (both of which are kept in check by their bad AP), and excessively powerful cartridges like .50 BMG. The Illestren is none of these.

Cadunkus:
> Ballistics, as a rule, should take 3-5 shots to kill a human target.

Dang, I thought it wasn't intended for PvP and was for hunting fauna. Okay gonna close this soon.

Apogee-dev:
> > Ballistics, as a rule, should take 3-5 shots to kill a human target.
> 
> Dang, I thought it wasn't intended for PvP and was for hunting fauna. Okay gonna close this soon.

Yeah nothing's intended _primarily_ for hunting fauna just because pvp balance also needs to be taken into account, even if described as a "hunting" weapon. We removed .38 hunting ammo, after all- if your main target is fauna (not counting Goliaths), use weapons that deal high damage but have low AP such as shotguns, magnums, or SMGs, especially with hollowpoint. We probably should add hollowpoint variants for the rifles for this purpose.

Cadunkus:
> Yeah nothing's intended _primarily_ for hunting fauna just because pvp balance also needs to be taken into account, even if described as a "hunting" weapon.

Gotcha. What about adding a damage bonus to non-carbons (excluding Borgs) on the Illestren then so it's viable for sniping hazardous fauna like ice whelps but doesn't dominate in the rare ship battle?

thgvr:
Snowflake damage modifiers are bad, something we had before but is no longer necessary.

Cadunkus:
Gonna see what damage will make it the most useful for hunting without passing the 3-shot threshold for carbons.

PR #2702, opened 2024-02-06
Removes firing pins and clumsy checks on guns

FalloutFalcon:
## About The Pull Request

Removes firing pins and clumsy checks from guns

## Why It's Good For The Game

Firing pins and clumsy checks are huge cruft.
Both are designed to limit around things we don't care about anymore

## Changelog

:cl:
del: Removed firing pins
del: Removed clumsy checks on guns
/:cl:


Cadunkus:
Does this mean you can't have guns without firing pins (that are therefore useless)?

FalloutFalcon:
> Does this mean you can't have guns without firing pins (that are therefore useless)?

No its a in depth removal of every pin and and check for it. This means every gun will fire regardless. Also simplifies paths for many weapons which were just gun/normalfiringpin basically 

currently it crashes the game. Haven't had time to look into it

code/game/machinery/computer/arcade.dm
```diff
@@ -88,12 +88,6 @@ GLOBAL_LIST_INIT(arcade_prize_pool, list(
 		prizes *= 2
 	for(var/i = 0, i < prizes, i++)
 		SEND_SIGNAL(user, COMSIG_ADD_MOOD_EVENT, "arcade", /datum/mood_event/arcade)
-		if(prob(0.0001)) //1 in a million
```
thgvr: Keep this

thgvr (changes requested):
Get the sprites as well please

thgvr (approved):

PR #2703, opened 2024-02-06
Adds the Cobra-20 as a buyable gun in the cargo market. (and fixes the stechkin's description)

Martinpachu:
## About The Pull Request

Title says all, i think.
Secretly also fixes the description on the stechkin to say that it's from scaraborough arms and not ''gorlex marauders''
## Why It's Good For The Game
Adds an option for buying an SMG in the market, since there's no SMGs currently available to buy, which felt like a weird skip since there's rifles, PDWs, and pistols. meant to be a sort of inbetween for from pistols and pdws/rifles
## Changelog

:cl:
add: Adds the Cobra-20 as buyable from the outpost alongside it's magazines.
spellcheck: fixes the stechkin being from the gorlex marauders and returns it to it's rightful home of scaraborough arms
/:cl:

### DIFF
--- a/code/modules/cargo/packs/gun.dm
+++ b/code/modules/cargo/packs/gun.dm
@@ -26,7 +26,7 @@
 
 /datum/supply_pack/gun/makarovs
 	name = "Stechkin pistol crate"
-	desc = "Contains two concealable stechkin pistols, produced by the Gorlex Marauders and chambered in 10mm."
+	desc = "Contains two concealable stechkin pistols, produced by Scaraborough Arms and chambered in 10mm."
 	cost = 2000
 	contains = list(/obj/item/gun/ballistic/automatic/pistol,
 					/obj/item/gun/ballistic/automatic/pistol)
--- a/code/modules/cargo/packs/gun.dm
+++ b/code/modules/cargo/packs/gun.dm
@@ -119,9 +119,8 @@
 	name = "Cobra-20 SMG Crate"
 	desc = "Contains two .45 submachine guns, manufactured by Scaraborough Arms and chambered in .45"
 	cost = 2500
-	contains = list(/obj/item/gun/ballistic/smg/c20r/cobra,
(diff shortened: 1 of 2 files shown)


code/modules/cargo/packs/gun.dm
```diff
@@ -26,7 +26,7 @@
 
 /datum/supply_pack/gun/makarovs
 	name = "Stechkin pistol crate"
-	desc = "Contains two concealable stechkin pistols, produced by the Gorlex Marauders and chambered in 10mm."
+	desc = "Contains two concealable stechkin pistols, produced by Scaraborough Arms and chambered in 10mm."
```
Apogee-dev: Typo, should be spelled "Scarborough Arms"

code/modules/cargo/packs/gun.dm
```diff
@@ -115,6 +115,14 @@
 					/obj/item/gun/ballistic/shotgun/flamingarrow)
 	crate_name = "rifle crate"
 
+/datum/supply_pack/gun/cobra20
+	name = "Cobra-20 SMG Crate"
+	desc = "Contains two .45 submachine guns, manufactured by Scaraborough Arms and chambered in .45"
+	cost = 2500
```
Apogee-dev: I'd bump this up to 3,000 credits. Having only a 500 credit gap between this and a crate of pistols does mean there's very little reason to ever buy the pistols when you're so close to SMGs, and it leaves enough of a gap between the SMG crate and Illestrens or shotguns to fill a useful budget gap. 2500's also exactly the same as the revolver crates, and it should probably cost a little more than those.

Apogee-dev (changes requested):

Cadunkus:
There is actually the WT550 smg but having another smg option would be nice especially for ships without R&D and the means to get 20k research points.

Apogee-dev (approved):

PR #2715, opened 2024-02-09
Adds HIDEHORNS bitflag (wigzards can live)

thgvr:
## About The Pull Request
Adds new HIDEHORNS bitflag for things that need to hide horns and shifts mask+head handling to use it instead of HIDEHAIR.
This means wigs now 

## Changelog

:cl:
fix: Wigs no longer hide Sarathi horns
/:cl:


tmtmtl30:
seems like it might be best to go through all clothes that currently have HIDEHAIR and determine if adding HIDEHORNS is appropriate, though that'd be a fair bit of work.

thgvr:
i figure it's better to do it on anything that's noticed

PR #2720, opened 2024-02-11
Multiple runtime error fixes

MarkSuckerberg:
## About The Pull Request
Basically just goes down the list for most frequent runtimes and tries to fix them one way or another. 

## Why It's Good For The Game
Runtime errors shouldn't happen

## Changelog

:cl:
fix: Headpikes actually work again
fix: Reinforced floors now don't spawn metal when decompression is experienced
/:cl:


MarkSuckerberg:
the horror

PR #2721, opened 2024-02-11
Gives the Riggs a touch-up

Apogee-dev:
## About The Pull Request
rigger_remap
The Riggs has aged incredibly well for being one of the oldest maps still in rotation, but it's definitely fallen out of line with newer ships as standards have been updated. This PR brings it in line with those standards and freshens up the layout and decoration a bit.

- Removed RnD and ORM
- Added separate cryo room
- Added crew lounge area and miner ready room
- Moved atmos slightly to make accessing the empty rooms easier and replaced the standing air tanks with a 1-tile open air tank (temporary, really, until we get better tanks coded)
- Added access restrictions to atmos and engine room (but not engineering bay)
- Reduced miner slots to 1 and assistant slots to 2
- Removed Mech Technician

## Why It's Good For The Game

Riggs was getting crusty and needed some love.

## Changelog

:cl:
tweak: Made the Riggs-class sloop nicer to live in
balance: Removed RnD and ORM from the Riggs
/:cl:


west3436:
Why the slot reductions?

meemofcourse:
the riggs does not need 13 crewmembers. that's 3 less than the twink, with half the space

meemofcourse (changes requested):
imagen
consider breaking up the monotiling here
imagen
since you're removing RnD from the ship, i don't think it needs wayfinding beacons anymore
imagen
please unleash the cat from the closet
imagen
i think we agreed on not giving fax machines to independent ships, so please remove this fax machine

Apogee-dev:
> please unleash the cat from the closet

Jonesy is in the closet because there's no way for me to put her in cryo :pensive:

anyway I followed the other responses. Kept the nav beacons for simplebots though. Not too much reason _not_ to have them since simplebots, particularly medbots, are so common and easy to get (just smash a malfunctioning one on a junkplanet) and the role of nav beacons is largely QoL

meemofcourse (approved):
sealed forever

Latentish:
TM Ready I think

XxNinMario64xX:
Mech Technician (or whatever it's called) feels like an unnecessary job now due to having no ability to construct more mechs. Just call them an engineer and don't imply they can create more mechs, as people will join the riggs thinking they can build more than a Ripley during the round unless there's a way to give the ability to create more mechs without re-adding R&D.

Apogee-dev:
> Mech Technician (or whatever it's called) feels like an unnecessary job now due to having no ability to construct more mechs. Just call them an engineer and don't imply they can create more mechs, as people will join the riggs thinking they can build more than a Ripley during the round unless there's a way to give the ability to create more mechs without re-adding R&D.

The mech technician was removed along with the reductions in other job slots.

Latentish (approved):
Big improvement. Tested on live, found no issues.

Latentish:
marg

PR #2722, opened 2024-02-11
Dissection tweaks: I'm tired of the corpse pile

SomeguyManperson:
## About The Pull Request

Problem: people are dissecting well-documented life like common animals and, distressingly, _human  corpses._ Not only is this behavior odd, cadaver disposal methods can end up being quite concerning. Or quite concerningly non-instanced.

Solution: Dissections no longer provide points as a default for animal/humanoid corpses, massively shortening the list of dissectable things. Pod and jellypeople have also been removed from the modified dissections table as people since they are both playable (pods less but still) and slimes can show up as legion corpses.
In their place, crystal mobs, ancient goliaths, and elites/megafauna can be dissected, since they are both unique and rare (relatively)

I also switched the dissection variant scaling to be fractional of the expected maximum at advanced surgery/experimental tier dissection, since that's what is going to be used throughout the majority of a round and having the commonly used numbers more directly present makes for easier balance checks.

New dissectable things at experimental dissection research outputs, halve and quarter for thorough and base outputs respectively:
Megafauna: 60k, on par with xeno bodies. Since you can't drag these, a clinic would have to be prepared on-site for the ones that leave a body (dragon) to extract the full value, and they are both rare and more than capable of killing a crew
Elites: 12k, uncommon spawn on some planets, can be lethal
crystal goliath: 10k, likely less lethal than the crystal legion but still dangerous
mega arachnid, 8k, Appears in one ruin as a miniboss, rare enough it'd be something to peek at
forgotten watcher: 6k, rare but relatively harmless compared to the other crystal mobs (unless I'm missing something)
watcher variants: 4k, uncommon subspecies of watchers
ancient goliath: 4k, it's a remarkably old goliath, probably some value there

## Why It's Good For The Game

Five. Hundred. Human bodies. In my medbay.

## Changelog

:cl:
add: ancient goliaths, crystal fauna (aside from legion), elites and megafauna now have boosted dissection outputs
del: human corpses, podpeople, jellypeople and animal corpses can no longer be dissected. What did you expect to find, an extra pair of lungs?
/:cl:


Sun-Soaked:
we are so back

thgvr:
Remind me to review this this weekend.

code/modules/surgery/experimental_dissection.dm
```diff
@@ -45,31 +45,34 @@
...
-			else if(isjellyperson(H) || ispodperson(H) || isalien(H))
-				cost = (BASE_HUMAN_REWARD*14)
-			else if(isskeleton(H))
-				cost = (BASE_HUMAN_REWARD * 0.5)
-	else
-		cost = (BASE_HUMAN_REWARD * 0.5)
-
-
+				cost = (MAX_DISSECTION_REWARD*20)
+	else for(var/type in mob_mult_list) // THIS. ELSE IF(). ENDS. HERE.
```
west3436, line 75: I think you can do this without iterating:
`if(target in mob_multi_list)
     cost = [...]`

code/modules/surgery/experimental_dissection.dm
```diff
@@ -45,31 +45,34 @@
 	user.visible_message("<span class='notice'>[user] starts dissecting [target].</span>", "<span class='notice'>You start dissecting [target].</span>")
 
 /datum/surgery_step/dissection/proc/check_value(mob/living/target, datum/surgery/advanced/experimental_dissection/ED)
-	var/cost = BASE_HUMAN_REWARD
+	var/cost = 0
 	var/multi_surgery_adjust = 0
 
 	//determine bonus applied
+	var/list/mob_mult_list = list(/mob/living/simple_animal/hostile/asteroid/elite = 6,
```
MarkSuckerberg:
```suggestion
	var/static/list/mob_mult_list = list(/mob/living/simple_animal/hostile/asteroid/elite = 6,
``` 
just so it doesn't need to get re-created every time, it won't matter since it's just a surgerybut it can't hurt

code/modules/surgery/experimental_dissection.dm
MarkSuckerberg: you wouldn't get subtypes, I'm pretty sure, but if that's not an issue it's possible yeh

MarkSuckerberg (changes requested):
absolutely wonderful PR

code/modules/surgery/experimental_dissection.dm
SomeguyManperson: Directly comparing types won't get subtypes, though the only thing that's using that right now is elites.

MarkSuckerberg (approved):
the honourable ship's doctor breathes a deep sigh of relief as a weight is lifted from their shoulders for good

PR #2723, opened 2024-02-11
Doubles time limits for all missions (except fishing ones)

SomeguyManperson:
## About The Pull Request

Current missions aren't built to be rushed, they're built surrounding planetary/space exploration and short time limits detract from that by forcing everyone to rush or lose the mission. 30 minutes plus or minus a bit isn't enough to reliably explore/interact/etc and get back to the station.

Also makes urgent delivery missions reasonably possible without pre-collecting the stuff they need.

## Why It's Good For The Game

Let me explore damnit

## Changelog

:cl:
tweak: missions outside of the fishing missions have had their time limits doubled
/:cl:

### DIFF
--- a/code/modules/overmap/missions/acquire_mission.dm
+++ b/code/modules/overmap/missions/acquire_mission.dm
@@ -72,7 +72,7 @@
 	name = "Diamond needed (urgent!!)"
 	weight = 3
 	value = 700
-	duration = 20 MINUTES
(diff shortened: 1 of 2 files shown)


Zevotech:
Thank you! I was planning something similar to this myself but this should solve the more pressing issues with the time limits I wanted to solve.

thgvr (approved):
Sure

PR #2725, opened 2024-02-12
Removes genetics from the techweb

SomeguyManperson:
## About The Pull Request

Gone.

## Why It's Good For The Game

Odd passive buff system that works off of magic

## Changelog

:cl:
del: genetics is no longer researchable, designs related to genetics no longer exist
/:cl:


Cadunkus:
Better idea: Genetics can be researched, printed, set up, and everything but instead of giving you useless crap like pressure adaption and telekinesis it gives you cat ears and a tail when you step inside it.

Chubbygummibear:
i'm going to cry

PR #2750, opened 2024-02-20
Medical Machinery expansion

thgvr:
## About The Pull Request
Ports https://github.com/tgstation/tgstation/pull/50432

## Why It's Good For The Game
WIP

## Changelog

:cl:
add: Added new things
add: Added more things
del: Removed old things
tweak: tweaked a few things
balance: rebalanced something
fix: fixed a few things
soundadd: added a new sound thingy
sounddel: removed an old sound thingy
imageadd: added some icons and images
imagedel: deleted some icons and images
spellcheck: fixed a few typos
code: changed some code
refactor: refactored some code
config: changed some config setting
admin: messed with admin stuff
server: something server ops should know
/:cl:


code/game/machinery/medical/_medical.dm
```diff
@@ -0,0 +1,86 @@
...
+
+/obj/machinery/medical/process()
+	update_overlays()
+	update_appearance()
+
+	if(!attached)
+		use_power = IDLE_POWER_USE
+		return
+
+	if(machine_stat && (NOPOWER|BROKEN))
```
west3436: `if(machine_stat & (NOPOWER|BROKEN))`

code/game/machinery/medical/_medical.dm
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Properly gets rid of status effects from the attached
+ *
+ * Internal function, you shouldn't be calling this from anywhere else. Gets rid of all the status effects, traits and other shit you might have
+ * put on the attached victim. Automatically updates overlays in case you have some, and changes power to idle power use.
+ */
+/obj/machinery/medical/proc/clear_status()
+	update_overlays()
+	qdel(attached_beam)
+	use_power = IDLE_POWER_USE
```
west3436: /obj/machinery/medical/process() already consumes IDLE_POWER_USE at the start of the proc so this is double-counting. Not sure if intended

RikuTheKiller:
TL;DR simple stasis is practically useless on this codebase, this has to do something more than that to be interesting. There's also space concerns and the resources used are too boring imo considering the potential the whole thing has.

Overall though, it can be done later down the line as the PR has barely any immediate impact due to stasis beds being scarce anyway.

Read below at your own risk, it's rather long.

Power consumption as a balancing measure is, frankly, extremely boring and there are ships that outright counter it, such as anything with a TEG. In general, power is *not* an issue even with excessive consumption unless power use gets doubled or tripled or something, which I'm guessing this won't do on it's own by a longshot, else it'd be absurd.

Instead? Custom resources for the machines. Filters for dialysis, refillable oxygen tanks for ventilators and such. Just general things you gotta do to maintain the machines apart from powering them, after all, they aren't magic. It's also more interesting gameplay wise, I suppose. Plus, we can make extra supplies for life support available via cargo.

Also, dialysis and liver stabilization should be combined. Another thing, life support seems particularly useless in some ways as it doesn't actually *revive* the target nor does it allow you to revive them below the regular threshold, which would be a cool feature considering it lowers the death threshold. Also, if the 1000 damage threshold stays unchanged, well, we've got a problem as 600 damage is the *maximum damage physically possible* and keeping someone whose body looks like ground pulp alive with some basic machinery is, frankly, impossible.

Also from what I can gather, life support only works on, well, people that are alive. The main function of stasis is to stop decay. It would be neat if it allowed you to revive people below the regular threshold such that you can stop them from decaying much earlier. (more interesting than copying formaldehyde anyway) It'd also allow you to apply live-only treatment earlier, allowing for faster revival. That seems like a good incentive to actually use these considering it's several clunky machines.

Yet another thing, the machines (or at least some of them) could be combined into a sort of "modular system" where you can slot in machines you have available. This wouldn't be a thing we gotta do *if* operating on people didn't require, you know, space. But it does. We can't just be out here putting 4 dense machines next to the patient. A lot of ships just don't have space for that and having to haul them back and forth is awfully inconvenient and further discourages even using them in the first place. Simply being able to hook someone up to a system is all we really need, to be honest, and having that system be a combination of independently maintained machines is, to be honest, extremely cool imo.

Overall though, the PR itself is fine. We don't have any urgent uses for stasis beds anyway. This has plenty of time to get implemented later down the line as it's immediate impact on the game is, well, practically zero. We already have formaldehyde and when a patient is dead and has formaldehyde, it's almost perfectly equivalent to regular stasis. No processing, no decay, nothing. And you don't get 16 injured patients in a row here, running out of epipens entirely is impressive on it's own and we have plenty of formaldehyde bottles that can be further diluted to give 20 injections with a water bottle.

rye-rice:
sowwy, nwo prwogwess awnd nwo tm iwn mwonths, rweopwen  wwen  wrorking on iwt agwian

PR #2756, opened 2024-02-22
Rebalances prices of guns.

Martinpachu:
## About The Pull Request

Rebalances the prices of guns aimed towards mostly making automatics such as the SKM, P16 and Cobra more expensive, and slightly lowers the price of shotguns.
Waiting until the SKM merge so that i can change the price of that without repercussions.

## Why It's Good For The Game

Automatics are very powerful and should have a price that matches so. Might also make non-automatic options such as the Illestren, Shotguns and Pistols more worthwhile.

## Changelog

:cl:

balance: Cargo gun crates are more granularized, having only one gun at a time and a price reflecting that. Prices were rebalanced making automatics more expensive overall.

/:cl:


thgvr:
Why keep the crates with two guns? Should be more granular

PR #2787, opened 2024-02-29
Removes the singularity, tesla and particle accelerator packs from cargo

SomeguyManperson:
## About The Pull Request

The singularity, tesla, and particle accelerator can no longer be bought since the engines are either unbalanced or impossible to effectively use on most ships

## Why It's Good For The Game

The singularity can't be effectively fitted to any ship without building in hyperspace. Even if it is, it's going to irradiate/EMP everything in a large area, which likely means somewhere important

The tesla is just free infinite power if set up, which is boring. It doesn't have any upkeep and containment is extremely simple

The particle accelerator doesn't do anything except turn on and power those two engines

## Changelog

:cl:
del: The tesla, singularity, and particle generator are no longer purchasable from cargo
/:cl:

### DIFF
--- a/code/modules/cargo/packs/machinery.dm
+++ b/code/modules/cargo/packs/machinery.dm
@@ -298,31 +298,9 @@
 	crate_name = "grounding rod crate"
 	crate_type = /obj/structure/closet/crate/engineering/electrical
 
-/datum/supply_pack/machinery/PA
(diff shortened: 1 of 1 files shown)


MarkSuckerberg (approved):
good riddance

PR #2788, opened 2024-02-29
Passive radiation structures/effects don't trigger if there are no players nearby

SomeguyManperson:
## About The Pull Request

Assuming some radiation comes from waste planets generating a bunch of radiation waves, and that there are a lot of radiation generating things players might not even be affected by, radiation/contamination can be reduced by preventing these from activating if they wouldn't affect a player

Also consolidates the random chance for a pulse to trigger into its timer for structures, since there's not much purpose to it

## Why It's Good For The Game

don't need to waste time with replicating radiation somewhere it's not going to end up doing anything

## Changelog

:cl:
tweak: radiation structures/effects only irradiate their surroundings if near a player
/:cl:

### DIFF
--- a/code/game/objects/effects/radiation.dm
+++ b/code/game/objects/effects/radiation.dm
@@ -19,7 +19,7 @@
 		return ..()
 
 	var/player_in_range = FALSE
-	for(var/mob/living/L in range(rad_spread * 2))
+	for(var/mob/living/L in range(rad_spread))
 		if(L.client)
 			player_in_range = TRUE
 			break
--- a/code/game/objects/effects/radiation.dm
+++ b/code/game/objects/effects/radiation.dm
@@ -25,7 +25,7 @@
 			break
 	if(!player_in_range)
 		return ..()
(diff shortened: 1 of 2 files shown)


MarkSuckerberg:
Oh hell yeah. Suggestion: (as in, if it ends up being too complex, don't worry about it) I'd use a connect_range component with a range like what you have now registered to start pulses on enter and stop on exit. These components aren't really used for much now but I think they could apply perfectly here and eliminate the need for `process()`ing.

Cadunkus:
Does this nerf radiation collectors so they don't generate power if nobody's near the engine room? Could maybe add radiation collectors as a proc for nearby radiation sources next to players.

SomeguyManperson:
> Does this nerf radiation collectors so they don't generate power if nobody's near the engine room? Could maybe add radiation collectors as a proc for nearby radiation sources next to players.

This doesn't effect all radiation producers, just mapped ones (barrels/background radiation), so a supermatter is unaffected

PR #2791, opened 2024-03-01
Tweaks plasmasoul anomalies

### LINKED ISSUE #2763: [BUG] Anomalies deal toxin damage to otherwise toxin immune races like Plasmamen and IPCs.
## Reproduction:

The plasmasoul anomaly deals toxin damage to races that should be immune to toxin damage like IPCs and Plasmamen which is especially bad since as far as I know you can't heal said toxin damage when that happens.

(also there appears to be another bug where when you ahelp it to hopefully get rid of the permanent 90 damage on your character you get the response "github" and then the ticket immediately closes with no further explanation which is very unhelpful.)

JimKil3:
## About The Pull Request
Resolves #2763 
Plasmasoul anomalies now check if a mob can process organic chemicals. If they cannot, as is the case with IPCs, instead of adding plasma to your system they burn you. Also cleans up the code a little bit - nothing else player-facing, though.

## Why It's Good For The Game
Being forced to make a niche chemical (system cleaner) just to remove otherwise-permanent damage from even going near one of these is not fun, and being revived if you _die_ from it is even more difficult. Plasmasoul anomalies are still quite dangerous to IPCs, just in a more manageable (and fun) way.

## Changelog

:cl:
tweak: Plasmasoul anomalies now burn IPCs instead of applying a toxin
/🆑

### DIFF
--- a/code/game/objects/effects/anomalies/anomalies_plasmasoul.dm
+++ b/code/game/objects/effects/anomalies/anomalies_plasmasoul.dm
@@ -15,34 +15,48 @@
 		return
 
 	COOLDOWN_START(src, pulse_cooldown, pulse_delay)
-	for(var/mob/living/mob in range(effectrange,src))
-		if(iscarbon(mob))
-			var/mob/living/carbon/target = mob
-			target.reagents?.add_reagent(/datum/reagent/toxin/plasma, reagent_amount)
-			to_chat(mob, span_warning("Your blood feels thick.."))
-			playsound(mob, 'sound/effects/bubbles.ogg', 50)
+	harm_surrounding_mobs()
 
 /obj/effect/anomaly/plasmasoul/Bumped(atom/movable/AM)
 	var/turf/open/spot = locate(rand(src.x-effectrange, src.x+effectrange), rand(src.y-effectrange, src.y+effectrange), src.z)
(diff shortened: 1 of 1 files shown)


Cadunkus:
What about plasmamen? They can process chemicals but are immune to toxin damage so the plasmasouls screw them up too.

SomeguyManperson:
> What about plasmamen? They can process chemicals but are immune to toxin damage so the plasmasouls screw them up too.

They aren't immune to toxins

RikuTheKiller:
> > What about plasmamen? They can process chemicals but are immune to toxin damage so the plasmasouls screw them up too.
> 
> They aren't immune to toxins

Codewise, at least on /tg/station, due to a bunch of biotype BS they are.
Not sure if that applies here but it may be where they got the idea from.

JimKil3:
> What about plasmamen? They can process chemicals but are immune to toxin damage so the plasmasouls screw them up too.

if they were immune to toxins damage plasmasouls wouldn't give them any because how they work is adding the plasma reagent to the mob's system. if they're taking toxins damage from that then they're not immune to toxins. Also if they can process reagents they can just drink charcoal or dylovene to fix it, same as any other species would do

PR #2798, opened 2024-03-03
Craftable Towels

Skies-Of-Blue:
## About The Pull Request
image

## Why It's Good For The Game
Crew shouldn't have to buy a sauna kit just for some towels. 
Players are now enabled to have a spontaneous beach episode, any time, anywhere!

## Changelog

:cl:
tweak: towels can now be crafted using two cloth
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/sheets/sheet_types.dm
+++ b/code/game/objects/items/stacks/sheets/sheet_types.dm
@@ -348,6 +348,7 @@ GLOBAL_LIST_INIT(cloth_recipes, list ( \
 	null, \
 	new/datum/stack_recipe("improvised gauze", /obj/item/stack/medical/gauze/improvised, 1, 2, 6), \
 	new/datum/stack_recipe("rag", /obj/item/reagent_containers/glass/rag, 1), \
+	new/datum/stack_recipe("towel", /obj/item/towel, 2), \
 	new/datum/stack_recipe("bedsheet", /obj/item/bedsheet, 3), \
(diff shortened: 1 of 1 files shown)


Skies-Of-Blue:
Unmarked as draft. The longer I stare at towels the more I want to change, but that's out of scope for this PR. I may get around to larger changes at a later date.

MarkSuckerberg (approved):
some people say it might be better off being a small crate from cargo, but I personally think this is fine-- unless you'd rather add both or switch to that

Skies-Of-Blue:
> some people say it might be better off being a small crate from cargo, but I personally think this is fine-- unless you'd rather add both or switch to that

As silly as it is to have these craftable, I don't think it's any sillier than chopping up a jumpsuit and making a duffelbag out of it. The issue is more so our crafting system, which is out of scope of this PR. Spacemen shouldn't have to settle for blankets, and shouldn't have to pay a towel tax at the start of every round imo. 

The only current way to obtain towels in-game is to buy a sauna crate, which is iffy. Ultimately it would make more sense if ships came stocked with them, but that's once again out of scope. For the time being, this band-aid of a solution slots neatly into our existing systems. If requested I can make a crate stocked exclusively with towels in addition to this change, though.

MarkSuckerberg:
Oh yeah I totally agree. Just figured I'd sound it out with you as well to see if you were more inclined towards the others' ideas. Towels are dead simple and probably one of the least bad offenders-to-be in the crafting system, since they're pretty much just large bolts of cloth anyway.

PR #2811, opened 2024-03-06
Surgery System/Process Rework (Omnisurgery)

XxNinMario64xX:
## About The Pull Request

**Big thanks to Len Kagamine for creating the backend for this system.**

Reworks the current surgery system by combing all surgeries into one large "omnisurgery" that allows players to move freely between 'layers' of the body, allowing for multiple surgeries to be done in a single instance.

## Why It's Good For The Game

Currently, surgeries require tedious amounts of movement between the popup "choose a surgery" UI and forces players to repeatedly do the same steps for each surgery despite them being virtually similar. (EX: Needing to tend wounds in the chest, then needing to cauterize and cut back INTO the chest to repair the bone in it.)

By making the surgery more akin to bay surgery and utilizing radial menu options, surgeries can be done in one instance rather through several instances requiring the same steps every time. It makes no sense how you have to completely close up the body of your patient so you can **open them back up** to fix their chest, less so if you have to tend wounds and then **repeatedly slice through said tended wounds** to fix their body.

This also reduces the amount of actions required between surgery steps, making it effectively the same surgery system with popups and inputs thanks to radial menus filling in for surgeries that require the same tool on the same layer. However, we are trying to avoid this redundancy in the first place, so it'd be best to not see the radial menu as an opportunity to add 20 scalpel surgeries to the same layer.

**This is still in development.**

## Changelog

:cl:
add: Added "Omnisurgery"
tweak: Makes surgeries that use the same layer and tool accessible via radial menu
balance: TBD
code: Added Omnisurgery framework to allow new surgeries to be added to the new system (Len Kagamine)
refactor: Old surgery system now uses a single "Omnisurgery" system; this still uses old surgery code, but restructures it into using said Omnisurgery rather than a bunch of regular surgeries.
/:cl:


XxNinMario64xX:
https://github.com/shiptest-ss13/Shiptest/assets/22011788/49b20d4b-2c3a-4d2b-912c-8405bd1a3ce0

XxNinMario64xX:
New draft PR since I'm the new owner of the branch: [here](https://github.com/shiptest-ss13/Shiptest/pull/2826)

PR #2823, opened 2024-03-09
Xeno fix for event

FalloutFalcon:
## About The Pull Request

A ton of small tweaks, fixes, and balances to xenos for the event

## Why It's Good For The Game

Bug fixes and makes xenos a little less powerful

## Changelog

:cl:
balance: some xeno stuff needed nerfs
fix: fixed a bunch of xeno issues
/:cl:


code/modules/mob/living/carbon/alien/humanoid/alien_powers.dm
```diff
@@ -291,7 +291,7 @@ Doesn't work on other aliens/AI.*/
 
 /obj/effect/proc_holder/alien/sneak/fire(mob/living/carbon/alien/humanoid/user)
 	if(!active)
-		user.alpha = 25
+		user.alpha = 75
```
thgvr:
```suggestion
		user.alpha = 150
```

code/modules/mob/living/carbon/alien/alien.dm
```diff
@@ -180,3 +180,7 @@
 /mob/living/carbon/alien/on_standing_up()
 	. = ..()
 	update_icons()
+
+/mob/living/carbon/alien/examine(mob/user)
+	. = ..()
+	. += "It's a strange creature."
```
thgvr, line 186:
```suggestion
	. += "It's a strange creature."
	
```

Zevotech:
Saw some complaints that the tail spin was way too nasty/stunned way too hard during the event. The flat disarm on xeno shoves/attacks also seemed like an oversight. Stuns could use some work, the neurotoxin spit was still as lethal as ever. Facehuggers probably shouldn't target IPCs at all or at least not stun them.
The brute damage from being chestbursted should probably be increased to 175 or even 200. Many times it left people alive which could potentially result in someone being chestbursted more than once. Many people weren't properly respecting the need for face protection which was both a mixture of stupidity on their part and lack of true threat from just 100 damage on the mechanics part.

Overall though, better than they were before. Good work.

Burning02:
Asked to comment this here by the PR owner, so uh here it is;

During the colony event mainly when I was observing, people with hardsuit helmets on (No mask I believe since why would you need a mask with a hardsuit fully on save for atmospheric uses) were getting "infected" by facehuggers, althought all they did was latch on and do the normal stun, minus the embryo implantation I believe (Didn't get the race of the people I saw it happen to) which resulted in at least one or two people I saw getting ganked by xenos before they could fire a shot back, albeit it didn't last for long.

PR #2824, opened 2024-03-09
Players can no longer open additional job slots 

ZephyrTFA:
## Changelog

:cl:
balance: Only admins can open job slots
fix: Cryoing correctly opens a job slot
code: Refactored ShipOwner into tsx
/:cl:

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -303,15 +303,25 @@ MAPPING_DIRECTIONAL_HELPERS(/obj/machinery/computer/cryopod/retro, 17)
 /obj/machinery/cryopod/proc/despawn_occupant()
 	var/mob/living/mob_occupant = occupant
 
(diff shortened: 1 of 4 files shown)


Apogee-dev:
depending on how much you're willing to add to this, it'd be nice if we could still let players reopen job slots they've previously closed or add job slots if a player is gibbed / DNR, up to the ship's designed maximum. i guess by checking the manifest against the status of those player mobs (alive? doesn't exist? DNR?)

ZephyrTFA:
> depending on how much you're willing to add to this, it'd be nice if we could still let players reopen job slots they've previously closed or add job slots if a player is gibbed / DNR, up to the ship's designed maximum. i guess by checking the manifest against the status of those player mobs (alive? doesn't exist? DNR?)

This is now performed

thgvr:
I trust your code is good honestly typescript is above me, rest looks fine though

PR #2848, opened 2024-03-21
Mob Spawner Rework

FalloutFalcon:
## About The Pull Request

reworks most mob spawners to fall under a much simpler spawner/burrow
each planet has one simple nest instead of the old system with difficultly levels i never even noticed were there despite the fact i used to basically just kill nests and powergame when i first played
this has the byproduct of killing off all the awful magic item spawns

attacking a bunch of other cruft, mostly lavaland

## Changelog

:cl:
add: A new generic nest used across all planets
del: Old portal and tendrils and other cruft
refactor: refactorted mob spawners to have one for each planet
/:cl:


Sun-Soaked:
poke poke (fnish the atomization of necro disease & crusher stuff)

code/modules/mob_spawner/spawner.dm
```diff
@@ -0,0 +1,40 @@
...
+
+	move_resist = INFINITY
+	anchored = TRUE
+	density = TRUE
+
+	var/max_mobs = 5
+	var/spawn_time = 300 //30 seconds default
+	var/mob_types = list(/mob/living/simple_animal/hostile/carp)
+	var/spawn_text = "emerges from"
+	var/faction = list("hostile")
```
Mirag1993, line 15: why hostile?

FalloutFalcon: This was not even my change. Most of the subtypes that matter use `"mining"`

Mirag1993: > This was not even my change. Most of the subtypes that matter use `"mining"`

That's the problem: the mobs go full Highlander mode — killing each other until there can be only one.

PR #2868, opened 2024-03-31
Craftable Empty Ammo Boxes

Gristlebee:
## About The Pull Request

Adds empty variants for every ammo box currently in-game, which can be crafted from blank ammo boxes.

Blank ammo boxes can either be made in the crafting menu with one cardboard and iron, or printed in the autolathe.

Also touches on ammo box initialization so ammo boxes that start empty actually return the correct amount of mats upon being recycled.

## Why It's Good For The Game

Unless you have an ammo box that already has space in it from using the bullets already in it, there's no efficient way to actually store loose ammunition. Say if you need to empty a magazine for whatever reason, all that ammo just ends up on the floor most of the time. You could make a cardboard box, but that only holds seven bullets, or a dufflebag but that's not really a pretty or thematic way to do it.

Being able to craft empty ammo boxes should hopefully make bullet storage more convenient compared to what we currently have.

## Changelog

:cl:
add: Blank Ammo Boxes, which can be crafted from one cardboard and iron, or at the autolathe
add: Empty ammo boxes can be crafted with blank ammo boxes
code: Ammo Box initialization now accurately makes empty boxes with the correct amount of mats.
/:cl:


FalloutFalcon:
In an ideal world we should prob just make it so you can craft an unlabeled ammo box that accepts all bullets. Give it bag logic to let you pick up specific types?
I will get to ammo crafting at some point. Will probably do this there

Anticept:
I'd also like said box to be able to unload spent ammo into lathes. That would be a huge QOL

west3436:
Why did you make empty subtypes for every ammo type instead of just removing the ammunition from a box when it's created? Your current approach just results in another failure point for future coders when adding new ammo types.

FalloutFalcon:
> Why did you make empty subtypes for every ammo type instead of just removing the ammunition from a box when it's created? Your current approach just results in another failure point for future coders when adding new ammo types.

Can you pass paramaters to an object it stack recipes? that seems like the best solution here

Gristlebee:
> Can you pass paramaters to an object it stack recipes? that seems like the best solution here

As far as Im aware, crafting and stack recipes don't currently support passing parameters to the crafted object. 

> Why did you make empty subtypes for every ammo type instead of just removing the ammunition from a box when it's created? Your current approach just results in another failure point for future coders when adding new ammo types.

It was a consequence of not being able to directly pass parameters, so the obvious solution at the time seemed to be make the new subtypes since there was already existing var. There was also some precedent for these kind of subtypes, as some guns also use subtypes to have them start unloaded without a magazine.

If I'm wrong and you can pass parameters through stack recipes, Falcon's solution would definitely be the most elegant way to do this.

west3436:
Yeah I am not sure if there's an easy way to do it then.

FalloutFalcon:
you could always make the box generic and have it accept the first ammo you load into it as the ammo type till you empty it

Gristlebee:
> you could always make the box generic and have it accept the first ammo you load into it as the ammo type till you empty it

That sounds like a good idea. I'll see about doing that. Thanks!

PR #2880, opened 2024-04-08
Movable Physics Subsystem

Sun-Soaked:
## About The Pull Request
Ports some mojave code (originally written on Daedalus by Spyroshark) hat allows you to fling ingame objects with a simple, bouncing physics quality.
This is currently implemented on bullet casings.

Technical issues currently:

- [x] Makes the bouncing less dramatic

- [x] Physics must have a "low grav" version for tiles without gravity, since mojave didn't think gravity was real

- [x] The code handling the angle physics objects travel at is fucky and needs to be fixed up

## Why It's Good For The Game
the hit feature from 2003, coming to a shiptest near you

## Changelog

:cl: Spyroshark, Sun-Soaked
add: A movable physics subsystem, deployed using a component.
add: Bullet casings now drop using movable physics
code: ports NO_PIXEL_RANDOM_DROP from TG.
/:cl:


meemofcourse:
what if we just became a mojave downstream and tested buses instead

Sun-Soaked:
fun fact testers: the solution to the angle problem was trigonometry. always pay attention in math class

code/datums/components/movable_physics.dm
```diff
@@ -0,0 +1,151 @@
...
+	var/turf/a_turf = get_turf(moving_atom)
+	if(istype(moving_atom, /obj/item/ammo_casing))
+		playsound(moving_atom, a_turf.bullet_bounce_sound, 50, TRUE)
+	else
+		playsound(moving_atom, bounce_sound, 50, TRUE)
+	moving_atom.SpinAnimation(speed = 1 SECONDS / numbounce, loops = 1)
+	moving_atom.pixel_z = z_floor
+	horizontal_velocity = max(0, horizontal_velocity + (vertical_velocity * -0.8))
+	vertical_velocity = max(0, ((vertical_velocity * -0.8) - 0.2))
+	numbounce += 1
```
west3436: numbounce is set to 1 on each call of this proc, so this does nothing

code/modules/projectiles/ammunition/_ammunition.dm
```diff
@@ -100,9 +101,14 @@
...
 
-/obj/item/ammo_casing/proc/on_eject()
+/obj/item/ammo_casing/proc/on_eject(atom/shooter)
 	forceMove(drop_location()) //Eject casing onto ground.
-	bounce_away(TRUE)
+	pixel_x = rand(-4, 4)
+	pixel_y = rand(-4, 4)
+	pixel_z = 8 //bounce time
+	var/angle_of_movement = !isnull(shooter) ? (rand(-3000, 3000) / 100) + dir2angle(turn(shooter.dir, 180)) : rand(-3000, 3000) / 100
+	AddComponent(/datum/component/movable_physics, _horizontal_velocity = rand(400, 450) / 100, _vertical_velocity = rand(400, 450) / 100, _horizontal_friction = rand(20, 24) / 100, _z_gravity = 9.80665, _z_floor = 0, _angle_of_movement = angle_of_movement)
```
west3436: `_z_gravity = 9.80665`
This should be a define (I am surprised it isn't already)

Sun-Soaked: I assume it's because mojave wanted it to be configurable- but you could just make more defines for that. I'll get it properly set up

code/datums/components/movable_physics.dm
Sun-Soaked: fuck.

Sun-Soaked:
it's slightly janky in 0g but it works now

PR #2882, opened 2024-04-09
Changed portable generator sheet burn times

Anticept:
## About The Pull Request

Portable generators all have different burn times. This unifies them.

## Why It's Good For The Game

The superpacman was actually inferior, and considerably so, to the regular pacman in total energy generated. A pacman had a time_per_sheet of 260, while a superpacman was only 85 for only double the power! This made it inferior in every way, because uranium is also quite rare.

While there are no obtainable mrspacmans as far as I am aware, I find no reason to have any of these generators set with different burn times. This simplifies it by unifying all of them into the same base time_per_sheet of 260.

I am aware this is going to mess with balance a little bit, but I believe that simplifying the amount of variables in play will help make balancing easier in the long run, because it isn't immediately obvious that there were different sheet burn times. Balance should be affected by changing power generated, uranium spawn rate, number of sheets and generators at round start... etc, not some property that you can't see unless you stand there comparing generators side by side.

## Changelog

:cl:
tweak: All pacman and its variants now have the same sheet burn times. Superpacmans are no longer the vastly inferior generator.
/:cl:


Thera-Pissed:
Uranium outputs triple the amount that plasma does, in about 1/3 of the time as plasma (86.6 vs 85). This will triple the effectiveness of uranium, which is effectively on par with plasma currently. If you want uranium to be better per sheet, you may still want to lower the base power_gen of uranium. Alternatively, the time per sheet could be shown to the player in some way.

Anticept:
I don't want to do anything in this PR but remove the differences in burn times. Further balance should be done in their own PRs.

Besides, they're power generators meant to just be turned on and off. Unless we want to make them into a reactor simulator... they should just be kept simple, as is, and mapped according to the power demand of the ship type.

FalloutFalcon:
I bet its this pr. 
https://github.com/shiptest-ss13/Shiptest/pull/2992/files#diff-88c50a053fb66bf3d7e6bc4fb8d91156017d286ba2a66f9f4e1872a7cbc0ef24
It had a bunch of stealth changes to species rendering and it never had a test merge.

FalloutFalcon:
wait shit wrong pr how did i do that

FalloutFalcon (approved):

PR #2883, opened 2024-04-10
Fixes wall locker directions

goober3:
## About The Pull Request
#2819 implemented mapping helpers for wall lockers, which required the sprite directions to be inverted, as they were previously facing the wrong way. This had the adverse effect of making every non-helper wall locker become inverted. This PR fixes that.

## Changelog

:cl:
fix: Wall lockers should be facing the right way again.
/:cl:


thgvr:
Probably just gonna send this

PR #2884, opened 2024-04-10
Removes default type space cash spawns from tesla coils & rad collectors

SomeguyManperson:
## About The Pull Request

These cant be added to cards or helm consoles and have a default sprite

## Why It's Good For The Game

Actually fixing this would require a rebalance of the output since its both passively farmable and quite large. I'm not against it but I dont think anyone is particularly for it either 

## Changelog

:cl:
del: tesla coils and rad collectors can no longer print monopoly money
/:cl:


FalloutFalcon:
It think it would make sense to gut the research mining from rad collectors or else you can prob circumvent are very limited research point handouts
*Both of them have this acctually

PR #2901, opened 2024-04-16
QoL advanced build mode 

cuddleandtea:
## About The Pull Request
Hello dear comrades! This pull request is changing QoL for admins who uses build mode. 
I truly believe that nobody is using simple build mode that allows you just to create airlocks, walls and floors while advanced have many more features for admins to use. 

## Why It's Good For The Game

With just a click of a button, administrators can access the most useful tool in build mode without disabling simple build mode every time.

## Changelog

:cl:
admin: faster access to advanced build mode

/:cl:

### DIFF
--- a/code/modules/buildmode/buildmode.dm
+++ b/code/modules/buildmode/buildmode.dm
@@ -26,7 +26,7 @@
 	var/atom/movable/screen/buildmode/preview_item/preview
 
 /datum/buildmode/New(client/c)
-	mode = new /datum/buildmode_mode/basic(src)
+	mode = new /datum/buildmode_mode/advanced(src)
 	holder = c
 	buttons = list()
 	li_cb = CALLBACK(src, PROC_REF(post_login))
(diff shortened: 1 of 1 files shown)


thgvr:
What's this doing exactly? Advanced mode is accessible from clicking the build mode button, and picking advanced. If you'd like to swap around the advanced/simple mode that's fine, but that's not what this PR does as of now. (looks like it at least)

Solution I'd prefer is making the default advanced, which this does, and adding simple as an option. It's still useful sometimes.

cuddleandtea:
> What's this doing exactly? Advanced mode is accessible from clicking the build mode button, and picking advanced. If you'd like to swap around the advanced/simple mode that's fine, but that's not what this PR does as of now. (looks like it at least)
> 
> Solution I'd prefer is making the default advanced, which this does, and adding simple as an option. It's still useful sometimes.

Hello! This PR makes that when you enter build mode - the advanced build mode is turning on by default, just as you've wrote! 

You still can change to simple buildmode by clicking the mode button. Advanced is used a way more often and admins forced every time to click the button to change from simple to advanced mode so I'm making it appear by default

thgvr:
Thanks for clarifying, this is fine then.

PR #2911, opened 2024-04-16
Nerfs kissing

Hardly3D:
## About The Pull Request
Nerfs the sound played upon kiss by halving the volume and reducing the range so you can only hear it if you are around 5 tiles to whoever was kissed, while fully blocking it behind walls.

https://github.com/shiptest-ss13/Shiptest/assets/66234359/5c41058c-8ec9-4838-a076-65a15148094a

## Why It's Good For The Game

Kissing is currently way too loud that it can be heard from nearly anywhere in most of the ships, this PR tries to fix it so it has a more reasonable range while blocking walls.

Additionally I've seen and heard a few players get uncomfortable over being able to hear some of the crewmembers get a bit kissy towards each other even when they tried to move far away from it.

I figured that 5 tiles may be a reasonable range considering that most of the rooms are around that size but I'm still open to changing it.

## Changelog

:cl: Hardly
tweak: Kissing volume and range has been reduced. Kissing cannot be heard beyond 5 ranges or behind walls.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/emote.dm
+++ b/code/modules/mob/living/carbon/emote.dm
@@ -455,7 +455,7 @@
  * This fake hit only happens if we can deal damage and if we hit a living thing. Otherwise, we just do normal on hit effects.
  */
 /obj/projectile/kiss/proc/harmless_on_hit(mob/living/living_target)
-	playsound(get_turf(living_target), hitsound, 100, TRUE)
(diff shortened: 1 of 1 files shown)


thgvr:
It's over

tontyGH:
It's over

Tsar-Salat:
It's joever

PR #2914, opened 2024-04-17
Adjusted large and plastic beaker costs, increased plastic recipe sheet creation 5x

Anticept:
## About The Pull Request

Large and XL beakers are pretty expensive. I've reduced their cost a bit. Because chemistry is more dangerous with more units in a container, you still pay a lot more for the extra space, but they're no longer crazy expensive.

Also, we don't often have a lot of access to plastic here on shiptest, and plastic gates a lot of things. The plastic from the outpost IS WAY MORE than you will EVER need in a typical round (barring something exotic that you are doing), but if no one is willing to do chemistry, it's still a reasonable thing to do.

This just makes it so that doing chemistry for plastic reasonable using the chemistry starter kit, since very few ships have a chem dispenser device.

## Why It's Good For The Game

Plastic making sucks. Now it's okay to consider a small batch. Plus large (100u) and especially XL beakers (120u) were very expensive.

## Changelog

:cl:
balance: costs reduced for large beakers (1500 glass, down from 2500) and XL beakers (1000 glass 1000 plastic, down from 2500 glass and 3000 plastic)
balance: the plastic sheets recipe now outputs 5 sheets instead of 1 sheet per reaction.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/recipes/others.dm
+++ b/code/modules/reagents/chemistry/recipes/others.dm
@@ -538,7 +538,7 @@
 
 /datum/chemical_reaction/plastic_polymers/on_reaction(datum/reagents/holder, created_volume)
 	var/location = get_turf(holder.my_atom)
-	for(var/i in 1 to created_volume)
+	for(var/i in 1 to created_volume * 5)
 		new /obj/item/stack/sheet/plastic(location)
 
 /datum/chemical_reaction/pax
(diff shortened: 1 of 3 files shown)


FalloutFalcon:
5x is . a bit much. I think your already fixing the issue tweaking the beaker recipes
If anything I would look around the items and see if there is anywhere you can add chem sources to make it.

Anticept:
Beakers aren't where I use this plastic. There's a bunch of other useful items that come up from time to time that I need to lathe and plastic is such a pain in the ass.

Anticept:
I have a better solution coming up where ghetto chem sources are significantly expanded across the board.

PR #2924, opened 2024-04-21
Intercomms, widebands, and headets defaults to on again. Handhelds continue to default to off.

### LINKED ISSUE #2923: [BUG] All widebands, AND intercoms are defaulting to speaker OFF
## Probable cause

https://github.com/shiptest-ss13/Shiptest/pull/2852

I believe this is the PR causing it. At minimum, widebands should not default to speaker off. Or at least, decouple their recording from having the speaker on.

Headsets also default off.

I don't think intercomms in general should either, as of this moment they have quite literally been made useless.

Anticept:
## About The Pull Request

This turns Intercomms, widebands, and headsets defaults to on again. HANDHELDS REMAIN OFF!

## Why It's Good For The Game

Wideband doesn't record if listening is not on. Intercomms are basically useless now unless you run around the ship turning them on. Headsets don't have a listening range, so these shouldn't default to off either.

Fixes https://github.com/shiptest-ss13/Shiptest/issues/2923

## Changelog

:cl:
tweak: intercomms, widebands, and headsets default speaker on again. Handhelds remain off by default.
/:cl:


FalloutFalcon (approved):
Fixes my foolish mistake! I would maybe argue we could leave headsets turned off since they currently just don't mater but this is a fix not a balance

Anticept:
Headsets are useful on ships with telecoms.

PR #2926, opened 2024-04-21
Makes intercomms and radios on by default again

FalloutFalcon:
## About The Pull Request

When I cleaned up some radio stuff in my add radio to box pr I turned listening to false cause i was only thinking about handheld radios where they should be off by default but intercoms and headsets should def start on

## Why It's Good For The Game
unintended change

## Changelog

:cl:
fix: intercomms and headsets now are back to on by default
/:cl:


Anticept:
Already PRed with https://github.com/shiptest-ss13/Shiptest/pull/2924

FalloutFalcon:
> Already PRed with https://github.com/shiptest-ss13/Shiptest/pull/2924

sweet thx for fixing my mistake

PR #2938, opened 2024-04-26
Removes reality integrity issues from EXOCOM

Sun-Soaked:
## About The EXO Request
A quantum superposition around the name of the EXO mining company has been isolated and destroyed by the shiptest reality force.
From hence on, all shall know only the name **EXOCOM**, rather than also, sometimes, EXOCON. 
**EXOCON is not, and was never real.** 

## Why It's Good For EXOCON (Which is not real)
EXOCON is not real. Please report any sightings of EXOCON personnel or equipment for immediate termination. EXOCON is not real.

## Nothing Has Changed. Please Remain Calm.

:cl:
fix: weird uncertainty around the name of EXOCOM. EXOCON is not real. 
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/crates/secure.dm
+++ b/code/game/objects/structures/crates_lockers/crates/secure.dm
@@ -105,7 +105,7 @@
 	else ..()
 
 /obj/structure/closet/crate/secure/exo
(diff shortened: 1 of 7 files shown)


thgvr:
1984

PR #2940, opened 2024-04-26
Generic ammo boxes and printable ammo cans

Gristlebee:
## About The Pull Request

Adds generic ammo boxes, printable in the autolathe. They can be set to hold any ammo type by using a bullet on it, and the capacity changes depending on the bullet type.

Also makes ammo cans autolathe printable.

Replaces a few instances of update_appearence with update_ammo_count in ammo box code so the materials properly update.

Thanks to FalloutFalcon for the idea

## Why It's Good For The Game

Basically a redo of my other PR, #2868 intentwise. This PR aims to make ammo storage more practical by adding these generic ammo boxes to prevent having loose rounds everywhere if you don't have a compatible ammo box with space. Ammo cans are basically just themed toolboxes, so they're just nice to have thematically as well.

Also some bug fixes I guess, that's good.

## Changelog

:cl:
add: Generic Ammo Boxes, printable in the autolathe
add: Ammo cans are printable in the autolathe
fix: Ammo boxes sometimes not properly updating their materials
/:cl:


Anticept:
You're doing awesome work.

Curious if there's room to consider the size of the round having an effect on how much the box holds?

Gristlebee:
> You're doing awesome work.
> 
> Curious if there's room to consider the size of the round having an effect on how much the box holds?

The thing is, all bullets with a few exceptions like grenades are code-wise all the same size. There's no real elegant way to check for it in the variables as far as I'm aware.

Anticept:
Got it. Sounds like a QOL for me down the road then to add something to each round type so they would have a size as well, and revisit these boxes.

Apogee-dev:
> > You're doing awesome work.
> > Curious if there's room to consider the size of the round having an effect on how much the box holds?
> 
> The thing is, all bullets with a few exceptions like grenades are code-wise all the same size. There's no real elegant way to check for it in the variables as far as I'm aware.

There’s no especially elegant solution to this right now afaik, but it’s not the end of the world if you just have to write a big dumb if-else chain to check the caliber of the round being added and define capacity based on that. if(cartridge.caliber = list(“9mm”, “.45”, “10mm”, etc)) capacity = 20 or whatever.

Gristlebee:
> There’s no especially elegant solution to this right now afaik, but it’s not the end of the world if you just have to write a big dumb if-else chain to check the caliber of the round being added and define capacity based on that. if(cartridge.caliber = list(“9mm”, “.45”, “10mm”, etc)) capacity = 20 or whatever.

Would you prefer it with the list, or is it fine as is?

Apogee-dev:
> > There’s no especially elegant solution to this right now afaik, but it’s not the end of the world if you just have to write a big dumb if-else chain to check the caliber of the round being added and define capacity based on that. if(cartridge.caliber = list(“9mm”, “.45”, “10mm”, etc)) capacity = 20 or whatever.
> 
> Would you prefer it with the list, or is it fine as is?

I’d prefer the list personally- It’s extra functionality, and if/when cartridges are updated with some kind of cartridge size/category variable it’ll be fairly easy to update the code here

Gristlebee:
> > > There’s no especially elegant solution to this right now afaik, but it’s not the end of the world if you just have to write a big dumb if-else chain to check the caliber of the round being added and define capacity based on that. if(cartridge.caliber = list(“9mm”, “.45”, “10mm”, etc)) capacity = 20 or whatever.
> > 
> > 
> > Would you prefer it with the list, or is it fine as is?
> 
> I’d prefer the list personally- It’s extra functionality, and if/when cartridges are updated with some kind of cartridge size/category variable it’ll be fairly easy to update the code here

It's been implemented. Max ammo values are open to be adjusted if needed.

FalloutFalcon:
oh you could just had var/ammo_per_box in the bullet itself

FalloutFalcon:
also we really should define our calibers.. i got that tho remind me if i forgrt

Anticept:
> oh you could just had var/ammo_per_box in the bullet itself

this is how I had thought about doing it in the future.

Apogee's list method certainly is simple, but if we ever do things like implement larger vs smaller boxes or other means of storing these, having vars for the individual bullets is more future proof.

code/modules/projectiles/boxes_magazines/generic_ammo_box.dm
```diff
@@ -0,0 +1,63 @@
...
+	/// Name of the currently set ammo type
+	var/ammo_name
+
+/obj/item/ammo_box/generic/update_ammo_count()
+	. = ..()
+	if(LAZYLEN(stored_ammo) == 0)
+		ammo_set = FALSE
+		ammo_type = /obj/item/ammo_casing
+
+/obj/item/ammo_box/generic/proc/update_max_ammo(obj/item/ammo_casing/ammo)
```
thgvr, line 18: I really don't like this proc at all, i'll think about a better method

FalloutFalcon: Agreed. you can just fetch a bullets_per_box from the ammo_casing.

Gristlebee: took falcon's suggestion, update_max_ammo is now handled by a bullet_per_box variable rather than a list checking calibers

code/modules/projectiles/boxes_magazines/generic_ammo_box.dm
```diff
@@ -16,19 +16,8 @@
...
-		// automatic rifle calibers
-		max_ammo = 30
-	else if (caliber in list("12ga"))
-		// shotgun ammo
-		max_ammo = 20
-	else if (caliber in list("8x50mmR","a858","a300",".300 BLK",".308",".50 BMG"))
-		// bolt actions, semi-auto rifles, snipers
-		max_ammo = 10
+	if(ammo.bullet_per_box)
+		max_ammo = round((ammo.bullet_per_box)*0.5)
```
FalloutFalcon: I assume this is so they are worse then normal ammo boxes? I'm unsure if they need to be worse then manufactured ammo boxes, I think it would be fine if they are equal.

FalloutFalcon (changes requested):
Everything else looks good.

code/modules/projectiles/boxes_magazines/generic_ammo_box.dm
Gristlebee: sure, I dont mind changing it to be the same then

FalloutFalcon (approved):

PR #2941, opened 2024-04-26
Fix half of verbs not showing up

SuhEugene:
## About The Pull Request

Hello comrades. This PR fixes a mistake, because of which none of `/mob` verbs was sent to the TGUI panel on client connection for clients having "Broadcast Log in/out" disabled.

## Why It's Good For The Game

Half of verbs wasn't showing up because of strange check condition completely unrelated to what's really needed to be checked.

## Changelog

🆑SuhEugene
fix: fixed incomplete TGUI Panel verb list
/🆑


TheIssueFairy:
This fixed my missing IC tab that had been bothering me for like 6 months. I owe you a million dollars.

PR #2944, opened 2024-04-28
Match bullets no longer have a highly above-average chance to ricochet into their user's foot

SomeguyManperson:
## About The Pull Request

Nothing is stopping the only person in ricochet range from being the person shooting the gun. Which looks really stupid if you are trying to do something funny and shoot yourself in the foot 6 times.

## Why It's Good For The Game

"bullet that shoots the person who shot it" isn't an ammo type

## Changelog

:cl:
tweak: match rounds will no longer consider the person firing them as a viable target when ricocheting
/:cl:


rye-rice (approved):
Huh.

PR #2950, opened 2024-05-02
Thrown Thing Improvments

FalloutFalcon:
## About The Pull Request
My original goal was to let you trigger mines with thrown items
Ports https://github.com/tgstation/tgstation/pull/60604
That pr allows thrown objects to target anything you can click on really. A side effect is you can bola a downed person so that's interesting
It also fixed a bug we had which caused finalize() to get called twice though i don't think that original pr even noticed that.
I added an hitby to landmines and other random items. Landmines trigger if they are hit with an object. the hitby for the mine can be anything more just proving it works.
I added a movement type for thrown objects, it could probably just be flying however..
An old issue meant that throwing an item above a landmine it would trigger it. This made chain reactions a lot more common. I added a check to only care about ground based movement so you will only trigger a mine if your grounded

https://github.com/shiptest-ss13/Shiptest/assets/86381784/c87a412c-a83f-49d5-a969-95bbb102e537
https://github.com/shiptest-ss13/Shiptest/assets/86381784/5f6c10e6-34f7-4e86-8267-6ea059a4a7c6
https://github.com/shiptest-ss13/Shiptest/assets/86381784/dee243fd-3a43-4027-8b1a-afe670d2be5d

Right now you can throw a person on and off a landmine tho. Will fix before undraft

## Why It's Good For The Game

Landmines can chain react but not overly so.
Targeting things that don't block the bola adds some new options
Allows for plenty of new neat interactions to be added

## Changelog

:cl:
add: You can trip landmines with a well placed throw
add: New interaction with throwing things at bottles, lights, and bear traps
add You can now target items and resting players with thrown objects. You have to click on them directly however.
fix: The end of finalize no longer gets called twice and targeting items seems like an intended behavior that was lost
/:cl:


FalloutFalcon:
should i refactor mine fields to have a center it places from with random offsets to make the mine field circular with more packed in the center?

Sun-Soaked:
> should i refactor mine fields to have a center it places from with random offsets to make the mine field circular with more packed in the center?

yeah sure thing, makes more sense then purely random distro

PR #2952, opened 2024-05-03
Minor fixes and improvement to Flavor Text

Hardly3D:
## About The Pull Request

Improves and fixes issues where flavor text would sanitize certain characters and refuse new lines while also giving new looks to the flavor text page. Bits of codes for the popup was taken from MapleStation but I can't find the specific PR for it.

Flavor text's length was also chopped from a whooping 4096 to 1024 for the sake of keeping it lag free, it should still allow you to make a pretty robust flavor text and it shouldn't really impact anyone as the average flavor text length I've seen on a round was around 362 characters, with the highest being 645 characters.

dreamseeker_RBS1nF48VM

Additionally changes the text a bit on the flavor text's edit popup so it leads to less confusion.

I'm intending to do a full refactor for flavor text itself and hopefully make use of TGUI but this might take quite a bit and I can't really make any guarantees, so hopefully this helps for now.

## Why It's Good For The Game

Flavor text has been buggy and limiting for quite a while, while this PR doesn't solve many code shenanigans it should at least help people describe how their characters a lot better and with less wall of texts.

As for the text changes; I believe this might make more sense and hopefully lead to less confusion that I'd see some have when it comes to flavor text.

## Changelog

:cl: Hardly
add: Added some style to the flavor text popup
fix: Fixed flavor text not accepting certain characters or new lines
tweak: Changed up Flavor Text editing text a bit
tweak: Flavor text's max length has been chopped from 4096 to 1024
/:cl:


Hlscientist:
Yippie!

Skies-Of-Blue:
Big if true

code/modules/client/preferences.dm
```diff
@@ -1670,9 +1670,9 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 						age = clamp(round(text2num(new_age)), pref_species.species_age_min, pref_species.species_age_max)
 
 				if("flavor_text")
-					var/msg = sanitize(stripped_multiline_input(usr, "Set the flavor text in your 'examine' verb. This can also be used for OOC notes and preferences!", "Flavor Text", features["flavor_text"], 4096, TRUE))
```
FalloutFalcon, line 1673: sanitize runs it through html_encode() , but so does stripped_multiline_input so thats fine
sanitize runs sanitize_simple with only the first param so it does ("\n"="#","\t"="#"). These aren't issues so it seems fine to remove sanitize

PR #2954, opened 2024-05-03
Empty Gun Helper

FalloutFalcon:
## About The Pull Request

Makes all the current var edited and manually typed spawnwithmag = false guns to be done with a helper you can slap under any gun you need it for

## Why It's Good For The Game

The defines are alot quicker to do if you need it for something not mapped, for mapping, it might still be better to var-edit.
At some point im probably going to detach the spawned mag for valid mag types which this will make easier to change. Why?
image
Not hating on retlaw. this was the only way to do it with how we have it set up rn

## Changelog

:cl:
refactor: spawning guns without mags are handled with a define
/:cl:


Sun-Soaked (approved):
looks guten

PR #2960, opened 2024-05-05
Nerfs the size of the Etherbor Pistol and Etherbor Rifle

Martinpachu:
## About The Pull Request

ohhh the lizard

## Why It's Good For The Game

I think they're too strong for their size, with the pistol being more or less equal to a Sharplite laser gun that is normal sized or a pistol that is normal sized, and the Etherbor Rifle being more or less equal to an SMG which is also bulky sized. The BG-16 stays as it was before.

## Changelog

:cl:

balance: The civilian etherbor weapons have been made larger (the SG-8 being normal sized and the BG-12 being bulky sized)

/:cl:

### DIFF
--- a/code/modules/projectiles/guns/faction/gezena/energy_gunsword.dm
+++ b/code/modules/projectiles/guns/faction/gezena/energy_gunsword.dm
@@ -73,8 +73,7 @@
 	desc = "Etherbor's current and sidearm offering. While intended for marines, it's also available for civillians"
 	icon_state = "kalixpistol"
 	item_state = "kalixpistol"
-	w_class = WEIGHT_CLASS_SMALL
-
+	w_class = WEIGHT_CLASS_NORMAL
 	modifystate = FALSE
 
 	wield_delay = 0.2 SECONDS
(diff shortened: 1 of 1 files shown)


goober3:
sidearms should be kept small i think. just nerf the capacity or damage instead.

goober3:
neverbmind.........

FalloutFalcon:
This does very little balance wise. They are the "best in slot" option for anyone using them and since most of them have armor they can store it on there. Not saying its a bad change it just doesn't address the core strengths they have which is insane dps for anything using a cell on-top of being bundled with high capacity cells. In other words its a bit of a nothing burger, thats fine tho!

FalloutFalcon:
> This does very little balance wise. They are the "best in slot" option for anyone using them and since most of them have armor they can store it on there. Not saying its a bad change it just doesn't address the core strengths they have which is insane dps for anything using a cell on-top of being bundled with high capacity cells. In other words its a bit of a nothing burger, thats fine tho!

Also consider everything in that file is weapon light
```
/// Allows you to dual wield this gun and your offhand gun
#define WEAPON_LIGHT 1
/// Does not allow you to dual wield with this gun and your offhand gun
#define WEAPON_MEDIUM 2
/// You must wield the gun to fire this gun
#define WEAPON_HEAVY 3
```
Even the /kalix/pgf/heavy

PR #2963, opened 2024-05-06
Adds New Religions To Replace Old Ones

FalloutFalcon:
## About The Pull Request
strong wip draft just trying to show someone code
removing alot of religion cruft along with adding some minor stuff for religions we actually want
if you have ideas for religions pitch them to me
https://hackmd.io/@FalloutFalcon/HkNHmQbjp
Requires #2936

## Why It's Good For The Game

We don't want any of the old tg religions, this is the starting point where we have actual good religions to work with.

## Changelog

:cl:
add: Added new trinkets and books for the in-lore religions, find some of them in the load-out menu!
del: Removed more religion cruft, should be nearly nothing left
/:cl:

PR #2966, opened 2024-05-07
Cuts the clown, mime, deity and religion names and loadout bibles into little itty pieces.

Martinpachu:
## About The Pull Request

Removes cruft!!!! finally!! 

## Why It's Good For The Game

im like doomguy except instead of killing demons im killing the ancient ghosts of tgstation humor
## Changelog

:cl:

del: Removed the bible from loadout, as well as the religion and deity names because god is dead. Clown and mime names too.

/:cl:


FalloutFalcon:
Do note that that this does cause minor issues if someone spawns as a chaplain as it trys to pick a religion.  It would probably just be a normal runtime. #2963 fixes this so it doesn't matter. Its fine, they don't need to do the duplicate work of cutting it out.

PR #2971, opened 2024-05-09
Disables Inteq Ships (and removes the Hound)

Apogee-dev:
## About The Pull Request

Removes all Inteq ships from normal spawning availability, and deletes the Hound entirely because that thing has not been spawned _once_ since it was replaced with the vaquero. The Talos, Vaquero, Colossus and Valor are all still available to spawn for admins.

## Why It's Good For The Game

Inteq players need to [straighten up and fly right.](https://youtu.be/vQGt4XwwFsc)
Honestly until we get whitelists for captaining on certain factions I'm pretty sure we should do this with all of the more military-themed factions, considering CLIP and Inteq's track record.

## Changelog

:cl:
del: Removed the IRMG Hound from the game
del: Made Inteq ships adminspawn-only
/:cl:


DrCrawler:
So what do you consider military-themed? Are you going to try to disable PGF, NGR and Hardliners too? Why outright ban people from playing these factions? Please give a more informative explanation to why is it good for the game.

Apogee-dev:
> Are you going to try to disable PGF, NGR and Hardliners too?

Yes.




Well, half yes. What I really want is a captain whitelist by faction, and PGF and Hardliners are definitely in whitelist territory. This is a temporary measure until then. Just ahelp if you want to play inteq.

DrCrawler:
> What I really want is a captain whitelist by faction, and PGF and Hardliners are definitely in whitelist territory.

So in the end we're just gonna have those three with whitelists? Not any other faction?

rye-rice:
> > What I really want is a captain whitelist by faction, and PGF and Hardliners are definitely in whitelist territory.
> 
> So in the end we're just gonna have those three with whitelists? Not any other faction?

Hey that's a pretty good idea actually, making other factions whitelisted, thanks for the suggestion

thgvr:
Temporary removal is, for all intents and purposes, a permanent removal.

Apogee-dev:
> Temporary removal is, for all intents and purposes, a permanent removal.

So we put a definite time scale on it instead of “when so-and-so happens.” Maybe two months.

FalloutFalcon:
Could we please have a whitelist system in place before this is merged? Just a simple menu an admin can allow a player to spawn a faction? Hell I can start some work on it (tho i already have alot of prs open). I also think these ships need to be visible but locked in the buy menu or else players who join after this may never know that you can be whitelisted for a ship.

Zevotech:
I'd rather have a more in-depth talk about this with the other maptainers before this PR is made. No offense Apogee, but I don't see this fixing anything with CLIP still being in the works and Nanotrasen mid-rework. That, and Thgvr has a point about the temporary removal.

PR #2980, opened 2024-05-11
Fixes plasma thruster acceleration

SomeguyManperson:
## About The Pull Request

If you burned the expected amount of fuel you got full thrust regardless of throttle

## Changelog

:cl:
fix: fueled thruster throttling now works right
/:cl:

### DIFF
--- a/code/game/machinery/shuttle/shuttle_engine_types.dm
+++ b/code/game/machinery/shuttle/shuttle_engine_types.dm
@@ -29,7 +29,7 @@
 	if(heat_creation)
 		heat_engine()
 	var/to_use = fuel_use * (percentage / 100) * deltatime
-	return resolved_heater.consume_fuel(to_use, fuel_type) / fuel_use * thrust //This proc returns how much was actually burned, so let's use that and multiply it by the thrust to get all the thrust we CAN give.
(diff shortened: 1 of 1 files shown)


Erring:
image
I checked this bug with plasma thrust, you still can abuse 0% throttle max thrust bug and I can't find code for plasma thrust's percentage thrust. There is a bug with plasma engines having full trust at 0 thrust. Btw, fuel consumption previously seems to be always maxed out.

SomeguyManperson:
Testing this locally it has worked, fuel expense isn't impacted and worked as intended, it's just the thrust only cared if it successfully burnt that amount of fuel and forgot to care about throttle % (so if you had your plasmas at 1% they'd burn 1% of fuel and say "yeah100% of that 1% burnt" and go at 100% speed)

Erring:
> Testing this locally it has worked, fuel expense isn't impacted and worked as intended, it's just the thrust only cared if it successfully burnt that amount of fuel and forgot to care about throttle % (so if you had your plasmas at 1% they'd burn 1% of fuel and say "yeah100% of that 1% burnt" and go at 100% speed)

Sorry, I forgot to tell you, but this bug exists even in your branch. That's where I sent the screenshot with the bug from. Here the gimmick of it is that it burns 0 percent plasma and gets full trust. Perhaps the multiplication by throttle change is lost.

SomeguyManperson:
When testing this myself it has worked fine

PR #2981, opened 2024-05-11
Does some things with customizable foods and in short you can use organs in them now

SomeguyManperson:
## About The Pull Request

things with the edible component are now viable for being added to custom foods and have support for adding filling colors
custom food fillings now take colors rather than objects because they only use colors. Technically this can maybe possibly let you add goofy ones but it's weird.

## Why It's Good For The Game

organ burge.r.......................

## Changelog

:cl:
add: you can now put organs inside burgers or on pizzas
/:cl:


generalthrax:
The Body And The Blood

FalloutFalcon (approved):
I enabled auto merge! fix your conflicts!

PR #2983, opened 2024-05-12
stops free circut board from drying rack

FalloutFalcon:
## About The Pull Request

crowbar drying rack, get all wood back plus a smart fridge board.

## Changelog

:cl:
fix: fixed free smart fridge circuit board from drying rack
/:cl:


SomeguyManperson:
Don't believe this will work, circuit is removed in deconstruct() and is apparently stored seperately from the rest of the parts and not cleared with them in drying rack Initialize

PR #2991, opened 2024-05-14
.45-70 Specialty Ammo

generalthrax:
## About The Pull Request

Adds .45-70 Match ammunition to cargo, where it used to only be available through R&D.

## Why It's Good For The Game

I love the Beacon but someone raised the very valid point that with default ammunition it hits just barely harder than slugs out of a shotgun. Giving it some options would allow it to fit its own spot on the market and release more ammo from R&D hell.

## Changelog

:cl:
add: .45-70 Match ammunition added to cargo
/:cl:


rye-rice:
if youre adding the shoulder holster, can you make it hold ONLY  a single normal sized pistol/revolver?

generalthrax:
> if youre adding the shoulder holster, can you make it hold ONLY a single normal sized pistol/revolver?

I do not know how but I will try

generalthrax:
I'm stumped so I'm just gonna do it in a separate PR when I'm not lazy

Apogee-dev:
I'm going to throw a veto out on hollow point. Yeah, it's fine on the Beacon, but on the Ashhand- which will in fact be rarely available, once we get it in as loot here and there- it turns into a very rapid two-tap without the range falloff and spread of buckshot, which is the only other readily available cartridge capable of killing in two hits right now- everything else takes 3-5. Match is fine.

generalthrax:
> I'm going to throw a veto out on hollow point. Yeah, it's fine on the Beacon, but on the Ashhand- which will in fact be rarely available, once we get it in as loot here and there- it turns into a very rapid two-tap without the range falloff and spread of buckshot, which is the only other readily available cartridge capable of killing in two hits right now- everything else takes 3-5. Match is fine.

I was under the impression that the Ashhand would be for SRM ERT only (to never see the light of day) from imaginos's pr. If Hollow Point is getting ruled out for something that's unlikely to be in game for months (possibly ever) I don't quite understand.

Apogee-dev:
> > I'm going to throw a veto out on hollow point. Yeah, it's fine on the Beacon, but on the Ashhand- which will in fact be rarely available, once we get it in as loot here and there- it turns into a very rapid two-tap without the range falloff and spread of buckshot, which is the only other readily available cartridge capable of killing in two hits right now- everything else takes 3-5. Match is fine.
> 
> I was under the impression that the Ashhand would be for SRM ERT only (to never see the light of day) from imaginos's pr. If Hollow Point is getting ruled out for something that's unlikely to be in game for months (possibly ever) I don't quite understand.

Ashhand’s slated for SRM ERTs and ruin loot. Hopefully pretty soon on the loot part. I’d rather just not put hollow point in now than have to remove it again in maybe a month.

generalthrax:
I understand that Ashhand with hollow points is ridiculous (against unarmoured targets, two shots to crit), but against targets you're more likely to face (the basic sec complement with an armour vest and helmet), they're significantly more gimmicky in that you have to cycle targets on limbs if you want to capitalize and not do below average .45-70 damage. I can't see players specifically seeking out the Rare Revolver to get the specialty ammo hollow points so they can dome unarmoured indies (rarer than you'd think) in two shots instead of three.

I want them in for the Beacon, which they'd be used for 99% of the time and, as you said, would be fine. If you really don't want them, is there anything I can muck around with here to get them in so they'd be able to work without the problematic point of the Ashhand?

Apogee-dev:
I don’t want to see .45-70 HP nerfed either, honestly. When I coded it, like .45-70 explosive, it was meant to be a high powered option (even moreso than standard .45-70) for ERTs and the like. I think the simple fact of the matter is that the Beacon should never really _be_ competing with slug shotguns. It’s a single shot break action longarm firing a large bore pistol cartridge- That’s never going to stack up against anything with similar or better stopping power and a larger magazine, and I really don’t think it should.

Honestly, I’m _very_ wary of power creep thanks to the ever increasing accessibility of guns and ammo in cargo to stand in for reduced RnD, without the tempering factor of a reputation system or anything like that (which RnD did have in the form of having to grind research points, having a tech tree, and having restricted lathes, for what little credit that earns it. With RnD at least it was only combat ships with easy access to specialized ammo types). That’s more of a big-picture complaint though

generalthrax:
Fair. But I do really hope the reputation system isn't on the backburner, because its only going to get worse from here. One sec

Apogee-dev:
Also ngl we should think about adding some other .45-70, .357, and .44 Short (whenever that gets added in the first place) variants to the game one of these days. .38’s weird cartridge variants are tons of fun.

FalloutFalcon (approved):

Sun-Soaked (approved):
yeah

PR #3004, opened 2024-05-19
Blastdoor and Atmos Holofield ID multitool syncing

Gristlebee:
## About The Pull Request

Multitools can be used on a button with it's panel open to add a blast door or holofield controller to it's buffer.

The ID can be copied onto a blast door the same way as changing it normally with a multitool. For holofields, you need to have the access mode on unlocked and it's panel closed.

Adds holofield controllers to the autolathe.

Changes all the span classes to span macros

## Why It's Good For The Game

When you have a roundstart blast door break or holofield break for whatever reason, it's almost impossible to replace and properly resync without ripping out the rest of the blastdoors or the other holofield. This should make reconstruction more convienent. 

Also being able to make and use your own holofield controller adds a more intuitive way to rig up player made holofields, as currently you have to use signallers on the activation wire.

## Changelog

:cl:
add: Autolathe can print holofield controllers.
add: You can sync blastdoors and holofields to a controller in a button with a multitool
code: span macros for blast doors, buttons and holofields
/:cl:


TheLeopold:
Does it work with multiple buttons having the same set of doors?

Gristlebee:
> Does it work with multiple buttons having the same set of doors?

~Should be able to, yes.~
To be more specific, as long as the buttons have the same ID code, yes.

code/game/machinery/buttons.dm
```diff
@@ -68,34 +68,40 @@
 			default_deconstruction_screwdriver(user, "button-open", "[skin]",W)
 			update_appearance()
 		else
-			to_chat(user, "<span class='alert'>Maintenance Access Denied.</span>")
+			to_chat(user,span_alert("Maintenance Access Denied."))
```
FalloutFalcon: insane nitpick
```suggestion
			to_chat(user, span_alert("Maintenance Access Denied."))
```

code/game/machinery/shieldgen.dm
```diff
@@ -511,9 +505,29 @@
 
 /obj/machinery/power/shieldwallgen/atmos/proc/can_be_rotated(mob/user, rotation_type)
 	if (anchored)
-		to_chat(user, "<span class='warning'>It is fastened to the floor!</span>")
+		to_chat(user, span_warning("It is fastened to the floor!"))
 		return FALSE
 	return TRUE
+/obj/machinery/power/shieldwallgen/atmos/attacked_by(obj/item/I, mob/living/user)
+	if(I.tool_behaviour == TOOL_MULTITOOL)
```
FalloutFalcon:
```suggestion
	return TRUE

/obj/machinery/power/shieldwallgen/atmos/attacked_by(obj/item/I, mob/living/user)
	if(I.tool_behaviour == TOOL_MULTITOOL)
```

FalloutFalcon (approved):
looks good, might just merge since i doubt anyone will test this in a test merge!

PR #3005, opened 2024-05-19
Holopad Caller History

Gristlebee:
## About The Pull Request

Holopads now show the location of the last holopad that called them on examine.

## Why It's Good For The Game

While having to shamefully ask over wideband to have someone call again after you missed their call is funny, knowing where to callback if you miss someone should help streamline ship to ship communication and improve quality of life.

## Changelog

:cl:
add: Holopads now display the location of the last holopad that called them on examine.
code: Span macros for hologram.dm
/:cl:

### DIFF
--- a/code/game/machinery/hologram.dm
+++ b/code/game/machinery/hologram.dm
@@ -78,6 +78,8 @@ Possible to do for anyone motivated enough:
 	var/secure = FALSE
 	/// If we are currently calling another holopad
 	var/calling = FALSE
+	/// The last holopad that called this one.
+	var/caller_history
 
 /obj/machinery/holopad/Initialize()
 	. = ..()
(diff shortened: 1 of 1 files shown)


code/game/machinery/hologram.dm
```diff
@@ -374,6 +378,7 @@ Possible to do for anyone motivated enough:
 	for(var/I in holo_calls)
 		var/datum/holocall/HC = I
 		if(HC.connected_holopad != src)
+			caller_history = get_area(HC.calling_holopad)
```
FalloutFalcon: Seems like this would be better to not have to deal with reference cleanup
```suggestion
			caller_history = get_area_name(HC.calling_holopad)
```

FalloutFalcon (commented):
Good QOL

PR #3009, opened 2024-05-20
Separates bleeding from damage, gauze now STOPS bleeding rather than postponing it

SomeguyManperson:
image

image

Bleeding rate is currently the same as it used to be (0.013 bleeding per damage) but can be increased beyond a limb's maximum damage from further attacks. Bleeding from damage caps at 2 per limb so exsanguinating someone isn't as easy as dumping lead into them until their blood falls out, and is hard-capped at 30 per limb for more direct sources such as throat cutting and heparin
Blunt weapons require a force of 10 or higher, and for the limb to have 25 damage to cause bleeding
Sharp weapons (including projectiles) need a force of 5 or higher, and for the limb to have 10 damage to cause bleeding
Heparin now causes existing bleeding to get worse, rather than magically creating internal bleeding

Bleeding is addressed through gauze, tape, cauterization or chemicals, with the former three halting current bleeding while it is being treated. Better equipment (primarily real medical gauze) heals faster.

Numbers are non-final and speculative

## Why It's Good For The Game

Causes bleeding to actually come from somewhere and be similarly addressed directly, gauze now exists to actually stop bleeding instead of being re-applied until the brute damage is fixed

## Changelog

:cl:
tweak: bleeding is now stored in the limbs, functioning similarly to bone breaking. Taking damage over a certain threshold and amount (lower for sharp weapons) will cause part of it to be turned into bleeding.
tweak: gauze, tape, and bleeding suppression are similarly no longer abstracted into a "bleed suppression" value
rscadd: you can now cauterize bleeding with a lighter. Which is cool.
rscadd: you can now also cauterize bleeding with suit storage decontamination. Which is hot.
rscadd: examine and examine closely will show whether or not someone is visibly bleeding or bandaged. This means people who are both will show both. Examine closely additionally shows which limbs are currently bleeding.
tweak: heparin now causes existing bleeding to worsen, instead of causing bleeding on its own
tweak: you can no longer cut the throat of someone who's head has been lopped off
/:cl:


Slim-Pickems:
Major problems with dead bodies bleeding out now. More of a teething issue than a bug if this is supposed to happen. The lack of something letting you know which limb is bleeding is also a big problem.

SomeguyManperson:
> Major problems with dead bodies bleeding out now. More of a teething issue than a bug if this is supposed to happen. The lack of something letting you know which limb is bleeding is also a big problem.

Bodies shouldn't bleed terribly much more than they used to, considering the same amount of care put into preventing blood from being dragged out

Limb damage healing is now significantly more effective at removing bleeding (if bleeding has been worsened by something it will still need to have gauze applied)

Help intent examining yourself now displays openly bleeding limbs

FalloutFalcon:
Did you fix the issues of last uptime and have it ready for test merges again?

SomeguyManperson:
minor misinput should be fine now

code/modules/food_and_drinks/kitchen_machinery/microwave.dm
```diff
@@ -385,6 +385,12 @@
 			playsound(src, 'sound/items/cig_light.ogg', 50, 1)
 			moveToNullspace()
 
+
+/obj/item/ration_heater/get_temperature()
+	if(!uses)
+		return 0
+	. = ..()
+
```
FalloutFalcon, line 393: i looked at the attackby and it looks like you would be use this infinity to cauterize is that correct?
it also SHOULD qdel itself after it has no uses left so this shouldn't realllly matter

FalloutFalcon (approved):
I like all the code, project zomber, plently of test merges

PR #3011, opened 2024-05-20
Prismwine's burn resistance is no longer goofy

SomeguyManperson:
## About The Pull Request

currently it uhhhhhhhhh stacks with itself :3

## Why It's Good For The Game

fix

## Changelog

:cl:
fix: prismwine's burn resistance can no longer end up stacking
/:cl:


FalloutFalcon:
Yea I have no clue why I had that. I have a pr in the works improving trick wines all around, in that i fix this but it might be a minute so this is good to atomize from it anyway.

FalloutFalcon (approved):

PR #3012, opened 2024-05-20
Asteroid ore veins

Gristlebee:
## About The Pull Request

Adds asteroid ore veins to asteroid generation, with the spawning weighted toward higher tier veins

## Why It's Good For The Game

Since the mining rework, there hasn't been much reason to visit asteroids anymore as their ore generation has been slashed and ore veins present a much more efficient way to mine.

Having high tier ore veins on the asteroids should make them desirable places to mine again, and a reliable if dangerous option to find sites the higher tier drilling missions at the outpost. Having to contend with zero gravity and EVA should present an interesting challenge while drilling.

## Changelog

:cl:
add: Asteroid ore veins
/:cl:

### DIFF
--- a/code/datums/mapgen/planetary/AsteroidGenerator.dm
+++ b/code/datums/mapgen/planetary/AsteroidGenerator.dm
@@ -160,11 +160,11 @@
 	)
 
 	feature_spawn_list = list(
-		/obj/structure/geyser/random = 10,
+		/obj/structure/geyser/random = 5,
 		/obj/structure/spawner/mining/carp = 5,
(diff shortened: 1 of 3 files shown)


SomeguyManperson:
Megacarp get randomized bonus stats on init which (if it works properly) will give them 20-60 extra health

Gristlebee:
> Megacarp get randomized bonus stats on init which (if it works properly) will give them 20-60 extra health

aaaah, I see. Megacarp should be fine as is then I think. Ill drop the buffs

PR #2618, opened 2024-01-02
Hunter's Pride

rye-rice:
## About The Pull Request

image

Resprites the entirety of the Hunter's Pride gun manufacturer and adds some extra weapons, specifically for the SRM.

As Hunter's Pride's lineup also intersects with the SRM weapon lineup, this was very interesting to do. However, thanks to the SRM lorema and lorehead in general Valorium for helping clear things up, and as usual Apogee-dev for any gun-related work I do. Without them this would have been much harder and much lower quality than how it came out. In fact, Apogee went above and beyond and designed two entirely original guns for this, the Firestorm and Vickland!

*"An unconventional submachinegun, rarely issued to Saint-Roumain Militia mercenary hunters for outstanding situations where normal hunting weapons fall short. Chambered in .45."*
image

*"The pride of the Saint-Roumain Militia, the Vickland is a rare semi-automatic battle rifle produced by Hunter's Pride exclusively for SRM use. It is unusual in its class for its internal rotary magazine, which must be reloaded using stripper clips. Chambered in .308."*
image

Special thanks to FleetingExpertise for drawing the Ammo box!

Double barreled shotguns and the Beacon (formerly the contender), also now work like actual break actions! Use the Unique Action key to break them open, load them up, then hit it again to close them.

Many other weapons are given sawn off options as the new gun recoil/twohanded mechanics means theres a better tradeoff to sawing them off rather than haha weapon in bag

Also, disposable guns are now 22lr with 10 round magazines. They're still horrible, just somewhat less so.

There's various smaller changes in the changelog I probably didn't mention here, go take a look at them.

Requires #2599

## Why It's Good For The Game

image

## Changelog

:cl: retlaw34, FleetingExpertise, Apogee-dev, Valorium,
add: Resprites almost every single hunter's pride weapon
add: Added the Vickland, SRM equivlent to the ARs except it's an battle rifle
add: The Illestren is now an en-bloc rifle and uses 8x50mmR. Functionally it's the exact same to aquire the ammo, just with a different name
add: You are able to saw off many more guns than before, just be sure to actually empty it before doing so!
add: Many Hunter's pride weapons have been renamed to something befitting the lore
add: Break action weapons
fix: Traubuco should work correctly again
fix: Some broken gun onmobs should now show up
tweak: guns tell you how to rack them or switching firing modes
balance: smgs fire 0.02 seconds faster and now have standard pistol recoil onehanded
[...]

PR #3014, opened 2024-05-21
Bandoliers now hold 40 items as intended

SomeguyManperson:
## About The Pull Request

Bandolier max items was set to 40 but max weight was not, meaning it only held up to 14 total weight which is significantly less than 40

## Why It's Good For The Game

Fix

## Changelog

:cl:
fix: bandoliers now hold the correct amount of bullets (several more)
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -698,6 +698,7 @@
 	. = ..()
 	var/datum/component/storage/STR = GetComponent(/datum/component/storage)
 	STR.max_items = 40
+	STR.max_combined_w_class = 40
 	STR.display_numerical_stacking = TRUE
 	STR.set_holdable(list(
 		/obj/item/ammo_casing


Imaginos16:
How long was this broken? This is wild, no wonder bandoliers were so useless.

tmtmtl30:
40 is a fuckin' hell of a lot though

SomeguyManperson:
This was apparently boosted from 18 in #2618 which also allowed them to hold non-shotgun rounds. Since they can only hold ammo casings the bandolier is only super useful if you are using directly loaded weapons like revolvers, lever actions and shotguns (shotguns specifically probably being a :/ here) but using one does take your belt slot

Zevotech:
Shotgun ammo boxes already hold a hell of a lot of rounds and fit in pockets/webbings. What the bandolier does different is it allows you to hold a greater variety of easily-accessed ammunition in exchange for the belt slot. You could use this as a way to carry a bunch of different shotgun ammo types, or carry both shotgun and revolver ammo in the same place.

FalloutFalcon:
Shotgun ammo boxes hold 25 so this seems plenty fair for the belt slot.

PR #3024, opened 2024-05-25
Broken bones now hurt less on rollerbeds

### LINKED ISSUE #2968: [BUG] Roller beds still cause broken bones to jostle
Issue reported from  Round ID: 2375 (Shiptest)

Reporting client version: 515.1634

Reproduction.
Put someone with broken bones on a roller bed
drag them halfway across the planet
image

SomeguyManperson:
## About The Pull Request

fixes #2968

splints would work for some stuff but they are both uncommon (only spawn in like brute kits/need to be crafted) and don't work on the head or chest. Also gives rollerbeds a niche

## Why It's Good For The Game

Life or random damage ticks for broken cat

it's a bit unintuitive that putting someone on a rollerbed won't make them resistant/immune to bone jostling

## Changelog

:cl:
balance: Rollerbeds stop broken bone damage ticks from occurring while buckled
/:cl:


Anticept:
This will also cause fireman carrying, piggybacking, borgriding, and other technically buckled but not really stabilized ways of preventing bone jostling too.

Cant remember if skate/hoverboarding is buckling too.

SomeguyManperson:
> This will also cause fireman carrying, piggybacking, borgriding, and other technically buckled but not really stabilized ways of preventing bone jostling too.
> 
> Cant remember if skate/hoverboarding is buckling too.

I kind of forgot about vehicles

Anticept:
You should be able to do a simple check to see what they are buckled TO and whitelist.

FalloutFalcon (approved):

Anticept:
I wanted to come back to this, and suggest a better way to handle, but I didn't make it in time.

I was thinking more about this, and I think for long term maintainability, the true, best way to handle is to check for a flag on an object. Sure, right now we only have rollerbeds, but that might change.

I guess though if that day comes up, it can be changed then.

PR #3067, opened 2024-06-04
SSD Indicators Will Be Real In 2020

Skies-Of-Blue:
## About The Pull Request
Loosely ports https://github.com/Skyrat-SS13/Skyrat13/pull/420, splicing Novasector's SSD mechanics with our own.

When a player goes SSD, they initially remain awake and gain a Zz icon above the character indicating they are away. 
<details>
  <summary>look here!</summary>

nyaru goes afk

</details>

After the player has been disconnected for three minutes, their character will resume our current implementation of SSD behavior - losing the Zz icon and falling asleep.

<details>
  <summary>see?</summary>

nyaru has been gone a while, huh

</details>

This also adds the time spent SSD to examine text.
<details>
  <summary>it's helpful!</summary>

taking a bit

</details>

## Why It's Good For The Game

Our current implementation of being SSD leaves some things to be desired. Whenever you disconnect, your character instantly passes out. This can lead to:
- dropping critical items in dangerous situations, which is just sort of the worst 
- aghost causing your character to faint on the spot, something I've heard more than a few admins complain about over the last few years
- players *reacting* to your character falling asleep as if it was a planned thing within your control
- players moving your character to a bed, disrupting a scene you may have been intent on continuing if it weren't for that pesky internet cut
- SSD players flooding the nearby chat with *snore out of nowhere

Of course, there's merit to our current implementation as well. Having characters fall asleep keeps immersion clean, and having an SSD Indicator can be immersion breaking. This is why I've opted to remove the indicator and have characters fall asleep after a grace period of three minutes. It allows players a time to reconnect, or admins time to view what they needed to, without all of the disruptions listed above.

## Changelog

:cl:
add: an SSD Indicator for when you have been disconnected for less than three minutes
balance: players will now remain awake for three minutes after disconnecting, with a new SSD icon to boot!
balance: players SSD longer than three minutes will lose their icon and fall asleep, much like the previous behavior
/:cl:


FalloutFalcon (approved):

Anticept:
> Of course, there's merit to our current implementation as well. Having characters fall asleep keeps immersion clean, and having an SSD Indicator can be immersion breaking.

I think you and skyrat have a good approach to this. No matter how that cake is sliced, **it is getting disconnected is the immersion break.** SSD/indicators are just a way to smooth over the immersion break and guide it back to the game, and I think what you did here is great.

Skies-Of-Blue:
This should probably be testmerged first on the off chance I missed a few edge cases.

goober3:
> This should probably be testmerged first on the off chance I missed a few edge cases.

i would like to see how this interacts with our helm console SSD autostop system. does that even work actually.

Skies-Of-Blue:
> i would like to see how this interacts with our helm console SSD autostop system. does that even work actually.

Doesn't touch that in the slightest, works exactly like it did before!

Skies-Of-Blue:
Well that first TM went terribly, this should be ready for round two.

PR #1900, opened 2023-04-14
Ports TG's Fishing and Beam Refactor

BogCreature:
## About The Pull Request
Adds god's most holiest of features and the refactor code needed to run it. Also adds in new missions to add even more fishing related gameplay depth. Makes beach planets no longer populated by flying fish too, god rest their souls.
image

Ports these TG Station PRs:
https://github.com/tgstation/tgstation/pull/67691
https://github.com/tgstation/tgstation/pull/56343
https://github.com/tgstation/tgstation/pull/55193
(And probably some minor bits of other ones I'm forgetting)

## Important Image
588824fad757cd7c47e56f12b604dab6119dd3815ed4843f971e725e16609cbb_1

## Why It's Good For The Game
Adds a fun yet rewarding downtime activity. Also, you look at beach planets and tell me those weren't meant to be fished on!

## Changelog

:cl:
add: TG fishing minigame
add: Aquariums
add: The ability to fish on beach, lava, and jungle planets
add: New outpost missions
add: New fishing related outpost crates
del: Removed flying fish from beach planets and made them into actual fish
tweak: Increased the amount of outpost missions that generate at once from 15 to 20
refactor: Refactored beam code to use signals
/:cl:

PR #3091, opened 2024-06-09
Permanently and horribly kills decay generators while fixing actual engines

SomeguyManperson:
## About The Pull Request

Radiation collectors were nerfed in an undocumented change in #1900 which reduced the amount of power they could effectively generate to the point the 1 KW handout the collector provides was making up the majority of its output
This makes the following changes, which should keep radiation-based generators effective while removing the effectiveness of decay generators almost entirely

Radiation pulses now need to have a strength of over 300 to trigger a radiation collector (this is subtracted from the collector's incoming power, too), which is conveniently the same amount as produced by the strongest set of radiation barrels. If you find a decayed supermatter, go ham I guess
coefficient returned to 100, from 10, which boosts the amount of produced power from radiation. This offsets the loss of power from the increased minimum.
power drain rate returned to 4%, from 0.1%, increasing the amount of power the collectors effectively generate

Current numbers place the twinkleshine's unmodified supermatter at ~40 kW per collector, numbers can be adjusted if necessary

## Why It's Good For The Game

In the event someone wants to spend 5 zillion credits on a supermatter engine, or the twinkleshine is run again, it might as well produce a reasonable amount of power

## Changelog

:cl:
balance: radiation collectors are now significantly more stingy about the potency of the radiation they will process. Don't expect old nuclear waste or openly stored uranium to do much.
balance: radiation collectors also produce significantly more power from the remaining viable sources
/:cl:


FalloutFalcon (approved):
not super informed on radiation stuff but this sounds fine in theory

Imaginos16:
Alright let's get this one under way as well

PR #3095, opened 2024-06-10
removes randomspread

FalloutFalcon:
## About The Pull Request
removes random spread from ammo, guns, and mechs

## Why It's Good For The Game
cruft var that either had no effect, guns, or didn't matter since there was already random spread being applied from another source, mechs

## Changelog

:cl:
del: Removed random spread var
/:cl:

### DIFF
--- a/code/datums/components/pellet_cloud.dm
+++ b/code/datums/components/pellet_cloud.dm
@@ -89,7 +89,7 @@
 //The arguments really don't matter, this proc is triggered by COMSIG_PELLET_CLOUD_INIT which is only for this really, it's just a big mess of the state vars we need for doing the stuff over here.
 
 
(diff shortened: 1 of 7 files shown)


Kapu1178:
Dunno why you're removing this when it's useful and actually works. Really stretching the definition of "Cruft". "Random spread" means that the spread per bullet will be completely random within a given range of spread. "Smart spread" (randomspread 0) means that each bullet will be evenly spaced to fill the given range in a fixed pattern.

FalloutFalcon:
> Dunno why you're removing this when it's useful and actually works. Really stretching the definition of "Cruft". "Random spread" means that the spread per bullet will be completely random within a given range of spread. "Smart spread" (randomspread 0) means that each bullet will be evenly spaced to fill the given range in a fixed pattern.

I only tested it with single shot weapons.. looking at the code I understand what you mean idk what I was doing in this pr I think I jumped the gun.

PR #3097, opened 2024-06-10
Removes generic_adjective sanity check

Anticept:
## About The Pull Request

Removed the sanity check I had put in for character adjectives during a previous fix.

It was well intentioned, but as it turns out, if someone is loading a charslot that existed before guestbook was ported or never saved with one, they wouldn't have an adjective saved in their save file, resulting in nulling the variable for generic_adjective, which causes this sanity check to fail and the selection not to show. The goal was meant to prevent a guestbook malfunction from locking up prefs as runtimes can break it, but after spending hours and hours on going over prefscode and analyzing character loading and generation, it's clear that I would have to address this at the point of charslot loading.

I just want this fixed in time for wednesday uptime. BETTER handling will come later, in a separate PR, when I am SURE that they won't break charslots, as well as some other prefs fixups. There's a lot of code in prefs that does not work or do anything, but for now, let's keep it simple. Should be ready to testmerge or full merge.
## Changelog

:cl:
fix: Generic adjective selection will show for everyone.
/:cl:

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -879,18 +879,19 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 					dat += "</td>"
 					mutant_category = 0
 
-			if(generic_adjective)
-				if(!mutant_category)
-					dat += APPEARANCE_CATEGORY_COLUMN
+			// begin generic adjective
+			if(!mutant_category)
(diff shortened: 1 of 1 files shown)


Anticept:
Still working on this. Prefs code is a monster of a clusterfuck and I want to do this right.

FalloutFalcon:
This has had plenty of test merges

PR #3107, opened 2024-06-14
adds telegraphing to brimdemon attacks

SomeguyManperson:
## About The Pull Request

Brimdemon beams now fade in before activating

Reduces brimdemon base vision range from 9 to 6 so you won't piss them off before even being able to see them

https://github.com/shiptest-ss13/Shiptest/assets/24857008/994e9111-7504-43e2-bfea-ad05e86e6399

## Why It's Good For The Game

Silence, tester. Offscreen brimdemon blast. -35.

## Changelog

:cl:
balance: brimdemon beams now have a fade-in effect before dealing damage, making it harder to get instantly chunked for a quarter of your health
balance: brimdemons can no longer aggro on you before you can see them
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/mining_mobs/brimdemon.dm
+++ b/code/modules/mob/living/simple_animal/hostile/mining_mobs/brimdemon.dm
@@ -150,7 +150,6 @@
 	balloon_alert(src, "brimbeam fired")
 	cut_overlay("brimdemon_telegraph_dir")
 	for(var/obj/effect/brimbeam/beam in beamparts)
-		beam.alpha = 255
 		var/turf/affected_turf = get_turf(beam)
 		START_PROCESSING(SSfastprocess, beam)
 		for(var/mob/living/hit_mob in affected_turf.contents)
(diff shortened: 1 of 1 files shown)


FalloutFalcon (approved):
It kinda seems like they are already pretty well telegraphed but this is fine as a Nerf since I do see them killing people a ton, reducing the range is also good.

Apogee-dev:
You know, I almost feel like it’d be better to make the beam deal damage over multiple ticks instead so a grazing hit isn’t as bad as standing in the beam for the full duration and in fact doesnt do much (i dunno, break it up into 7 sequential shots of 5 damage??), but that’s a fair bit more work.

SomeguyManperson:
> You know, I almost feel like it’d be better to make the beam deal damage over multiple ticks instead so a grazing hit isn’t as bad as standing in the beam for the full duration and in fact doesnt do much (i dunno, break it up into 7 sequential shots of 5 damage??), but that’s a fair bit more work.

it does both currently

Imaginos16:
It has been test merged on Wednesday and nothing bad happened, so I'll get this merged.

PR #3113, opened 2024-06-16
Reduces the orice of 9mm bullets so they arent objectively worse than anything else

SomeguyManperson:
## About The Pull Request

Presently, 9mm loses 20% damage with no other direct stat changes
This pr reduces the price of 9mm boxes by 20% so the bullets aren't as expensive as any other standard pistol bullet while dealing less damage

## Why It's Good For The Game

Usually bullets are significantly more expensive to counteract QOL stuff like killing things significantly faster or being fired from a contender and 9mm does neither

## Changelog

:cl:
balance: 9mm boxes are now cheaper than other pistol calibers to account for the lower damage per bullet
/:cl:


FalloutFalcon (approved):
Good change, not much to note

Imaginos16:
Sure

PR #3116, opened 2024-06-16
Mob trophies will no longer attempt to drop if there aren't any

SomeguyManperson:
## About The Pull Request

killing 200000 goliaths so I can harvest nothing from them and upgrade my crusher with no trophies

## Why It's Good For The Game

runtime I think

## Changelog

:cl:
fix: mobs will no longer stick around after being butchered sometimes
/:cl:


FalloutFalcon (approved):
oops. I really did not test that change it seems like, I don't remember encountering this issue, thanks for the fix.

PR #3124, opened 2024-06-18
tailthump emote tied to tail instead of species

generalthrax:
## About The Pull Request

you can now tailthump if you have a tail organ instead of being specifically tied to srothy (vox can still tail thump even though they have no tail organ)

## Why It's Good For The Game

this was made for elzu in mind but you could do it if you have like a fox tail or something. the thumpers

## Changelog

:cl:
add: You can now tail thump if you have a tail instead of it being tied to species
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/emote.dm
+++ b/code/modules/mob/living/carbon/human/emote.dm
@@ -106,7 +106,7 @@
 /datum/emote/living/carbon/human/tailthump/get_sound(mob/living/user)
 	if(!ishuman(user))
 		return
-	if(islizard(user) || (isvox(user)))
+	if(!isnull(user.getorgan(/obj/item/organ/tail)) || (isvox(user)))
 		return 'sound/voice/lizard/tailthump.ogg' //https://freesound.org/people/TylerAM/sounds/389665/
 
 /datum/emote/living/carbon/human/weh //lizard


Anticept:
Technically, this enables the sound for anyone with a tail. Tailthump itself isn't species specific, at least not in this file.

Imaginos16:
> Technically, this enables the sound for anyone with a tail. Tailthump itself isn't species specific, at least not in this file.

Is it that much of an issue though? Any animal with a tail can in theory thump the ground with it. Maybe someone with a cat tail doing it is a bit weird, but I think it's weirder that elzu or fox people with the tails *can't* when they very much *should* be able to.

thgvr:
I don't think this should work for the smaller tails we added (rabbit,dog, etc)

Anticept:
> > Technically, this enables the sound for anyone with a tail. Tailthump itself isn't species specific, at least not in this file
> 
> Is it that much of an issue though? Any animal with a tail can in theory thump the ground with it. Maybe someone with a cat tail doing it is a bit weird, but I think it's weirder that elzu or fox people with the tails _can't_ when they very much _should_ be able to.


There is a misunderstanding. I am pointing this out because the emote is already available to all. What is being changed is making the sound play for all. I am pointing out that the changelog and PR description are slightly misleading.

Imaginos16:
Well, I trust players will be able to use this emote when appropriate.

PR #3135, opened 2024-06-21
Gives the lever-action rifle a gunslinger ability and removes the gunslinger damage buff from the shadow.

Martinpachu:
## About The Pull Request

Gives the lever-action rifle an effect similar to revolver fanning, but slower and while wielded. ALSO removes the +5 damage and AP
## Why It's Good For The Game

The lever action rifle right now is very underwhelming and i think this would be a fun way to use it. Also, the +5 damage and AP bonus is stupid and unneeded now that the .44 roumain round is merged.
## Changelog

:cl:
add: A technique for gunslinger with the lever-action flaming arrow! Firing it while wielded will quickly and automatically rack the lever, less quickly than a .38 revolver or fanned shadow though.
del: The +5 damage and AP bonus on the shadow while using it as a gunslinger.

/:cl:


FalloutFalcon (approved):
gunslinger should feel cool.. no random buffs

PR #3144, opened 2024-06-23
removes gondola transformation

generalthrax:
## About The Pull Request

Removes gondola transformation from the Tranquility reagent (it still remains for a flavour profile for gondola pockets)

## Why It's Good For The Game

We already removed transformation toxins from the game but this remained. No longer

## Changelog

:cl:
del: Tranquility reagent no longer turns you into a gondola
/:cl:


thgvr:
Just remove Tranqulity i dont think it does anything else

Tsar-Salat:
Maybe just replace the tranquility with Pax? Would keep the reference without the bad transformation

Then as headmin said, remove tranquility reagent as it's now pointless.

thestripes:
I see no reason why we'd even want to keep the reference at all, admittedly

generalthrax:
I'll just axe the reagent I guess. I'll be like 15 minutes

PR #3066, opened 2024-06-04
removes blob

FalloutFalcon:
## About The Pull Request

cruft

## Changelog

:cl:
del: Removed blob and alot of stuff close to blob
/:cl:

PR #3149, opened 2024-06-24
removes blob_act

FalloutFalcon:
## About The Pull Request

atomization of #3066

## Why It's Good For The Game

cruft

## Changelog

:cl:
del: Removed blob_act
/:cl:


thgvr:
I don't think its necessary to atomize removing blob, it's weirder to remove it in bits since it won't even be functional. The issue arises when it is several different things being changed/removed all at once so it's hard to check everything that goes through.

FalloutFalcon:
ok! less work for me

PR #3166, opened 2024-06-28
Reduces ship spawn limit for everything to 1

Apogee-dev:
## About The Pull Request

Reduces the spawn limit for every ship to 1, and makes the default limit for new ships 1.

## Why It's Good For The Game

Nearly every ship was already set to a limit of 1. Making it the default just saves time. Most of the handful of ships that still had a cap of 2- mudskipper, li tieguai, and delta, for instance - were ships that would just compete for pop or nudge each other out of their niche if double-spawned, and a greater variety of ships are available to accept players after existing ones fill up than there previously were.

## Changelog

:cl:
balance: reduced default ship spawn limit to 1.
/:cl:

### DIFF
--- a/_maps/configs/independent_beluga.json
+++ b/_maps/configs/independent_beluga.json
@@ -12,7 +12,7 @@
 		"Service"
 	],
 	"starting_funds": 4000,
-	"limit": 2,
+	"limit": 1,
 	"job_slots": {
 		"Captain": {
 			"outfit": "/datum/outfit/job/independent/captain",
--- a/_maps/configs/independent_kilo.json
+++ b/_maps/configs/independent_kilo.json
@@ -14,7 +14,6 @@
 	"map_short_name": "Kilo-class",
 	"starting_funds": 1500,
 	"map_path": "_maps/shuttles/independent/independent_kilo.dmm",
-	"limit": 2,
 	"job_slots": {
 		"Captain": {
 			"outfit": "/datum/outfit/job/independent/captain/western",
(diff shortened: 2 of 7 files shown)


SomeguyManperson:
The mudskipper has about the same crew compliment as a kilo and is only picked more often because it has an interior space larger than five rooms full of goop

Apogee-dev:
> The mudskipper has about the same crew compliment as a kilo and is only picked more often because it has an interior space larger than five rooms full of goop

yeah exactly. the mudskipper has the same crew complement, but it's better equipped. it should have the same limit as literally everything else that isn't the kilo, including other ships with crews smaller than 10 players (vaquero, delta, etc)

generalthrax:
Might wanna change spawn limit of the Li Tieguai and Shetland to 1 while you're at it.

Apogee-dev:
> Might wanna change spawn limit of the Li Tieguai and Shetland to 1 while you're at it.

good point yeah. Shit, I'll make the default limit 1 and just set kilo to 2. Almost every ship has a limit of 1 anyway, it makes more sense to make that the default.

Apogee-dev:
Updated the PR description and title to account for the expanded scope.

FalloutFalcon:
Having a ship limit per amount of people connected might be a good idea down the line. 
Say you set it at 30 people so once we pass 60 it opens another mudskipper slot or something like that.

I honestly think every ship can be 1 right now. I don't see the reason for the exception.

FalloutFalcon:
And keep in mind.. people can still just ahelp for another mudskipper or kilo slot if we are high pop... I will miss the low pop two mudskippers with a combined 3 crew

Apogee-dev:
Yeah fair enough, I'll cut the exception for the Kilo

PR #3167, opened 2024-06-29
Makes .44 and .45 not the exact same cartridge

generalthrax:
## About The Pull Request

Increases .44 roumain damage to reflect the fact that .44 is a significantly more powerful round than .45

## Why It's Good For The Game

same
martin
chu

## Changelog

:cl:
balance: .44 is no longer near-indistinguishable from the far less powerful .45 caliber, about 5 more force
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/bullets/revolver.dm
+++ b/code/modules/projectiles/projectile/bullets/revolver.dm
@@ -157,15 +157,15 @@
 
 /obj/projectile/bullet/a44roum
 	name = ".44 roumain bullet"
-	damage =  25
+	damage =  30
 
 /obj/projectile/bullet/a44roum/rubber
 	name = ".44 roumain bullet"
-	damage =  7
+	damage =  10
 	stamina = 38
 	armour_penetration = -20
 
 /obj/projectile/bullet/a44roum/hp
 	name = ".44 roumain bullet"
-	damage =  45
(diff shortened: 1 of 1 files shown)


rye-rice:
does this not make it as good as 357?

generalthrax:
It's actually a more powerful cartridge than .357, funny enough. I'm not sure if anyone wants a 35 force cartridge though.

rye-rice:
smiles evily

i think apogee wanted 357 to be more powerful than .44 short though, i need to ask though

generalthrax:
I'll wait on Martin and Apogee to see what to do from here. I just thought it seems a little off for a cartridge as powerful as .44.

Apogee-dev:
> It's actually a more powerful cartridge than .357, funny enough. I'm not sure if anyone wants a 35 force cartridge though.

.44 Roumain is _not_ intended to be equivalent to .44 Magnum. It's not a magnum cartridge at all, in fact. It's intended to be something like .44 Special, which was .44 Magnum's parent case - similar to the relationship between .38 Special and .357 Magnum. If we ever add an actual .44 Magnum revolver it would need to be nearly as rare as the ashhand.

generalthrax:
Well there it is

Martinpachu:
liiimbuss companyyyyyyyy

PR #3176, opened 2024-07-02
[IDB IGNORE][MDB IGNORE] Reorganizes Guns Sprites and Repaths

rye-rice:
## About The Pull Request

This reorganizes most of the gun spritesheets into separate files, making it not fucking agony to look through

This also repaths pistols and revolvers, this is a mystery tool we will need later.

This also take the liberty of removing belt_mirror.dmi because fuck that file, all suit storage slots will pull from either mob overlay icon or back.dmi as a fallback.

## Why It's Good For The Game

Would be very helpful for the 5 million gun rewokrs that are done, but not implmented yet. While it is true i'm waiting on gun attachments, this is more pressing than that.

## Changelog

:cl:
add: Resprites the E-50 and gives it proper inhands now!
fix: The Syndicate sniper rifle fires at a normal rate of fire and has the correct recoil now.
/:cl:


Imaginos16:
LGTM

PR #3189, opened 2024-07-08
removes the atmospherics hardsuit from the cargo

Martinpachu:
## About The Pull Request

no

## Why It's Good For The Game

its faction gear!!!!!!!!!! you're not getting that
## Changelog

:cl:

del: The outpost has received a cease and desist letter originating from Nanotrasen, and no longer offer their iconic atmospherics hardsuits.
/:cl:

### DIFF
--- a/code/modules/cargo/packs/spacesuit_armor.dm
+++ b/code/modules/cargo/packs/spacesuit_armor.dm
@@ -72,14 +72,6 @@
 	crate_name = "engineering space suit crate"
 	crate_type = /obj/structure/closet/crate/secure/engineering
 
-/datum/supply_pack/spacesuit_armor/atmos_hardsuit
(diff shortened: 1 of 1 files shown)


SomeguyManperson:
Just change the description if its that dissonant

Martinpachu:
> Just change the description if its that dissonant

its supposed to be used on NT ships though. other ships are supposed to use their own hardsuits on the independant engineering spacesuits which are already on the outpost

generalthrax:
> its supposed to be used on NT ships though

image

This could easily be solved by simply adding a NT atmos hardsuit instead of removing a supposedly NT hardsuit that isn't even on NT ships that could easily be rebranded in this PR.

FalloutFalcon:
imma close this since i dont think this change makes sense as explained by the comments.

Issue #2893, opened 2024-04-12
[BUG] Social anxiety trait checks for radio me's even though it shouldnt


Hlscientist:
## Round ID:
 2329

## Testmerges:
 N/A, bug should not be caused by any active TMs

## Reproduction:
1 - Have social anxiety trait
2 - have social anxiety
3 - attempt to radio me ( ; beeps*)
4 - cry as the moment is ruined as you have social anxiety and the radio me does not go through

FalloutFalcon:
I cant reproduce this. I think its related to eye contact which is annoying to replicate on local host by myself

FalloutFalcon:
fixed in #3222

PR #3222, opened 2024-07-18
Social Anxiety quirk fixes and tweaks

Zevotech:
## About The Pull Request

Changes some things about the social anxiety quirk to make it more bearable to deal with.

- Removes the ability for it to silence you, which would eat messages players tried to send while roleplaying.
- Decreases the severity of stuttering across the board, and prevents it from removing characters in the message.
- Psicodine now actually suppresses the main effects of social anxiety while it is active, making the medicine actually usable once more.

## Why It's Good For The Game

Players were unable to use this quirk due to a variety of reasons. It would randomly silence you around people, give you an exaggerated stutter that made your runechat impossible to read, and couldn't even be suppressed because it's treatment medication was bugged and had very little use outside of it.

Now the quirk is a little more managable, and players should be able to take it and roleplay with it again. I may mess with the stutter further if needed as I still think its a little extreme in some cases.

## Changelog

:cl:
balance: the Quirk Social Anxiety no longer silences you randomly
fix: Stuttering is now far less extreme and will not delete characters
fix: Psicodine now actually suppresses the effects of social anxiety
/:cl:

### DIFF
--- a/code/datums/traits/negative.dm
+++ b/code/datums/traits/negative.dm
@@ -451,15 +451,14 @@
 
 /datum/quirk/social_anxiety/on_process()
 	var/nearby_people = 0
+	if(HAS_TRAIT(quirk_holder, TRAIT_FEARLESS))
+		return
 	for(var/mob/living/carbon/human/H in oview(3, quirk_holder))
 		if(H.client)
 			nearby_people++
 	var/mob/living/carbon/human/H = quirk_holder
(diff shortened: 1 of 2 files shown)


rye-rice (approved):
Still don't like something that can be roleplayed out be w mechanic, but at least it no longer eats your messages

FalloutFalcon:
#2893
fixed this

PR #3225, opened 2024-07-18
Sandbags can be bought at the outpost and deconstructed by click dragging them onto yourself

Gristlebee:
## About The Pull Request

Adds empty sandbags to the outpost, 7 sandbags for 150 under the sec supply tab
Sandbags can be deconned by click dragging them onto your mob and a short do after. Adds an examine message for how to decon.

## Why It's Good For The Game

Sandbags are odious to use currently, being effectively single use once deployed and just sitting there once the fight's done. If you want to clean them up afterwards you're stuck smashing them apart until they break These changes should make sandbags more accessible and practical to use. 

## Changelog

:cl:
add: Sandbags can be deconed by click dragging them onto your mob.
add: Sandbags to the outpost for 150 credits.
/:cl:

### DIFF
--- a/code/game/machinery/deployable.dm
+++ b/code/game/machinery/deployable.dm
@@ -111,6 +111,24 @@
 	smoothing_groups = list(SMOOTH_GROUP_SANDBAGS)
 	canSmoothWith = list(SMOOTH_GROUP_SANDBAGS, SMOOTH_GROUP_WALLS, SMOOTH_GROUP_SECURITY_BARRICADE)
 
+/obj/structure/barricade/sandbags/MouseDrop(over_object, src_location, over_location)
+	. = ..()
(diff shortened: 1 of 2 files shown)


Apogee-dev:
I still think we could also use a folding armor plate barricade for the combat/sec ships like I talked about last time this came up, personally, but giving sandbags a long-ish teardown is fine and a nice improvement

Gristlebee:
> I still think we could also use a folding armor plate barricade for the combat/sec ships like I talked about last time this came up, personally, but giving sandbags a long-ish teardown is fine and a nice improvement

I can make it longer if it's needed, but I think it's fine as is at the moment. You're not going to be ripping apart sandbags mid-firefight, so it's usually going to be after the battle is said and done and you're cleaning up, which would make it unnecessary added waiting time I feel.

PR #3240, opened 2024-07-24
Autolathable stripper clips and 7.62 polymer rifle.

Martinpachu:
## About The Pull Request
Requested by Apogee.
Makes all stripper clips able to be created at the autolathe, since they're supposed to be simple, cheap, disposable things, as opposed to magazines which are more complex and important. They don't get built with ammo, of course. And the polymer rifle gets changed to being chambered in 7.62, which has the same stats as .300 blackout. They've been done in the same PR since it'll add the polymer rifle's 762 speedloader to the autolathe too.

## Why It's Good For The Game
Clips are supposed to be disposable cheap things, when right now they're equal to speedloaders and magazines. The polymer rifle having a special bullet type all to itself that you could get only by killing hermits was a bit befuddling.

## Changelog

:cl:
add: Clips of all kinds (not magazines!) can now be produced on the autolathe. 
del: .300 blackout. The polymer rifle now fires 7.62.
del: Clip cargo packs, as they are now redundant.
/:cl:


rye-rice (approved):
Tempted to resprite this thing now, heh

PR #3248, opened 2024-07-25
Moves everything from autolathe contraband. Removes a couple items from autolathe in general.

Sadhorizon:
## About The Pull Request
Everything you needed to hack for before, you no longer have to.
Also axes these items completely because:
- foilhat - meme item
- flamethrower - printable gun, probably not good
- electropack - Did anyone like, use it, ever? It's very gimmicky on station servers too.
- .357 casing - It was locked behind emag, and emags died. Plus, ammo in the lathe is usually surplus version.
- slime scanner - slimes died
- toy armblade - clings aren't that common in the setting, iirc

## Why It's Good For The Game
Hacking as a mechanic is very very dumb.

## Changelog

:cl:
tweak: Moved everything out of autolathe contraband - you no longer have to hack it, ever.
del: Removed foilhat, flamethrower, electropack, .357 casing and slime scanner from the autolathe.
/:cl:

### DIFF
--- a/code/modules/research/designs/autolathe_designs.dm
+++ b/code/modules/research/designs/autolathe_designs.dm
@@ -286,7 +286,7 @@
 	build_type = AUTOLATHE | PROTOLATHE
 	materials = list(/datum/material/iron = 750)
 	build_path = /obj/item/tank/internals/emergency_oxygen/engi/empty
-	category = list("hacked","Misc","Equipment")
+	category = list("initial", "Misc","Equipment")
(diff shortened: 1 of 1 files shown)


Martinpachu:
cool pr!

Anticept:
Electropacks were a sort of holdover from LONG LONG ago. They were used as part of xenoarcheology for certain checks. They could also be emagged and would be lethal to whoever you put them on.

Outside of that, could be used to keep a prisoner from acting up as I don't think a person can remove them, it requires another person? Still, doesn't seem like a lathe item.

If it's reasonable within scope, GPS should be put on autolathes.

Sadhorizon:
Out of scope.

PR #3251, opened 2024-07-28
makes the industrial jacket warm

generalthrax:
## About The Pull Request

Gives the industrial jacket cold protection

## Why It's Good For The Game

It looks pretty comfy and warm but isn't. It's kind of Strange

## Changelog

:cl:
add: Industrial jacket now keeps you warm
/:cl:

### DIFF
--- a/code/modules/clothing/suits/jobs.dm
+++ b/code/modules/clothing/suits/jobs.dm
@@ -114,6 +114,7 @@
 	item_state = "highvis"
 	blood_overlay_type = "coat"
 	body_parts_covered = CHEST|ARMS
+	cold_protection = CHEST|GROIN|ARMS
 	togglename = "zipper"
 	allowed = list(/obj/item/flashlight, /obj/item/tank/internals/emergency_oxygen, /obj/item/tank/internals/plasmaman, /obj/item/t_scanner, /obj/item/radio)
 	resistance_flags = NONE


Apogee-dev:
Are any of the other jackets available via loadouts warm? The things that I know add cold protection off the top of my head are generally jackets for officer roles and winter coats, which are parkas.

thgvr:
hoodies should be if they aren't

generalthrax:
> Are any of the other jackets available via loadouts warm? The things that I know add cold protection off the top of my head are generally jackets for officer roles and winter coats, which are parkas.

Hoodies (I Think). Leather duster. Leather overcoat. The bomber jacket (of which this is Also a bomber jacket). Several others which I'm too lazy to check as of Right Now

Apogee-dev:
> > Are any of the other jackets available via loadouts warm? The things that I know add cold protection off the top of my head are generally jackets for officer roles and winter coats, which are parkas.
> 
> Hoodies (I Think). Leather duster. Leather overcoat. The bomber jacket (of which this is Also a bomber jacket). Several others which I'm too lazy to check as of Right Now

Fair enough, then. Though the idea of a simple hoodie protecting someone from deep subzero temperatures on a frozen planet is a bit silly. Wish we could handle this kind of hazard with different degrees of protection.

Anticept:
Only winter coats are warm enough for Kepori from loadouts. Jackets, dusters, etc aren't enough. I never did look into why but this might explain it.

That said, it also makes sense that winter coats would be just about the only thing warm enough for them.

PR #3280, opened 2024-08-05
Fixing batteries and adding a medium category

molnizz:
## About The Pull Request
Removes the ability to insert batteries of any size. Also adds a new category of items
and changes them everywhere they are mentioned.

How it was:
ezgif com-video-to-gif-converter

## Why It's Good For The Game

This fixes the battery issue and uses a previously unused define

## Changelog

:cl:
balance: Maybe this will make laser weapons worse because you can't fit the BIGGEST battery in them.
fix: Adds medium batteries
/:cl:


FalloutFalcon:
I'm already working on cleaning up the logic for mags. I honestly don't think we need a separate sub type here. I'm changing the allowed mag logic to not include sub types which fixes this.

FalloutFalcon:
#3429 will fix this.. i know its not a finished pr but I massively prefer its solution over this one.

PR #3288, opened 2024-08-11
Fixes "The Food" - Allows spaghetti to be processed.

Aquidu:
Literally four characters. Technically second PR but my first fix.

## About The Pull Request
Fixes an error where processing a dough slice in a processor or on a cutting board would make you get the parent spaghetti item, rather than the intended one.

The Creation
Screenshot (2912)
The Reaction
Screenshot (2913)

## Why It's Good For The Game
As funny as "The Food" is, I would like to be able to cook spaghetti. And I don't mean code spaghetti.

## Changelog

:cl:Aquidu
fix: Fixed a bug relating to literal spaghetti code.
/:cl:


goober3:
we finally found it. the spaghetti code.

PR #3292, opened 2024-08-14
Wormholes no longer decay when used

SomeguyManperson:
## About The Pull Request

Wormholes are now permanent if one spawns on the overmap, rather than being usable a few times then deleting themselves. Ideally this will make them slightly more interactable in the event they help transit enough

## Why It's Good For The Game

I don't feel like wormholes being usable maybe up to 5 times makes them particularly more interesting than not, since there's no reason to check what is on the other side if you won't be able to use it for more than 2 round trips in the best case

## Changelog

:cl:
balance: wormholes on the overmap no longer decay when used
/:cl:

### DIFF
--- a/code/modules/overmap/objects/event_datum.dm
+++ b/code/modules/overmap/objects/event_datum.dm
@@ -194,54 +194,24 @@
 	chance_to_affect = 100
 	///The currently linked wormhole
 	var/datum/overmap/event/wormhole/other_wormhole
-	///Amount of times a ship can pass through before it collapses
-	var/stability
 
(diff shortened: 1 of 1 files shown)


Aquidu:
No hate to you, but I actually really like how wormholes are currently. Since there is an inherent risk(ship damage) but the reward is minimal, there isn't really an incentive to use wormholes, unless you're _escaping from another ship._ In that scenario, having the wormholes decay is _extremely_ useful as it allows you a nearly-guranteed escape, with good luck.

thgvr:
Let's try it out, they're rare enough

PR #3294, opened 2024-08-14
Grille defib will be real in 2024

SomeguyManperson:
## About The Pull Request

Repurposes old and useless IPC grille defib code to work with anyone. This is a 25% chance on shocking a viable body to revive them. Currently there's no present downside since the only thing I can think of would be brain damage which sucks I think. It should have one though.

## Why It's Good For The Game

image

## Changelog

:cl:
add: you can now defibrillate people by administering a controlled shock, if you get lucky.
/:cl:


Anticept:
I believe this also does heart damage on other codebases and scales with the power grid. On top of the shock damage.

thgvr:
Kind of meh

PR #3303, opened 2024-08-21
Removes restrictions on renaming guns (You can rename all of them now)

SomeguyManperson:
## About The Pull Request

This take may be steaming but I need to get it on the table
Adds the unique rename flag to all guns (removing unnecessary re-applications on guns that currently have it), allowing any gun to be renamed.
Concerns with this would be people masking the lethality of their weapons, but I think the sprites are unique and visible enough that a gun is identified by appearance rather than examine/chat text

If any gun absolutely should not be renamable for flavor purposes (trabuco or stuff like it) it can have the flag removed

## Why It's Good For The Game

Allows players to add more flavor to their equipment

## Changelog

:cl:
balance: You can now rename any and all guns.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -7,6 +7,7 @@
 	lefthand_file = GUN_LEFTHAND_ICON
 	righthand_file = GUN_RIGHTHAND_ICON
 	flags_1 =  CONDUCT_1
+	obj_flags = UNIQUE_RENAME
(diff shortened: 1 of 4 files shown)


thgvr:
I can't say I'm enthusiastic about this

goober3:
maybe it would be better to have your name appear in quotations after the actual name of the weapon, to avoid confusion.

PR #3312, opened 2024-08-26
Gun Safety PSA - Keep your gun safeties on to prevent accidental discharges!

Gristlebee:
## About The Pull Request

Unsafetied guns have a small chance to go off and fire on their own under select conditions, which are currently:
- Being thrown and hitting something.
- Being on a ship when it takes off or docks. (Having it stored in a stable container eg. lockers/crates, protect it from going off)
- Falling down while carrying a gun (Shoves, slips, etc)
- Paralyzes while holding a gun
  - Having your gun holstered lowers the chance of it going off from falling down due to the aforementioned causes
- Being EMPed if it's an energy weapon.

Modifies fire_casing so it works without a user.

## Why It's Good For The Game

Makes guns a bit more dangerous (to yourself mainly), and reinforces the fact guns are Dangerous Weapons and you should be careful with them by adding a mechanical reason to have your safety on outside of combat and properly securing your weapons.

~~adds more medical gameplay due to increased injuries~~

## Changelog

:cl:
add: Unsaftied guns can go off on their own.
code: Changes ammo casing firing code so it's able to work without a user.
/:cl:


Orchidthederg:
I thought this was already a thing! Overdue tbh. Very nice.

Martinpachu:
how does this interact with storage like bags and guncases?

Gristlebee:
> how does this interact with storage like bags and guncases?

If it's loaded, and unsafe. It can still go off and hit someone

Gristlebee:
> > how does this interact with storage like bags and guncases?
> 
> If it's loaded, and unsafe. It can still go off and hit someone

hm. just checked it actually doesnt go off if it's in a storage and not on a person. Thats probably easy enough to fix

Anticept:
While they are dangerous weapons, guns do not readily go off even if you drop them. Usually this is because of poor care. I like the change, but if its stored like in a holster, it should be less likely to occur. Someone hitting the ground with it in their hand might accidentally hit the trigger and usually THAT is what causes negligent discharges.

Formerxboxgamer:
This is a fun idea, though I don’t know if it makes sense for holstered lasers to go off when they’re jostled from a ship moving. Though this never really happens - ballistic weapons could plausibly be explained as going off from sudden movement if the firing pin is dislodged and hits the bullet, but lasers… I dunno. Any weapon in someone’s hand could go off when falling down, since most negligent discharges in life happen when the gun is in someone’s hand and their finger is where it’s not supposed to be.

rye-rice:
Needs a TM, amazing pr I've been wanting this for so long

code/modules/projectiles/gun.dm
```diff
@@ -945,6 +950,45 @@
 		flash_loc.vis_contents -= muzzle_flash
 	muzzle_flash.applied = FALSE
 
+// for guns firing on their own without a user
+/obj/item/gun/proc/discharge(cause, seek_chance = 50)
+	var/target
+	if(!safety)
+		// someone is very unlucky and about to be shot
+		if(prob(seek_chance))
```
thgvr, line 958: This needs to be VERY VERY VERY low or not exist at all.

Anticept:
> This is a fun idea, though I don’t know if it makes sense for holstered lasers to go off when they’re jostled from a ship moving. Though this never really happens - ballistic weapons could plausibly be explained as going off from sudden movement if the firing pin is dislodged and hits the bullet, but lasers… I dunno. Any weapon in someone’s hand could go off when falling down, since most negligent discharges in life happen when the gun is in someone’s hand and their finger is where it’s not supposed to be.

Firing pins are spring loaded with a pretty powerful spring and take a pretty considerable force to even unseat them... like the hammer. Not all of them, but its really uncommon. A really crappy built improvised weapon, old design, or modded one might be more prone but usually that would be more issues with the bolt carrier jamming due to the precision needed.

Modern safeties such as the drop safety mechanism aside, really the most prone are long arms. 

STILL, fun PR.

code/modules/projectiles/gun.dm
```diff
@@ -945,6 +950,45 @@
...
+	for(var/obj/item/gun/at_risk in get_all_contents())
+		var/chance_to_fire = GUN_NO_SAFETY_MALFUNCTION_CHANCE_HIGH
+		if(ishuman(src))
+			var/mob/living/carbon/human/holder = src
+			// the gun is slightly more secure in a holster
+			if(at_risk == holder.s_store)
+				chance_to_fire = GUN_NO_SAFETY_MALFUNCTION_CHANCE_MEDIUM
+		if(at_risk.safety == FALSE && prob(chance_to_fire))
+			if(at_risk.process_fire(src,src,FALSE, null,  pick(BODY_ZONE_L_LEG,BODY_ZONE_R_LEG)) == TRUE)
+				visible_message(span_danger("\The [at_risk.name]'s trigger gets caught as [src] falls, suddenly going off into [src]'s leg! Should have had the safeties on."), span_danger("\The [at_risk.name]'s trigger gets caught on something as you fall, suddenly going off into your leg without it's safeties on!"))
```
meemofcourse:
```suggestion
				visible_message(span_danger("\The [at_risk.name]'s trigger gets caught as [src] falls, suddenly going off into [src]'s leg!"), span_danger("\The [at_risk.name]'s trigger gets caught on something as you fall, suddenly going off into your leg!"))
```
i understand the thought but maybe its, like, too on the nose.

Gristlebee: it is a bit on the nose, but I think it's important to have feedback why you got shot/punished so you can correct. I dont think it feels great to not know why a punishment mechanic happened.

code/modules/projectiles/gun.dm
Gristlebee: base chance has been reduced to 10%, and on throw now has been reduced to a 50% chance, as a coin flip hail mary option to throw your gun as a last resort

thgvr: Still think it's too much honestly, should be pretty rare and absolutely not reliable

Gristlebee: 5 and 20 sound okay for base and throw respectively?

code/__DEFINES/guns.dm
```diff
@@ -69,6 +69,11 @@
 #define MANUFACTURER_PGF "the Etherbor Industries emblem"
 #define MANUFACTURER_IMPORT "Lanchester Import Co."
 
+// Misfire chances if the gun's safety is off
+#define GUN_NO_SAFETY_MALFUNCTION_CHANCE_LOW 5
+#define GUN_NO_SAFETY_MALFUNCTION_CHANCE_MEDIUM 40
+#define GUN_NO_SAFETY_MALFUNCTION_CHANCE_HIGH 80
```
tmtmtl30: i feel like these probabilities should be lowered SUBSTANTIALLY. remember that the character is (at least conceptually) "trying" to make sure the gun DOESN'T fire when they trip or stumble. additionally, guns are designed *not* to go off randomly when they're jostled. i think you could make these probabilities a lot lower -- 5, 10, 15, maybe -- and it would still be a memorable mechanic that would keep people on their toes about safeties.

code/modules/shuttle/on_move.dm
```diff
@@ -299,6 +300,17 @@ All ShuttleMove procs go here
 	if (oldT && !is_reserved_level(oldT))
 		unlocked = TRUE
 
+/obj/item/gun/lateShuttleMove(turf/oldT, list/movement_force, move_dir)
```
tmtmtl30, line 303: i'm not super comfortable with firing these procs directly from shuttle movement, but at least for the items themselves, there's not any other easy way. i would still lower the chance down from HIGH (even given the percentage chance changes i mentioned before) -- it should be pretty uncommon for guns to fire JUST because they were unsafetied during takeoff; it also strikes me as somewhat unnecessary to search all items' contents just for guns that might go off. it feels like an anti-cheese measure, but i'd argue that this is a flavor feature, not a balance one; i don't expect people to TRY to cheese it.

code/modules/projectiles/gun.dm
```diff
@@ -945,6 +950,45 @@
...
+/obj/item/gun/proc/unsafe_shot(target)
+	if(chambered)
+		chambered.fire_casing(target,null, null, null, suppressed, ran_zone(BODY_ZONE_CHEST, 50), 0, src)
+		playsound(src, fire_sound, 100, TRUE)
+
+/mob/living/proc/trip_with_gun()
+	for(var/obj/item/gun/at_risk in get_all_contents())
+		var/chance_to_fire = GUN_NO_SAFETY_MALFUNCTION_CHANCE_HIGH
+		if(ishuman(src))
+			var/mob/living/carbon/human/holder = src
```
tmtmtl30: i'd move this check outside the for() loop, instead of inside of it; just cache the result or something. istype()s can be a little slow at scale. not a big deal though

code/modules/shuttle/on_move.dm
```diff
@@ -325,6 +337,7 @@ All ShuttleMove procs go here
 	var/knockdown = movement_force["KNOCKDOWN"]
 	if(knockdown)
 		Paralyze(knockdown)
+		trip_with_gun()
```
tmtmtl30: for humans, directly adding misfire code to shuttle movement isn't something i'm comfortable with. couldn't it be a result of knockdown / paralysis generally instead of caused, specifically, by slipping and shuttle movement? in addition to resulting in cleaner code, i think it would feel more consistent.

that should be done by adding the trip_with_gun() call to /datum/status_effect/incapacitating/knockdown/on_apply() and /datum/status_effect/incapacitating/paralyzed/on_apply(), respectively (in case of paralysis, it should come BEFORE the traits are added, as TRAIT_HANDS_BLOCKED will cause items to drop. i considered pushing it even further down, to TRAIT_FLOORED, but that wouldn't be right, as TRAIT_FLOORED is also used when, for example, buckling people to beds)

tmtmtl30 (changes requested):

Gristlebee:
This still needs to cook some more, so Im setting this to draft

code/__DEFINES/guns.dm
Gristlebee: Chance for the misfires have been lowered to the values you suggested.

code/modules/shuttle/on_move.dm
Gristlebee: I mainly had it there for consistency, but Im happy to cut it out. Content check has been removed.

code/modules/projectiles/gun.dm
Gristlebee: Check's been moved outside

code/modules/shuttle/on_move.dm
Gristlebee: Proc has been moved out and now triggers on knockdown and paralyzed like you suggested.

code/modules/projectiles/gun.dm
Gristlebee: Chance to seek is just a flat 10% now, and should be seen even less with the reduced misfire chances now.

Gristlebee:
Requested changes have been implemented, the bugs that Im aware of have been ironed out. This should be good for review again unless Im stupid and missed something

code/modules/projectiles/gun.dm
```diff
@@ -977,16 +977,18 @@
 		playsound(src, fire_sound, 100, TRUE)
 
 /mob/living/proc/trip_with_gun()
+	var/chance_to_fire = GUN_NO_SAFETY_MALFUNCTION_CHANCE_HIGH
+	var/mob/living/carbon/human/human_holder
+	if(ishuman(src))
+		holder = src
```
tmtmtl30: 2 things:
1. this should be "human_holder = src", not "holder = src"
2. assigning a reference to a new variable that is typed is, typically, free. that can still sit inside the loop, not that it makes a huge difference here since you're using the null value to represent "they weren't a human"

code/modules/projectiles/gun.dm
```diff
@@ -977,16 +977,18 @@
...
-			// the gun is slightly more secure in a holster
-			if(at_risk == holder.s_store)
+		if(human_holder)
+			// gun is less likely to go off in a holster
+			if(at_risk == human_holder.s_store)
 				chance_to_fire = GUN_NO_SAFETY_MALFUNCTION_CHANCE_MEDIUM
 		if(at_risk.safety == FALSE && prob(chance_to_fire))
 			if(at_risk.process_fire(src,src,FALSE, null,  pick(BODY_ZONE_L_LEG,BODY_ZONE_R_LEG)) == TRUE)
-				visible_message(span_danger("\The [at_risk.name]'s trigger gets caught as [src] falls, suddenly going off into [src]'s leg! Should have had the safeties on."), span_danger("\The [at_risk.name]'s trigger gets caught on something as you fall, suddenly going off into your leg without it's safeties on!"))
+				visible_message(span_danger("\The [at_risk.name]'s trigger gets caught as [src] falls, suddenly going off into [src]'s leg without it's safties on!"), span_danger("\The [at_risk.name]'s trigger gets caught on something as you fall, suddenly going off into your leg without it's safeties on!"))
```
tmtmtl30: should be "its safeties", grammatically

code/modules/projectiles/gun.dm
Gristlebee: whoops, my bad. Thanks for the catch

code/modules/projectiles/gun.dm
Gristlebee: fixed

tmtmtl30 (approved):

Apogee-dev:
> > This is a fun idea, though I don’t know if it makes sense for holstered lasers to go off when they’re jostled from a ship moving. Though this never really happens - ballistic weapons could plausibly be explained as going off from sudden movement if the firing pin is dislodged and hits the bullet, but lasers… I dunno. Any weapon in someone’s hand could go off when falling down, since most negligent discharges in life happen when the gun is in someone’s hand and their finger is where it’s not supposed to be.
> 
> Firing pins are spring loaded with a pretty powerful spring and take a pretty considerable force to even unseat them... like the hammer. Not all of them, but its really uncommon. A really crappy built improvised weapon, old design, or modded one might be more prone but usually that would be more issues with the bolt carrier jamming due to the precision needed.
> 
> Modern safeties such as the drop safety mechanism aside, really the most prone are long arms.
> 
> STILL, fun PR.

On this note, we DO have quite a number of open bolt SMGs that would reasonably have a dramatically higher chance to misfire than other arms

Zevotech:
Crawling over a slippery surface (motor oil, mainly) can still cause a gun to misfire.

Leon9621:
> Crawling over a slippery surface (motor oil, mainly) can still cause a gun to misfire.

In this same vein, someone accidentally tackled a table, got stunned, and had a safety's commisioner go off into their leg.

Gristlebee:
> Crawling over a slippery surface (motor oil, mainly) can still cause a gun to misfire.

That was one fixed, and was from earlier version of the PR.

> In this same vein, someone accidentally tackled a table, got stunned, and had a safety's commisioner go off into their leg.

So with tackle code, it knocks you down (which rolls the chance for a misfire) then if you hit a table with items on it, it applies paralyze to you (which can also roll misfire). That's intended behavior, though it triggering with the safety on isnt. I wasn't able to recreate the bug though with a commisioner with the safety on and also tackling a table, so this also might be a case of a bug from the earlier version.

fighterslam:
image
Misfires aren't logged very well. It just shows if someone threw an item, and that they got hit by a projectile - but if this was several different players, it would be difficult and unclear to try to track this down.

code/__DEFINES/guns.dm
```diff
@@ -69,6 +69,11 @@
 #define MANUFACTURER_PGF "the Etherbor Industries emblem"
 #define MANUFACTURER_IMPORT "Lanchester Import Co."
 
+// Misfire chances if the gun's safety is off
+#define GUN_NO_SAFETY_MALFUNCTION_CHANCE_LOW 100
```
SomeguyManperson, line 73: What happened here

PR #3319, opened 2024-08-29
Removes insulation from captain and combat gloves, hardsuits

Apogee-dev:
## About The Pull Request

Removes electrical insulation from all gloves except for insulated gloves and their children (including captain, tackler, and combat gloves) and nixes the siemens coefficient on hardsuits that aren't engineer hardsuits of some form. Adds siemens coeff to engineering space suits. Replaces combat gloves with insulated gloves in outfit datums and locker contents for engineers.

## Why It's Good For The Game

shock protection is a bit too widely available with how many jobs spawn with combats and how many captain players are in an average round. Since electrical hazards should become a bit more significant in the future due to planned mapping hazards and potentially electrical weapons, this generally reduces it to the players directly responsible for engineering in some capacity. We shouldn’t have ships where the entire crew is shockproof at roundstart- if you want to do that it should require buying some crates, at least.

## Changelog

:cl:
balance: Removed electrical insulation from all gloves except insulated gloves
balance: Removed siemens coefficient from hardsuits, except for engi atmos and CE
balance: Added siemens coefficient to engineer space suits
balance: Replaced combat gloves with insulated gloves for some factions' engineer lockers
/:cl:

### DIFF
--- a/_maps/shuttles/inteq/inteq_valor.dmm
+++ b/_maps/shuttles/inteq/inteq_valor.dmm
@@ -539,11 +539,11 @@
 	},
 /obj/effect/turf_decal/borderfloorwhite,
 /obj/machinery/button/door{
+	dir = 1;
 	id = "valor_surgery";
 	name = "Privacy Shutters";
-	pixel_y = 8;
-	dir = 1;
-	pixel_x = 8
+	pixel_x = 8;
+	pixel_y = 8
 	},
 /turf/open/floor/plasteel/white,
 /area/ship/medical)
@@ -640,10 +640,10 @@
 /area/ship/hallway/port)
 "fO" = (
 /obj/structure/closet/secure_closet{
+	icon_door = "prisoner";
 	icon_state = "prisoner";
 	name = "shuttle pilot locker";
-	req_access = list(3);
-	icon_door = "prisoner"
+	req_access = list(3)
 	},
 /obj/item/clothing/head/beret/sec/inteq,
 /obj/item/clothing/head/soft/inteq,
@@ -1951,6 +1951,7 @@
 	},
 /obj/effect/turf_decal/corner_techfloor_gray/full,
 /obj/machinery/cell_charger,
+/obj/item/clothing/gloves/color/yellow,
 /turf/open/floor/plating,
 /area/ship/maintenance/port)
 "sc" = (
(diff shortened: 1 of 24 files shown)


FalloutFalcon (commented):
dunno whats goin on with the varibles getting reorginized in those DMMs.
I wish hardsuits could keep some siemens_coefficent but it currently functions like a binary so shrug.
Also modsuits have protection still.

Apogee-dev:
> dunno whats goin on with the varibles getting reorginized in those DMMs. I wish hardsuits could keep some siemens_coefficent but it currently functions like a binary so shrug. Also modsuits have protection still.

_do_ modsuits have protection? Because I searched for everything with a siemens_coefficient variable of 0 when I was doing this. do they use an alternate method or something. Also, does siemens_coefficient actually function like a binary? Most gloves have a coefficient of 0.5 and latex has 0.3, which seems like it would indicate otherwise

Anticept:
A suggestion I might have as an alternative: shock thresholds. Gloves would protect against low power, but at extreme power even hardsuits would leak through.

Apogee-dev:
> A suggestion I might have as an alternative: shock thresholds. Gloves would protect against low power, but at extreme power even hardsuits would leak through.

This is an idea I've thought about before, where high voltage would even bypass proper insulated gloves, but it'd be significantly more work. And a bit beside the point of this particular pr, since i mostly just wanted to de-insulate combat and captain gloves. Maybe later.

FalloutFalcon:
> > dunno whats goin on with the varibles getting reorginized in those DMMs. I wish hardsuits could keep some siemens_coefficent but it currently functions like a binary so shrug. Also modsuits have protection still.
> 
> _do_ modsuits have protection? Because I searched for everything with a siemens_coefficient variable of 0 when I was doing this. do they use an alternate method or something. Also, does siemens_coefficient actually function like a binary? Most gloves have a coefficient of 0.5 and latex has 0.3, which seems like it would indicate otherwise

code/modules/mod/mod_theme.dm, Line 31 and 317

the damage is multiplied by the coeff 
but mob siesmens is also calculated so awfully
it checks if gloves and a few other things have any protection and subtracts it by a hardcoded amount
IF its a tesla shock, otherwise it just multiply only by the glove
but if they can be shocked, the only part that currently matters, is a complete binary.
cables and grilles and airlocks all use slightly different procs but amount to ```if(gloves.siemens > 0)```

sigh....

oh and also a very large amount of shocks, just skip gloves, no protection frick you.

PR #3094, opened 2024-06-10
tweaks to recoil and spread

FalloutFalcon:
## About The Pull Request

gunslingers work now, min recoil for ballistic, much cleaner and clearer code
gunslingers and poor aim can SOMETIMES hit a clean shot

thgvr edit: cleaned up changelog.
## Changelog

:cl:
balance: Ballistic weapons now have a minimum camera shake.
fix: Gunslinger now functions as intended.
/:cl:

PR #3330, opened 2024-09-05
minimum recoil for ballistics

FalloutFalcon:
## About The Pull Request
gives a very small kick to all ballistics, not enough to affect your aim but makes them feel alot better to use.
atomization of #3094
currently debating what the value should be..

0
https://github.com/user-attachments/assets/1630a089-bf9f-4ec5-8853-c2728f420f4e

.1
https://github.com/user-attachments/assets/2064b49b-37f5-4860-b273-1f28274724f7

.175
https://github.com/user-attachments/assets/6baba2ad-8100-403e-b6c4-d4e958ad21d6

.25
https://github.com/user-attachments/assets/ca3bdf37-bf30-493e-bb24-ae452cbc6481

a bit of an extreme case being a fast firing auto but it happens to be a gun with a recoil of 0 rn so it will be affected by the change

## Why It's Good For The Game
without it, a surprising amount of ballistics get 0 recoil just but wielding them which feels really jarring.

## Changelog

:cl:
add: Ballistics now have a minimum recoil, not enough to mess up your shot!
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -316,6 +316,8 @@
 	///This prevents gun from firing until the coodown is done, affected by lag
 	var/current_cooldown = 0
 
+	var/min_recoil = 0
+
 	var/gunslinger_recoil_bonus = 0
 	var/gunslinger_spread_bonus = 0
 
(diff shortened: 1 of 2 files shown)


rye-rice:
im still very against this, but it was suggested to reduce unwielded recoil by 1 to compensate, could be useful to rebalance unwielded recoil sometime as most the values were taken from tgmc bare?

PR #3334, opened 2024-09-06
Nerfs hollowpoint rounds to bring them under limb crit threshold

Apogee-dev:
## About The Pull Request

Nerfs most hollowpoint rounds by 5 damage. Nerfs 8mm hollowpoint by 6 damage for a total of 49 damage.

## Why It's Good For The Game

This is mostly because .357 Hollowpoint does exactly 50 damage, which crits an unarmored limb (which is most players' limbs) in a single shot. The nerf doesn't hit HP so hard it isn't worth it, because you're still doing the same damage as weapons much higher up on the totem pole, but it ensures the only weapon that can crit a limb in a single shot is a .45-70 firing hollowpoint or explosive ammo. And sniper rifles like the Scout, if those ever get something like hollowpoint.

## Changelog

:cl:
balance: Reduced hollowpoint damage by 5 (and 8mm Hollowpoint by 6)
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/bullets/lmg.dm
+++ b/code/modules/projectiles/projectile/bullets/lmg.dm
@@ -67,7 +67,7 @@
 
 /obj/projectile/bullet/mm712x82/hp
 	name = "7.12x82mm hollow point bullet"
-	damage = 45
+	damage = 40
 	armour_penetration = -20
 
 /obj/projectile/bullet/incendiary/mm712x82
--- a/code/modules/projectiles/projectile/bullets/pistol.dm
+++ b/code/modules/projectiles/projectile/bullets/pistol.dm
@@ -16,7 +16,7 @@
 
 /obj/projectile/bullet/c9mm/hp
 	name = "9mm hollow point bullet"
-	damage = 40
+	damage = 35
 	armour_penetration = -50
 
 /obj/projectile/bullet/incendiary/c9mm
@@ -49,7 +49,7 @@
 
 /obj/projectile/bullet/c10mm/hp
 	name = "10mm hollow point bullet"
-	damage = 45
+	damage = 40
 	armour_penetration = -50
 
 /obj/projectile/bullet/incendiary/c10mm
@@ -82,7 +82,7 @@
 
 /obj/projectile/bullet/c45/hp
 	name = ".45 hollow point bullet"
-	damage = 45
+	damage = 40
 	armour_penetration = -50
 
 /obj/projectile/bullet/incendiary/c45
--- a/code/modules/projectiles/projectile/bullets/revolver.dm
+++ b/code/modules/projectiles/projectile/bullets/revolver.dm
@@ -6,7 +6,7 @@
 
 /obj/projectile/bullet/a50AE/hp
 	name = ".50 AE hollow point bullet"
-	damage = 60
+	damage = 55
 	armour_penetration = -50
 
 // .38 (Colt Detective Special & Winchester)
@@ -110,7 +110,7 @@
 
 /obj/projectile/bullet/a357/hp
 	name = ".357 hollow point bullet"
-	damage = 50
+	damage = 45
 	armour_penetration = -50
 	ricochet_chance = 0 //mushroom on impact, no bounces
 
@@ -133,7 +133,7 @@
 
 /obj/projectile/bullet/a4570/hp
 	name = ".45-70 hollow point bullet"
-	damage = 60 //it's the pre-nerf .357 with less armor pen
+	damage = 55 //it's ALMOST the pre-nerf .357 with less armor pen
 	armour_penetration = -50
 
 /obj/projectile/bullet/a4570/explosive //for extra oof
@@ -167,5 +167,5 @@
 
 /obj/projectile/bullet/a44roum/hp
 	name = ".44 roumain bullet"
-	damage =  45
+	damage =  40
 	armour_penetration = -20
--- a/code/modules/projectiles/projectile/bullets/rifle.dm
+++ b/code/modules/projectiles/projectile/bullets/rifle.dm
@@ -16,7 +16,7 @@
 /obj/projectile/bullet/a8_50rhp
 	name = "8x50mmR bullet"
 	speed = 0.3
-	damage = 55
+	damage = 49
 	armour_penetration = 0
 
 // .300 Magnum (Smile Rifle)


generalthrax:
Is discouraging limb shots truly something to move away from, though? You're already resigning yourself to Dealing Less damage by targeting limbs due to capped limb damage, and you're avoiding The Bulk of potential damage that you would get targeting the chest / head (armoured, in most cases). You lose out on that raw damage that you'd get targetting center of mass in exchange for the tactical choice of crippling your opponent rather than killing them outright.

It's also very picturesque to see people getting their arms shot and dropping their weapon. A focal point of a lot of gunfights others and I have been in. It's Peak. Could always just worsen its AP malus, since hollows going to fairly diminished with Armoured NPCs in ruins already.

Apogee-dev:
> Is discouraging limb shots truly something to move away from, though? You're already resigning yourself to Dealing Less damage by targeting limbs due to capped limb damage, and you're avoiding The Bulk of potential damage that you would get targeting the chest / head (armoured, in most cases). You lose out on that raw damage that you'd get targetting center of mass in exchange for the tactical choice of crippling your opponent rather than killing them outright.
> 
> It's also very picturesque to see people getting their arms shot and dropping their weapon. A focal point of a lot of gunfights others and I have been in. It's Peak. Could always just worsen its AP malus, since hollows going to fairly diminished with Armoured NPCs in ruins already.

I mean, the main reason I did this is having your limb disabled with a _single shot_ is a bit much, and I don't think that should really happen outside of the most powerful weapons in the game; before the nerf multiple weapons were capable of doing this, including 8mm and .357, both of which are very accessible as of the lever action PR and one of which is capable of rapid follow-up shots anyway. Taking 49 or 45 damage to a limb means it's going to be disabled if anything else so much as sneezes on it, or if you had basically any pre-existing damage; and it gives the player on the receiving end a chance to react while still scaring them senseless.

generalthrax:
> I mean, the main reason I did this is having your limb disabled with a _single shot_ is a bit much

This is a fair reasoning. If me or someone else was to make it so being shot in the arm (by a sufficiently powerful calibre) makes them drop their weapon without a full limb disable that would require immediate treatment to pick up your weapon again, do you think this could pass?

The main draw to hollow point (for me, at least) was the turnaround you could pull by having people fumble their weapons. It'd be a shame to do away it with it entirely.

If anything, maybe tune the hollowpoint 8mm to 50 instead of 49. If it's doing so much damage that it needs to be tuned down by exactly 1 point of damage to not limb disable it's probably powerful enough to be a limb-disabling cartridge.

Apogee-dev:
> > I mean, the main reason I did this is having your limb disabled with a _single shot_ is a bit much
> 
> This is a fair reasoning. If me or someone else was to make it so being shot in the arm (by a sufficiently powerful calibre) makes them drop their weapon without a full limb disable that would require immediate treatment to pick up your weapon again, do you think this could pass?
> 
> The main draw to hollow point (for me, at least) was the turnaround you could pull by having people fumble their weapons. It'd be a shame to do away it with it entirely.

Maybe, sure, though I still think making it all but guaranteed with one hit (provided it lands) is still a bit much, even if it's just a fumble rather than a disable. Two hits to a limb is perfectly doable, especially when you have rapid fire capability- even 9mm hollow point can pull that off, after all. For the most part, this nerf is really just hitting .357 and 8mm on the disable front, since everything else still has the same breakpoints (assuming it's the only source of damage, anyway)

generalthrax:
I think it's a fair assumption to think you getting shot in the arm by Any cartridge let alone hollow point from 8x50mm to put you out of commission with that for a Good Long While. You can argue for 357 not having that same capability easily because of how fast you can fire out of every gun that shoots it but You Just Can't with the Ilestren.

Apogee-dev:
> I think it's a fair assumption to think you getting shot in the arm by Any cartridge let alone hollow point from 8x50mm to put you out of commission with that for a Good Long While. You can argue for 357 not having that same capability easily because of how fast you can fire out of every gun that shoots it but You Just Can't with the Ilestren.

I mean, we're going to add a semi auto 8mm one of these days. Hell, I could see upgrading the Vickland to use it, that thing could use some love compared to the proper DMRs since it's worse in every respect right now. Relegating that ability to .45-70 and snipers is reasonable, in my opinion. .308s wouldn't hit the breakpoint either, if .308 hollow point existed. Prior to this nerf the ability to oneshot a limb was relegated to exactly four guns when using hollow point: the illestren, the syndie 357, the beacon and the ashhand (and shotguns, technically, but you'd expect that of them). I don't really think one hit gun disarms are something that _should_ be a common feature, frankly, as cool as they feel to pull off. It's cheap for the victim, and being able to potentially one-hit disarm somebody from significant range with a very fast projectile cheapens dedicated non-lethal options as well- most of which are barely used as it is.

generalthrax:
> I mean, we're going to add a semi auto 8mm one of these days

We should keep small balance changes relevant for the present rather than something we don't have an ETA on.

>  It's cheap for the victim

This is the price people pay choosing between the different types of armour. Choices matter.

> cheapens dedicated non-lethal options as well- most of which are barely used as it is.

Non-lethals are very good, _especially_ against armoured targets. Less so cheapened and more so people Want To Kill

Apogee-dev:
> This is the price people pay choosing between the different types of armour. Choices matter.

most of the time, there _isn't_ a choice there. properly armored hardsuits are rare, and the accessible ones (sec suits) have historically been unpopular due to slowdown (here's hoping the hardsuit pr fixes that). I would very much like vests to be a bit more relevant given what a bad rap they get right now. If you still want to one-hit disarm people, use a Beacon. God knows that gun needs all the help it can get anyway, being the only common gun able to pull that off wouldn't hurt.

generalthrax:
Fair. I'll lay off, for now, but I'll still hold that I think limb shots should still matter a fair bit in combat. Up until Now they're just Statistically Worse to target than the chest or head because of capped damage. We all know what it's like to forget to toggle off of limbs and fire a full magazine into someones left leg and see they're not even in crit.

I'll experiment with limb shots giving some (temporary) malus a little bit later. It Should matter where you shoot someone and having it all the same is antithetical to engaging gunplay, imo.

generalthrax:
> If you still want to one-hit disarm people, use a Beacon

Actually, I failed to realize. .45-70 deals 45 brute, not enough to cripple a limb. Beacon's slow enough and Suffers Sufficiently as a weapon already. Probably not too out of scope to bring the force up a little bit to fill that gap.

Apogee-dev:
> > If you still want to one-hit disarm people, use a Beacon
> 
> Actually, I failed to realize. .45-70 deals 45 brute, not enough to cripple a limb. Beacon's slow enough and Suffers Sufficiently as a weapon already. Probably not too out of scope to bring it up a little bit to fill that gap.

I was referring to .45-70 Hollowpoint, not standard. Which, admittedly, isn't readily available because 60 (55 in this PR) damage is a lot to say the least. We're at the point in development that we probably _could_ put it on the black market or something. .45-70 itself does 45 damage instead of 50 because I didn't want Ashhands being able to two-tap a player without the funny ammo.

PR #3335, opened 2024-09-06
Rebalances hardsuit slowdowns

Apogee-dev:
## About The Pull Request

Rebalances slowdown factors on various hardsuits to make some better and some worse.

- Blood-red hardsuits now have a flat 0.5 slowdown at all times; toggling modes does not affect speed
- Security hardsuits and other combat hardsuits (excluding swat suits) have a slowdown of 0.5; having worse armor than blood-reds is bad enough as it is
- the heavy mining hardsuit now has the same speed as the mining hardsuit because it was faster than its lighter counterpart for some bizarre reason
- The solgov hardsuit also got nerfed to 0.5 slowdown because it was full speed for some reason
- The quixote mobility suit, despite being rare as hen's teeth, has 0.3 slowdown because thgvr doesn't want any hardsuit to be fullspeed

## Why It's Good For The Game

Blood-red hardsuits (and their rarer sibling, the elite hardsuit) were unequivocally the kings of the armor scene, boasting both very high armor values and high mobility. if you found a blood red suit, you basically never had any reason to take it off or made any kind of trade-off for it. by contrast, security hardsuits and their relatives are extremely unpopular both for their slowdown and their mediocre armor values. this PR at least makes the slowdown much more tolerable, making all combat hardsuits (not counting swat suits) the same speed as a blood-red hardsuit in EVA mode. Ramzi suits were meant to be worse than blood reds in part because the speed boost doesn't work on them; this downside is now represented by just making them the same speed as a miner's hardsuit at all times. It still represents rusted and nonfunctioning servos and overall clunk like before.

## Changelog

:cl:
balance: Made security hardsuits faster and blood-red hardsuits slower
balance: Normalized some weird hardsuit speed outliers
/:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/hardsuit.dm
+++ b/code/modules/clothing/spacesuits/hardsuit.dm
@@ -305,7 +305,6 @@
 	allowed = list(/obj/item/flashlight, /obj/item/tank/internals, /obj/item/storage/bag/ore, /obj/item/pickaxe)
 	helmettype = /obj/item/clothing/head/helmet/space/hardsuit/mining/heavy
 	custom_price = 4500
-	slowdown = 0.5
 
 /obj/item/clothing/head/helmet/space/hardsuit/mining/heavy
 	name = "heavy mining helmet"
@@ -387,13 +386,11 @@
 		if(on)
 			linkedsuit.name = initial(linkedsuit.name)
 			linkedsuit.desc = initial(linkedsuit.desc)
-			linkedsuit.slowdown = 1
 			linkedsuit.clothing_flags |= STOPSPRESSUREDAMAGE
 			linkedsuit.cold_protection |= CHEST | GROIN | LEGS | FEET | ARMS | HANDS
 		else
 			linkedsuit.name += " (combat)"
 			linkedsuit.desc = linkedsuit.alt_desc
-			linkedsuit.slowdown = linkedsuit.combat_slowdown
 			linkedsuit.clothing_flags &= ~STOPSPRESSUREDAMAGE
 			linkedsuit.cold_protection &= ~(CHEST | GROIN | LEGS | FEET | ARMS | HANDS)
 			if(linkedsuit.lightweight)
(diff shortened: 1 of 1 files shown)


generalthrax:
Good changes all around. There being a Choice you have to make between hardsuits and conventional armour for the hardliners is something I like to see, no more John Sodium Thiopental wearing the suit everywhere. Probably go down to 0.7 for the Rusted-Red, though. Currently it's their only Defining Piece Of Equipment for player pirates with already pretty hefty-slowdown for poor armour, and there's not exactly a cleanly-marked Ramzi armour-set for them to use in its place for non-EVA operations.

Apogee-dev:
> Good changes all around. There being a Choice you have to make between hardsuits and conventional armour for the hardliners is something I like to see, no more John Sodium Thiopental wearing the suit everywhere. Probably go down to 0.7 for the Rusted-Red, though. Currently it's their only Defining Piece Of Equipment for player pirates with already pretty hefty-slowdown for poor armour, and there's not exactly a cleanly-marked Ramzi armour-set for them to use in its place for non-EVA operations.

Fair point, I actually had it at 0.7 when I was first going over things. Also took this opportunity to make it "Rust-red" because rusted-red isn't actually a color and it was BUGGING ME, it's either make it rust-red or remove the hyphen to fix the grammar. i chose the one consistent with the other gorlex suits

Yurek1:
out of curiosity, what is the point of combat mode in syndicate hardsuits if not the speed bonus? does it provide better protection? and if not, what is the point of combat mode without that speed bonus?

Apogee-dev:
> out of curiosity, what is the point of combat mode in syndicate hardsuits if not the speed bonus? does it provide better protection? and if not, what is the point of combat mode without that speed bonus?

Combat mode pops the helmet open so you can eat while wearing the suit. It also looks cool. _calling_ it combat mode isn't completely accurate anymore, but I don't want to go into changing that on what's meant to be a quick and simple balance change.

Yurek1:
> > out of curiosity, what is the point of combat mode in syndicate hardsuits if not the speed bonus? does it provide better protection? and if not, what is the point of combat mode without that speed bonus?
> 
> Combat mode pops the helmet open so you can eat while wearing the suit. It also looks cool.

that doesn't seem like a great bonus, but i understand the intention of the PR- i just feel like it makes combat mode mostly pointless

Apogee-dev:
> that doesn't seem like a great bonus, but i understand the intention of the PR- i just feel like it makes combat mode mostly pointless

Some things can just be there for looks, you know. And if their existence must be justified more than that, I'm more likely to remove it completely than I am to reverse course on a balance change.

thgvr:
all it should ever do is flip the visor up

rye-rice:
I know you have coding bursts then disapear for a bit, but do you think you could flavor it as flipping up the visor and down? just would be a nice touch

PR #3343, opened 2024-09-08
Add tgfont namespace

ZephyrTFA:
chat icons


FalloutFalcon:
shows up
self merges own pr
refuses to elaborate
so based..

Issue #772, opened 2022-02-05
[FEATURE] Join Delay on Ship Spawn/Pre-emptive Memo/ShipName


ta3370:
## Name of feature: 
Ship Spawn Join Delay/Pre-naming/Memo-ing

### Related Pull Request on other codebase: 
N/A

### Additional Info:

It would be helpful to have some mechanic in place that made it more accessible to set Memo and Name, whether a 2 minute join delay after ship spawn, or the ability to pre-emptively set Memo and Name before it spawns. It's kind of disheartening when there's a mismatch between goals in the crew because literally over half a dozen people joined before you even got your face off the cryo floor to set the memo so they'd know what to expect of the ship and it's plans.

Sinestia:
Why not simply have spawning disabled until enabled by the captain?

Maybe an initial state that still shows the ship in the spawn list, but with the name "Pending."

Of course, a popup or an inclusion into the ship purchasing menu would be ideal. Bit more work though since I believe you'd have to write a small tgui page to replace the input list for buying a ship to do this correctly.

PR #3346, opened 2024-09-09
Ship joining starts on locked

FalloutFalcon:
## About The Pull Request
Ship's start on locked instead of open
Idea from #772

## Why It's Good For The Game
No longer can an assistant join your ship faster then you can, rejoice in time to write your memo and change your ships name.

## Changelog

:cl:
fix: ships now start closed. shiptesters be writing there memos and ship names.
/:cl:

### DIFF
--- a/code/modules/overmap/ships/controlled_ship_datum.dm
+++ b/code/modules/overmap/ships/controlled_ship_datum.dm
@@ -56,7 +56,7 @@
 	var/owner_check_timer_id
 
 	/// The ship's join mode. Controls whether players can join freely, have to apply, or can't join at all.
-	var/join_mode = SHIP_JOIN_MODE_OPEN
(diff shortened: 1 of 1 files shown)


thgvr:
Not very intuitive, how about requesting a memo when you attempt to spawn a ship

Sadhorizon:
> Not very intuitive

How about setting it to apply?
Less confusing for newer captains, but still prevents people from joining from second one.

FalloutFalcon:
> Not very intuitive, how about requesting a memo when you attempt to spawn a ship

This was so simple. I can work on this at some point but this an improvement to having people bypass your application.

Anticept:
Keep reminding players every few minutes their ship is locked until they open the ship control interface for the first time and set a memo. That should be good enough in case someone forgets or doesn't know.

PR #3348, opened 2024-09-10
Removes extra redundant var from hermit simplemobs

rye-rice:
## About The Pull Request

Removes the var as it was unusued and also redundant

## Why It's Good For The Game

Cleaner code

## Changelog
Literally nothing player facing


code/modules/mob/living/carbon/human/knpc.dm
FalloutFalcon, line 1: this entire file is a undocumented change?

PR #3353, opened 2024-09-13
Adds welding goggles to autolathes

zimon9:
## About The Pull Request

Moves welding goggles out of misc_designs.dm and into autolathe designs. They retain the PROTOLATHE flag, so protolathes should still be able to print them out.

## Why It's Good For The Game

This PR makes it so that you no longer require RnD in order to print welding goggles out.

## Changelog

:cl:
add: Added welding goggles to the autolathe
/:cl:

### DIFF
--- a/code/modules/research/designs/autolathe_designs.dm
+++ b/code/modules/research/designs/autolathe_designs.dm
@@ -146,6 +146,14 @@
 	build_path = /obj/item/clothing/head/welding
 	category = list("initial","Tools")
 
+/datum/design/welding_goggles
+	name = "Welding Goggles"
+	id = "welding_goggles"
+	build_type = AUTOLATHE | PROTOLATHE
+	materials = list(/datum/material/iron = 500, /datum/material/glass = 500)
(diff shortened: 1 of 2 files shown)


Apogee-dev:
We have welding helmets in the autolathe already, don’t we? They do the same job, they’re just less convenient. I don’t see the point in adding an item to the lathe that obsoletes another item already on the lathe- it’s redundant and supposed to be for bare necessities and simple consumables anyway.

FalloutFalcon:
add it to cargo!

zimon9:
> We have welding helmets in the autolathe already, don’t we? They do the same job, they’re just less convenient. I don’t see the point in adding an item to the lathe that obsoletes another item already on the lathe- it’s redundant and supposed to be for bare necessities and simple consumables anyway.

That's true, but I see these as an option for those that wish to have welding protection and can't use a helmet slot. In any case, it's not anything vital, and I do like the autolathe being for more bare items. I'll be closing this.

PR #3366, opened 2024-09-15
Anti-Radiation Foam

Erikafox:
## About The Pull Request
Buyable foam that destroys radiation emitters, and decontaminates objects very efficiently.

## Why It's Good For The Game
Makes wasteplanet survival less of a hassle when prepared

## Changelog

:cl: Erika Fox
add: Anti-Radiation Foam is now available at the outpost
/:cl:


thgvr:
I'm not sure if we use those invisible radiation sources anymore really, and regardless they were invisible so not sure how useful this will be

FalloutFalcon:
could easily be reworked as a general anti rad shield foam

FalloutFalcon:
just as in it still looks like it would be useful regardless of if we have those rad sources. i will properly review later, i skimmed

FalloutFalcon (approved):
Looks fine, small enough to probally merged without test merge. Unless someone wants to run a wasteplanet only gimmick or something lol

PR #3371, opened 2024-09-16
Meteor "Shield"

Erikafox:
## About The Pull Request
Adds purchasable anti-meteor turrets to the outpost.

## Why It's Good For The Game
Sometimes you gotta go fast. 

## Changelog

:cl:
add: Ships can now purchase anti-meteor grids at the outpost. Installation not included.
/:cl:


code/modules/station_goals/shield.dm
```diff
@@ -1,47 +1,123 @@
...
+				to_chat(user, span_notice("You unsecure \the [src] from the floor!"))
+				if(active)
+					to_chat(user, span_notice("\The [src] shuts off!"))
+				set_anchored(FALSE)
+	if(W.tool_behaviour == TOOL_MULTITOOL)
+		var/a = stripped_input(usr, "Please enter desired ID.", name, id, 20)
+		if (!a)
+			return
+		id = a
+		. = TRUE
```
FalloutFalcon, line 68: these should probably use tool_acts but i forget if we actually like those

FalloutFalcon (approved):
Ill probally refactor these to use a computer in the helm at some point but this is fine for now

code/modules/station_goals/shield.dm
```diff
@@ -1,47 +1,123 @@
...
+	. = ..()
+	icon_state = active ? "syndie_lethal" : "syndie_off"
 
-/datum/station_goal/station_shield/check_completion()
-	if(..())
-		return TRUE
-	if(get_coverage() >= coverage_goal)
+/obj/machinery/meteor_shield/proc/los(meteor)
+	for(var/turf/T in getline(src,meteor))
 		return TRUE
```
SomeguyManperson: This seems to do nothing rn

code/modules/station_goals/shield.dm
```diff
@@ -1,47 +1,123 @@
...
+	density = TRUE
+	use_power = FALSE
+	processing_flags = START_PROCESSING_MANUALLY
+	subsystem_type = /datum/controller/subsystem/processing/fastprocess
+	idle_power_usage = 100
+	active_power_usage = 1000
+	var/id = ""
+	var/active = TRUE
+	var/kill_range = 6
+	var/fire_delay = 7 SECONDS
```
SomeguyManperson: These seem a bit strict for something you're spending probably 5k+ on; high delay means you'd chance losing a covered area because the turret's cooling (unless you have overlap) and low range means you'll need quite a few turrets even disregarding that

code/modules/station_goals/shield.dm
```diff
@@ -51,133 +127,20 @@
...
-/obj/machinery/computer/sat_control/ui_data()
+/obj/machinery/computer/meteor_shield/ui_data()
 	var/list/data = list()
 
 	data["satellites"] = list()
-	for(var/obj/machinery/satellite/S in GLOB.machines)
+	for(var/obj/machinery/meteor_shield/S in GLOB.machines)
 		data["satellites"] += list(list(
 			"id" = S.id,
 			"active" = S.active,
```
SomeguyManperson: This might need some work to get the console to only show turrets on the same ship, otherwise you'll get a list of everyone's turrets regardless of if you can touch them & I think you can touch turrets if the ships are docked to the same location

Erikafox: I'll add them to a list on the ship. Shouldn't be too hard.

Erikafox:
heartbreaking: I am lazy

PR #3380, opened 2024-09-16
Scarab-class Heavy Mining Ship

Sadhorizon:
## About The Pull Request
New ship, hooray!
This one is kinda quirky - it's a mining ship, but without any miners. It's main focus is on the drilling mechanic, which means it has extended engineering and security.

**TODO:**
- [x] ~~fix the airlock in the back~~ for the most part, it works. It's as broken as the AACs are.
- [x] desc
- [x] get the ballistic goggles in
- [x] Do the bridge again.
- [x] Newscasters 

<p>

scrnshot1

StrongDMM-2024-10-17 16 26 54

StrongDMM-2024-10-17 16 26 56

</p>
</details> 

Main features:
- Roundstart drill.
- A Talos-style maint that can be used for engineering projects.
- Strong guns for an indie ship - 1 Indie Cobra, 1 HP Hellfire, 2 HP Detective Specials.
- Roundstart angle grinders.

Crew manifest:
- 1x Captain
- 1x Foreman
- 2x Technician
- 2x Security Detail
- 2x Deckhand

Desc revised by apogee <3

## Why It's Good For The Game
This ship gives indies three new niches at once: drilling focus, engineering focus, potential for some sec gimmicks. Also has gear for salvaging.
Also gives indies a Hardline ship at last.

## Changelog

:cl:
add: Added the Scarab-class Heavy Mining Ship
/:cl:


Apogee-dev:
Minor note, but "clipper" feels sort of unfitting for this class, mining ship that she is. Historically the word referred to merchant ships designed to go as fast as possible while carrying a decent cargo, and a big chunky rig miner is almost the polar opposite. Something like "heavy mining vessel" or "drilling ship" might do the trick instead.

Gristlebee (commented):
Very pretty ship you've made, though I have a few bits of feedback and concerns.

- I'm not entirely sold on this section here. It seems odd to me to have to pass through maintenance on the way to bridge, and it's rather dark for what's going to be a regularly trafficked area. I think a normal hallway section would be better.
image
- Unique way to store your gear, I like it. I'd suggest putting some decals underneath the cabinet, or changing it's name to something like gear cabinet as to clearly indicate there's something inside. 
image
 
- I'm not sold on this door here. I assume the intent is something similar to the loading door for the smelter on the Delta, though as far as I can tell there's no switch to open it.  There's also the fact the main airlock is there right above it, making it rather redundant. I'd prefer if replaced the you door with another wall.
image
- I'd suggest adding a captain's duster to the captain's locker for parity with other indie vessels.
- A megaphone in either the foreman or the captain's locker (or both) I feel would be a useful addition for coordinating the drills.

Sadhorizon:
> I assume the intent is something similar to the loading door for the smelter on the Delta, though as far as I can tell there's no switch to open it. There's also the fact the main airlock is there right above it, making it rather redundant. I'd prefer if replaced the you door with another wall.

It's a maintenance door in case belts get fucky wucky. Can't really deconstruct the smelter stuff rn so I figured it's the second best. Very much an edge case, but I think it makes sense lore-wise.

> I'd suggest adding a captain's duster to the captain's locker for parity with other indie vessels.

I do not want them to have a western aesthetic. They are industrial-people, not cowboys.

Gristlebee:
> It's a maintenance door in case belts get fucky wucky. Can't really deconstruct the smelter stuff rn so I figured it's the second best.
> I do not want them to have a western aesthetic. They are industrial-people, not cowboys.

That's fair enough. It can be kept as is.

Gristlebee (commented):
Feedback after testing on live:
---
Ship worked great on live, but there's a couple concerns.

The power flow console here is capable of accessing APCs on other ships which is a concern. Swap it out for a power monitor console instead.
image


Cargo orders are able to drop onto the ore processing belts. There is the maintenance door to grab it, but that still isn't great. I'd suggest giving it its own separate area from the cargo bay, though if you have any other solutions in mind, that's fine as well.

I'd suggest linking the maintenance door for the ore belts to an access locked button on the foreman's desk or the bridge. It feels a bit odd to need to hack it open in order to fix your ship.

The way the cycle airlock is set up in the rear can have it's plating ripped up due to pressure differentials from not decompressing first.

The large airlock means it's also using a lot of air to cycle. It eats through the ship's distro pretty rapidly. I'd recommend adding a regular canister of air for refilling in a pinch.

Sadhorizon:
> I'd suggest linking the maintenance door for the ore belts to an access locked button on the foreman's desk or the bridge. It feels a bit odd to need to hack it open in order to fix your ship.

Nah. Fixing that part would be too rare to warrant a button. I think you say that because your cargo orders got stuck there, which is 100% an oversight.

Rest I'll fix later.

meemofcourse (approved):

Apogee-dev:
Good to go, send it.

PR #3381, opened 2024-09-16
PGF Turrets

Erikafox:
## About The Pull Request

https://github.com/user-attachments/assets/a8715ea0-0721-4025-8a5a-74296290ae77

Defines some pgf flavored laser turrets. Maps the light variant onto the Crying Sun as a replacement for the weak turrets currently installed (shouldn't be a major upgrade).
The turrets are:
Standard - uses the pgf assault rifle laser
Light - uses the pgf pistol laser
Heavy - uses the pgf sniper laser and a heavier disabler beam.

Also organizes the file slightly and renames the faction define for solcon (solgov -> solcon)

## Why It's Good For The Game
More turret definitions is good. Also I need these for a ship.

## Changelog

:cl:
add: The PGF now has it's own turret offerings. They come in light, normal, and heavy flavor. 
/:cl:


DIB-DOG:
Quick question, what do you have the new lasers defaulted to on the Crying Sun? 

The turret control panel in the bridge of the Crying sun does not work in controlling the ship's lasers so whatever they are set to in the mapping is what they are set to permanently.

Erikafox:
I'm just gonna go ahead and fix the turret control tbh.

Also they were set to lethal in editor so don't worry. Behavior is consistent

DIB-DOG:
Not as simple as it looks. We're not allowed to make their tile "bridge" so they will connect to the controller. Which is the easiest "fix" (workaround) I know of.

Cool cool, I also took a peek at the files changed and saw the mode stayed the same so sorry for the obvious question there.

PR #3386, opened 2024-09-17
Makes Ship Turrets Connectable Without Areas

Erikafox:
## About The Pull Request
Hacks some stuff onto the code for turrets and turretid machines so that they can see turrets outside the area (thanks fallcon). Applies this to every single ship so that ship turrets are a little less weird.

## Why It's Good For The Game
Extendable + Useful for mappers. Also less area gore.

## Changelog

:cl:
code: Ship Turrets can now be attached to control consoles outside the area they are in. please look at ship maps for examples on how.
add: Every single ship with turrets now does the above
/:cl:


FalloutFalcon (approved):
The code is due for a real refactor but this is good for now!

PR #3393, opened 2024-09-20
Adds Flares To Outpost

Erikafox:
## About The Pull Request
You can now buy flares at the outpost

## Why It's Good For The Game
flares r cool

## Changelog

:cl:
add: You can now buy flares at the outpost
/:cl:

### DIFF
--- a/code/modules/cargo/packs/exploration.dm
+++ b/code/modules/cargo/packs/exploration.dm
@@ -135,8 +135,8 @@
 	desc = "Contains 4 boxes of flares (28 total)! Great for lighting things up."
 	cost = 100
 	contains = list(
-		/obj/item/storage/box/flares
-		/obj/item/storage/box/flares
-		/obj/item/storage/box/flares
-		/obj/item/storage/box/flares
+		/obj/item/storage/box/flares,
+		/obj/item/storage/box/flares,
+		/obj/item/storage/box/flares,
+		/obj/item/storage/box/flares,
 	)
--- a/code/modules/cargo/packs/exploration.dm
+++ b/code/modules/cargo/packs/exploration.dm
@@ -129,3 +129,14 @@
 	contains = list(
 		/obj/item/gps
 	)
+
+/datum/supply_pack/exploration/flares
+	name = "Flare Supply Pack"
+	desc = "Contains 4 boxes of flares (28 total)! Great for lighting things up."
+	cost = 100
+	contains = list(
+		/obj/item/storage/box/flares
+		/obj/item/storage/box/flares
+		/obj/item/storage/box/flares
+		/obj/item/storage/box/flares
+	)


FalloutFalcon (approved):
flares are cool

Martinpachu:
since theyre meant to be so gosh darn disposable id kinda rather they be on the autolathe but this is good! like it

Erikafox:
consider: 
Flares have fuel and the autolathe isn't supposed to Do That

MemeSnorfer:
To sorta agree with Martin here, flares _are_ cheap and incredibly disposable, and thus would really be better off being in the autolathe. **Basic** tools for exploration should be accessible by each individual crewmember, the same way extinguishers, flashlights, oxygen tanks, analyzers, and marker beacons are.

I fail to see how anybody is going to go asking their captain to go buy more flares than they'll ever use in a round, _every_ round. A cargo pack like this would never be bought, when a player can just print a dozen flashlights from the autolathe for the exact same effect.

Erikafox:
I will admit you're tempting me to remove bulk flashlight printing. 
I'll drop the price down to like. 50 credits when I'm not on mobile though

Anticept:
Flares were already discussed in discord and there was overwhelming support to add them to the lathe as well as the outpost. I was going to do it when the request was changed to make it a box of 10 and i haven't sat down yet to figure out how to make boxes of flares.

Erikafox:
If they're added to the lathe then they should require the addition of some sort of burnable source, just as welding tools start empty.

Anticept:
Flares are full of metal. That's is a big part of the fuel. The other part is an oxidizer.

I thought the same things but a couple of the staff said just do it. They would prefer seeing flares tossed around over flashlights and like the moodiness.

Erikafox:
sur. I'd nix the flashlight printing then.

Anticept:
Flashlights could be more expensive to print and you would largely achieve the same goal.

It sounds funny to sell flares and flashlights if they're printable sure but this won't be the case forever. And there are cases where they might be desireablr to order instead, such as if a ship lost its lathe and is trying to hunt down a new one

Erikafox:
I don't think there'd be a meaningful difference in flare / flashlight print costs unless it was somewhat unreasonable. (The current print cost for a flashlight is 20 units of glass and 50 units of metal, one sheet is 2000 units).

Orchidthederg:
Smhh Can't Believe They Remove FLash light. Can't Have Shit in the Shipped Test. 
(jokes aside, yes, please add flares to lathes)

Anticept:
> I don't think there'd be a meaningful difference in flare / flashlight print costs unless it was somewhat unreasonable. (The current print cost for a flashlight is 20 units of glass and 50 units of metal, one sheet is 2000 units).

Thay're cheap because thay're needed.

I would say up flashlights a bunch but not too crazy.1k/500 metal to glass?

I was told to make flares CHEAP. I said 500 metal per flate and they still said thats a lot. I was going to try 250 per flare but how about for the test merge just try 100 per flare in a box of 10 (1000 metal total) just to encourage people to go flare crazy to test things out?

Printable flares won't make them unneeded, rather they fill different lighting niches. Flashlights are still useful for personal sight, the main issue is littering the drillsites and caves. I also mentioned flares would be a stealth nerf to frontiersmen who love to go lights out but again the discussion was its a good balance, and i agree. Throwing around lots of flares does make sense. It would be neat during xenomorph events too.

Anyways, even if an item is available on lathes doesn't mean it's a bad idea to also have them on the outpost. The teo aren't mutually exclusive :)

Apogee-dev:
Eh, just axe flashlights from the lathe and make flares dirt cheap. Or, ideally, make them come in a box of 10 or so. You can put flashlights (and seclights) in cargo instead. It’s an electronic device anyway, it’s more complex than most of the stuff the lathe manufactures.

MemeSnorfer:
There's an _electronics tab_ in the autolathe.

The autolathe is also capable of printing a variety of things arguably more complex than a flashlight.

For example:
Circuit boards for a variety of essential machinery. (APCs, air alarms, firelocks, airlocks)
Assembly parts that can receive and output signals. (voice analyzer, remote signaller, proximity sensor, infrared sensor, health sensor, timer)
As well as a multitude of _other_ basic and/or essential devices. (handheld radio, analyzer, multitool, camera, T-ray scanner, universal recorder, geiger counter)

Erikafox:
please go talk about this on the discord. This is a pr for adding a crate of flares to the outpost and I have no interest in doing anything else in it.

PR #3408, opened 2024-09-22
Makes Vegan rations actually Vegan

zimon9:
## About The Pull Request

Exchanges the pizza crackers snack, which is a meat foodtype, with fruit puree, in the vegan chili ration. 

## Why It's Good For The Game

It always felt like a really minor and silly oversight that pizza crackers weren't actually vegan, which, in turn, made the _vegan ration_ not actually entirely vegan. This isn't really a gamebreaking change, but I thought it might help with consistency.

## Changelog

:cl:
add: Added fruit puree to vegan rations
del: Removed pizza crackers from vegan rations
/:cl:

### DIFF
--- a/code/game/objects/items/storage/ration.dm
+++ b/code/game/objects/items/storage/ration.dm
@@ -54,7 +54,7 @@
 		/obj/item/reagent_containers/food/snacks/ration/entree/vegan_chili = 1,
 		/obj/item/reagent_containers/food/snacks/ration/side/vegan_crackers = 1,
 		/obj/item/reagent_containers/food/snacks/ration/side/cornbread = 1,
-		/obj/item/reagent_containers/food/snacks/ration/snack/pizza_crackers = 1,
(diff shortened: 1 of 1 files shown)


FalloutFalcon:
the vegan part was specifically the first two items.
The idea was that every ration include a wide variety of foodtypes which encourages snack trading.

zimon9:
While I've actually participated in snack trading because of this, I haven't gotten the opportunity to do it very often. 

Usually, a vegan character of mine ends up ripping open another ration pack or two and stealing the non-meat products in _those_ rations, instead, in the same form of someone drinking from a gallon of milk with their mouth, instead of using a cup.

zimon9:
While I still think that it's a slightly silly inconsistency, what if, instead, the name of the pizza crackers were changed so as to make it clear that it has meat? Something like "pepperoni pizza crackers"?

PR #3410, opened 2024-09-22
Exosuit generator power production change and efficiency decrease

zimon9:
## About The Pull Request

This change increases the power produced by exosuit generators twenty-fold. It also reflavors the description of the generators. This change also decreases the maximum fuel that each generator can hold, to the point where the standard plasma generator functions for only a minute, maximum, when fully fueled. The exonuclear reactor only functions for slightly longer than this. 

## Why It's Good For The Game

Exosuit reactors are very slow, with regards to powering equipment. This change should make them more useful, both in combat, and out of combat. This change is meant to make them work more like the battery module present in FTL, which gives you an energy boost, but only for a very limited time.

## Changelog

:cl:
balance: rebalanced exosuit generators to produce more power, at the expense of longevity and total energy production
/:cl:

### DIFF
--- a/code/game/mecha/equipment/tools/other_tools.dm
+++ b/code/game/mecha/equipment/tools/other_tools.dm
@@ -360,15 +360,15 @@
 
 /obj/item/mecha_parts/mecha_equipment/generator
 	name = "exosuit plasma converter"
-	desc = "An exosuit module that generates power using solid plasma as fuel. Pollutes the environment."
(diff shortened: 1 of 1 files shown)


Apogee-dev:
The gennie modules as they currently are are more for powering a mech over a long time (say, while scouting a planet’s surface) without needing to return to a charging station. I don’t think what you want here is a bad idea, but it would be better served by making a new module instead of reworking an existing one.

zimon9:
> The gennie modules as they currently are are more for powering a mech over a long time (say, while scouting a planet’s surface) without needing to return to a charging station. I don’t think what you want here is a bad idea, but it would be better served by making a new module instead of reworking an existing one.

Ah, okay! That makes sense to me, thank you. I'll leave these generators alone, then, and I'll start work on a high-powered short-term usage generator, instead.

PR #3413, opened 2024-09-22
[DRAFT] Adds a high-performance exosuit generator

zimon9:
## About The Pull Request

This change adds a supercharged version of the standard exosuit plasma generator. 

It produces power at twenty times the standard rate, but it does so by burning plasma in a far less efficient manner. It also has a decreased fuel storage capacity, owing to the larger engine using up more space in the module. 

Fully fueled with sixty plasma sheets, this generator produces power for only one minute, so users of it must be careful with regards to when they choose to utilize it. [NOTE: Following modifications, this will need to be looked at again. Likely, the rate at which it eats up fuel may need to be modified]

They can be purchased from the outpost for 1,500 credits, and they can be printed from an exofabricator. They require advanced plasma research to be completed before they can be printed out.

## Why It's Good For The Game

This device is intended to be like the battery module in FTL, which grants you increased power for a limited amount of time. I believe that this can increase the complexity of exosuit combat, if only marginally, and thus prove to be a fun addition to the current set of equipment that they have. 

## Changelog

:cl:
add: Adds a supercharged exosuit generator
/:cl:

### DIFF
--- a/code/game/mecha/equipment/tools/other_tools.dm
+++ b/code/game/mecha/equipment/tools/other_tools.dm
@@ -471,6 +471,15 @@
 		radiation_pulse(get_turf(src), rad_per_cycle)
 
 
+/obj/item/mecha_parts/mecha_equipment/generator/supercharged
+	name = "supercharged exosuit plasma converter"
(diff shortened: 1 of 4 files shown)


thgvr:
Stale, ask a maintainer to have this reopened.

PR #3296, opened 2024-08-17
Crafting Cruft Removal

Sadhorizon:
## About The Pull Request
Removes the following recipes because:

Clothing 
- Scrap Armor - it sucks, it's ugly, it's expensive and it's honestly a bait to have it craftable

Drinks
- Holy Water Flask - god died
- Nothing Bottle - mimes died

Misc
- Flashlight eyes - just why.
- Extendo-hand - TG slop
- Guillotine - meme item
- Moth Plushie - plushies are cool, but the recipe itself is a meme

Robots
- Honkbot - clowns died

Tribal
- Bonfire, Distiller, Mushroom Bowl and Charcoal Stylus moved to Misc - they are okay
- Bone Spear, Bone Axe and Bone Sword moved to Weapons - Hate to keep them, but there isn't really any other okay melee weapons available, outside of the crusher I suppose, but that one's inconsitent to get.
- Bone Armlet, Wolf Fang Necklace, Goliath Cloak, Hunters Belt and Polar Cloak moved to Clothing - they are okay
- **Everything else got removed, including the category itself** - it's about time to let ash walker gear go I think

Weapons/Ammunition
- Explosive Lance - item's definitely on the goofier side
- Teleprod - teleportation is often fucked
- Gonbola - gondolas barely exist, might as well axe it
- Flamethrower - feels like this should be locked behind more than crafting
- Chainsaw moved to Misc - it's a gardening tool, you MANIACS
- Pipe Bow and All The Arrows - bows are kinda broken rn? Commented it out cuz feels like it can be used later, but having broken stuff here is dumb.

**Note here: It's just the crafting recipes, not foods themselves. Crafting cleanup, not food cleanup.**

Breads
- Mimana Bread - mimes died

Burgers
- Ghost Burger - magic is not real
- Clown Burger - clowns died
- Mime Burger - mimes died
- Spellburger I and II - magic is not real

Cakes
- Slime Cake - you are NOT getting slime extracts for this
- Angel Food Cake's recipe was changed - god died, so it uses cream instead of holy water
- Clown Cake - clowns died

Frozen
- Honkdae - clowns died
- Mime Snowcone - mimes died
- Clown Snowcone - clowns died
- Also renamed all the cola cones according to names from the drink rename

Meats
- Meat Clown - clowns died

Pastries
- Honk-pockets - clowns died

Pies
- Mime Tart - mimes died

Salads
- Monkey's Delight - it's very questionable

Soups
- Clowns Tears - clowns died

Spaghettis
[...]

PR #3417, opened 2024-09-23
Obliterates the lasso (goliath riding no longer requires it)

SomeguyManperson:
## About The Pull Request

No longer player accessible as of #3296 which means less yeehaw

## Why It's Good For The Game

~~horse~~ goliathback cowboy combat

## Changelog

:cl:
balance: you no longer need a lasso to ride a goliath
del: no more lasso
/:cl:


FalloutFalcon:
should.. we be able to ride golaiths in the first place?

PR #3426, opened 2024-09-24
mudskipper and shetland window fix

Jedi-Toothpaste:
## About The Pull Request
Added windows to the mudskipper and shetland's engines as well for consistency see #3434 for the Kilo fixes too.

Also seperated the Shetland's engine rooms so both buttons don't open both sides.
Now the button in each engine room will open only it's respective side.

## Changelog

:cl:
fix: Added windows to the mudskipper and shetland's engines.
fix: Adjusted the blast doors which open on the Shetland's engines.
/:cl:


Zevotech (approved):
im going to bed after this

PR #3441, opened 2024-09-26
Reduces the chance for grenades to go off in your hands

zimon9:
## About The Pull Request

This change reduces the probability of a grenade blowing up in one's hands from 15% down to 5%.

Please let me know if this probability is too low. I just chose an oddball low number to use.

The check was also modified so that the dice roll is only made if a projectile hits a hand holding a grenade.

## Why It's Good For The Game

It always felt a little too high of a chance to me. The probability roll happens for _each_ projectile that one gets hit with, and with how often and how rapidly one can be hit in a firefight here, the chance of it blowing up in ones hands goes from being somewhat unlikely to being more likely to happen than not.

image
This calculator gives the probability of the event occurring as being 68% likely, if one were shot 7 times. This is around how many shots it takes to bring someone down, on average, from my anecdotal experience.

This currently has the effect of making grenades useful in only a few very specific scenarios. 

If the probability of this were to go down, it would encourage the use of grenades more, which I feel can enhance fights.

## Changelog

:cl:
balance: rebalanced the probability of a grenade going off in one's hands
/:cl:

### DIFF
--- a/code/game/objects/items/grenades/grenade.dm
+++ b/code/game/objects/items/grenades/grenade.dm
@@ -154,11 +154,12 @@
 	var/list/valid_hands = list(FALSE, FALSE)
 	if (istype(owner.held_items[1], (/obj/item/grenade)))
 		valid_hands[1] = TRUE
-		message_admins("<span class='danger'>[owner.held_items[1]]</span>")
+		message_admins("<span class='danger'>Left arm</span>")
 	if (istype(owner.held_items[2], (/obj/item/grenade)))
 		valid_hands[2] = TRUE
-		message_admins("<span class='danger'>[owner.held_items[2]]</span>")
+		message_admins("<span class='danger'>Right arm</span>")
 
+	// P.def_zone = "r_arm" and P.def_zone = "l_arm" are the way to go
 	if(damage && attack_type == PROJECTILE_ATTACK && P.damage_type != STAMINA && prob(2)) //2% chance to go off
 		owner.visible_message("<span class='danger'>[attack_text] hits [owner]'s [src], setting it off! What a shot!</span>")
 		var/turf/T = get_turf(src)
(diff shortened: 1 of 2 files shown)


code/game/objects/items/grenades/grenade.dm
```diff
@@ -150,7 +150,7 @@
 
 /obj/item/grenade/hit_reaction(mob/living/carbon/human/owner, atom/movable/hitby, attack_text = "the attack", final_block_chance = 0, damage = 0, attack_type = MELEE_ATTACK)
 	var/obj/projectile/P = hitby
-	if(damage && attack_type == PROJECTILE_ATTACK && P.damage_type != STAMINA && prob(15))
+	if(damage && attack_type == PROJECTILE_ATTACK && P.damage_type != STAMINA && prob(1) && prob(50))
```
FalloutFalcon:
```suggestion
	if(damage && attack_type == PROJECTILE_ATTACK && P.damage_type != STAMINA && prob(0.5))
```

FalloutFalcon (dismissed):

code/game/objects/items/grenades/grenade.dm
zimon9: Oh, thank you! I wasn't sure if the argument only took integers or if it took floats, so I just did it like this to ensure that it worked. I'll fix this, thank you.

zimon9: The modification should be complete now

Apogee-dev:
Honestly this should just be a chance that rolls for each projectile that hits that specific arm rather than the player in general. 15% would make a lot more sense if that were the case.

zimon9:
I'll try and implement that. I will say, however, I'd prefer if the probability were still reduced down to, say, 2-5% even with this change, as bullets can hit limbs just from natural spread, and with the current probability calculation, the chance of it going off would still be fairly high, if it were kept at 15%.

zimon9:
> Honestly this should just be a chance that rolls for each projectile that hits that specific arm rather than the player in general. 15% would make a lot more sense if that were the case.

I've implemented a check to see if a projectile hits an arm holding a grenade, and I also set the probability to 5% for this. Please let me know if this looks okay? Thank you.

PR #3378, opened 2024-09-16
Cabinet Refactor

Erikafox:
## About The Pull Request
Mildly refactors fireaxe cabinets to allow them to be subtyped with different items. Also adds an is_wielded proc to most weapons that can be wielded to prevent potential weirdness with this down the line.

## Why It's Good For The Game
Cabinets are soulful

## Changelog

:cl: Erika Fox
code: Fireaxe cabinets have been repathed, and now function as a more general cabinet object. please report any inconsistencies with behavior
/:cl:

PR #3470, opened 2024-10-02
[DNM]Revert "Cabinet Refactor" to see if defibs get fixed

rye-rice:
Ditto

If it does it will be kept on TM 
Reverts shiptest-ss13/Shiptest#3378


Erikafox:
yeah it's probably the checking for trait_wielded instead of the var or whatever in the defibs

rye-rice:
cwan ywou fwix iwt erwika qwq

Erikafox:
im gaming right now otherwise i would. i need to get my weekly autistic girlthing rep in.

Erikafox:
if you want to fix it search for TRAIT_WIELDED in defib.dm

FalloutFalcon:
> yeah it's probably the checking for trait_wielded instead of the var or whatever in the defibs

its the opposite.
its checking for the wielded var but you removed the thing that updates it
you just need to refactor the stuff to use the trait

Erikafox:
MANY such cases of a problem just because of me being stupid

PR #3482, opened 2024-10-06
Un-nerfs Kepori

thgvr:
## About The Pull Request
Gonna try this out to see if it improves how Kepori actually feel. I don't really think straight up taking more damage is an interesting downside, considering their only upside was being a little faster. This also greatly lowers the movespeed modifier. 

Personally I don't really think it's a good idea to have a straight numbers modifier as a nerf like this, considering we balance things without even thinking about this. Projectiles and mob attacks do a high amount of damage, and it's all but guaranteed to break or completely disable a limb in one hit. With limb armor being fairly rare to come by, and the area a projectile/melee targets having RNG on where they actually hit, it's not actually that interesting.

## Changelog

:cl:
balance: Removed Kepori damage modifiers.
balance: Decreased Kepori move speed modifier. (They are still a bit faster than average)
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/kepori.dm
+++ b/code/modules/mob/living/carbon/human/species_types/kepori.dm
@@ -17,7 +17,7 @@
 	species_clothing_path = 'icons/mob/clothing/species/kepori.dmi'
 	species_eye_path = 'icons/mob/species/kepori/kepori_eyes.dmi'
 	heatmod = 0.67
-	// coldmod = 1.5
+	coldmod = 1.5
 	// brutemod = 1.5
 	// burnmod = 1.5
 	speedmod = -0.10
--- a/code/modules/mob/living/carbon/human/species_types/kepori.dm
+++ b/code/modules/mob/living/carbon/human/species_types/kepori.dm
@@ -17,10 +17,10 @@
 	species_clothing_path = 'icons/mob/clothing/species/kepori.dmi'
 	species_eye_path = 'icons/mob/species/kepori/kepori_eyes.dmi'
 	heatmod = 0.67
-	coldmod = 1.5
-	brutemod = 1.5
-	burnmod = 1.5
-	speedmod = -0.25
+	// coldmod = 1.5
+	// brutemod = 1.5
+	// burnmod = 1.5
+	speedmod = -0.10
 	bodytemp_normal = HUMAN_BODYTEMP_NORMAL + 30
 	bodytemp_heat_damage_limit = HUMAN_BODYTEMP_HEAT_DAMAGE_LIMIT + 30
 	bodytemp_cold_damage_limit = HUMAN_BODYTEMP_COLD_DAMAGE_LIMIT + 30


Anticept:
I just want to share some thoughts. I get taking out the damage mod, but at the same time... I do like the RP the fragility brings where crew are being accomodating for some of kepori's drawbacks, and it brings a difference in playstyle since I have to adapt to a speed/fragility build. Speed for damage makes Kepori feel unique instead of just look unique.

I don't think I personally am okay with taking it out completely but perhaps leave in a slight damage mod at least until we have something interesting to replace it. I agree regarding how fast limbs are taken out are a little crazy, but I really like the "I am speed" approach to playing them.

Unathi on paradise never did get anything unique to them other than looks until recently (now they do the welding fuel fire breath trick), all they were was re-skinned humans and the maintainers blocked any attempts at giving them strengths and weaknesses despite the other races having something for them.... and it just made them *less* interesting overall.

thgvr:
I don't really think damage modifiers are interesting or fun, we can do better

Anticept:
I agree, but I kind of wish we could keep a little bit of it until we figure out something new. I like species mechanical differentiation and speed.

Orchidthederg:
The speed was a pretty notable part of the race so maybe don't nerf it too substantially. Just my 2 cents tho coming from a kep main

Sun-Soaked:
> I don't really think damage modifiers are interesting or fun, we can do better

I agree, but I'm not sure about bringing them closer to baseline in the mean time

thgvr:
> > I don't really think damage modifiers are interesting or fun, we can do better
> 
> I agree, but I'm not sure about bringing them closer to baseline in the mean time

What would you suggest?

If I bring back sprinting (probable), I could edit their max stamina values or something, or have them sprint faster and drain their sprint faster.

Anticept:
That would be fun to have them have high burst speed for very short times if you do. I'm just really on the fence about taking away their speed right now as we don't have that sprint feature back yet, and the speedmod as we already have at this moment isn't a whole lot faster. cutting it down by 3/5ths will make their speed unnoticeable short of running several screens.

TDHooligan:
I don't think a wholesale 'damage modifiers are bad' holds water.
If the modifier individually feels impactful and fun, it's a net positive. 'damage mods bad', in my experience, holds true if the number isn't significant enough to impact player choice. and a 50% increase in damage is definitely impactful. 

Balance on damage numbers used across the codebase is a tough one. I think as a result it's got to be more vibes-based, perhaps lowering the multiplier if it's really that crippling for kepori.

If limbs alone are the issue, perhaps there's a better adjustment to keep the frailty high without outright causing limb breaks. they'd just need some more refactoring. 
IE: Redirecting some damage to the torso, lowering limb hit chance somehow, or partially replacing the damagemod with a max health modifier somehow, so you crit just as fast, losing less limb HP.

As is, this PR removes the race's biggest tradeoff, leaving easily avoidable downsides (wear warmer clothes. dont punch shit) for a hard to obtain speed increase. Lots of players (myself included) rarely remove a hardsuit once it's on. In that common case, this race is a straight upgrade you never have to think about playing around.

code/modules/mob/living/carbon/human/species_types/kepori.dm
```diff
@@ -17,10 +17,10 @@
 	species_clothing_path = 'icons/mob/clothing/species/kepori.dmi'
 	species_eye_path = 'icons/mob/species/kepori/kepori_eyes.dmi'
 	heatmod = 0.67
-	coldmod = 1.5
-	brutemod = 1.5
-	burnmod = 1.5
-	speedmod = -0.25
+	// coldmod = 1.5
```
SomeguyManperson: This is a temperature damage modifier like heatmod, and probably doesn't need to be changed

code/modules/mob/living/carbon/human/species_types/kepori.dm
```diff
@@ -17,10 +17,10 @@
...
+	// brutemod = 1.5
+	// burnmod = 1.5
```
SomeguyManperson, line 22:
```suggestion
	brutemod = 1.25
	burnmod = 1.25
```
Increases effective health from 66 to 80. For something even more slight go with 1.1, which would increase effective health to a little over 90. Speed boosts are noticeable, but I don't think it's -30 health levels of noticeable

code/modules/mob/living/carbon/human/species_types/kepori.dm
thgvr: yeah fair, must have caught it on accident

code/modules/mob/living/carbon/human/species_types/kepori.dm
thgvr: 1. The damage mods just aren't interesting and mess with the way we balance bullet/mob damage. I'd like to come up with something more interesting.
2. The speed boost was always a *lot*, I should have adjusted it when I adjusted config move modifiers but didn't.

PR #3483, opened 2024-10-06
Increases flavor text length by 512 characters

zimon9:
## About The Pull Request

This change increases the maximum length that one's flavortext can be up from 1024 characters to 1536 characters.

## Why It's Good For The Game

While I think that the maximum length that one's flavortext can be is large enough as it is, it may be nice for it to be increased slightly. This would increase the maximum length to be equal to 1.5 times the current value.

## Changelog

:cl:
code: increased maximum flavortext length to 1536 characters
/:cl:

### DIFF
--- a/code/__DEFINES/say.dm
+++ b/code/__DEFINES/say.dm
@@ -96,7 +96,7 @@
 
 //Don't set this very much higher then 1024 unless you like inviting people in to dos your server with message spam
 #define MAX_MESSAGE_LEN 1024
-#define MAX_FLAVOR_LEN 1024
+#define MAX_FLAVOR_LEN 1536
 #define MAX_SHORTFLAVOR_LEN 100
 #define MAX_NAME_LEN 42
 #define MAX_BROADCAST_LEN 512


thgvr:
The point of flavor text is it's not supposed to be a wall of text

rye-rice (approved):

zimon9:
> The point of flavor text is it's not supposed to be a wall of text

That is entirely true, and while I do think I should make some of my character descriptions more concise, I just feel that there is just so much space to describe a character and their minute details, sometimes. 

There's something that's nice about being able to describe certain wordier aspects of a character that cannot be shown in-game (as of right now), like jewelry, for instance.

FalloutFalcon:
I like the idea of needing to constrain our flavor texts to a semi easy to read length. The current length feels like it can convey most information that would be needed and the rest should probably be shown in role play.
visual representaiton of the character limits

current
image
new
image

FalloutFalcon:
For the reasons listed above most of the maints seem to disagree with this pr. 
"we have more than enough walls of flavor text as is. we shouldn't encourage people to extend theirs further. they should be readable without having to stop your game for like 5 minutes to read someone's novel about their character."

zimon9:
Oh well. I understand the reasoning behind this being closed. Thank you all for the consideration.

PR #3323, opened 2024-09-02
Remaps the Kansatsu-class Scout Courier.

Sadhorizon:
## About The Pull Request
**THE SHIP IS NO LONGER SPAWNABLE FREELY**

**TODO:**
- [x] floors :D
- [x] ship desc
- [x] more missions
- [x] make command bed room not suck
- [x] give IO a privacy switch or whatever
- [x] 22 cargo desc
- [x] gut the disks

Biggest changes:
- **The Kansatsu now gets mission holodisks. Look at Ranger and LP missions to see what I mean.** The idea behind those is that they are gimmicky objectives based around spy stuff because that's cool. Intel gathering and minor theft. Don't want them to be a license to antag either because that's dumb. **I AM IN NEED OF MORE IDEAS. PLEASE GIMME.** 
- It's no longer a 1 tile hell.
- You generally get more gear now.
- The ship is now more flavorful.
- There are now security cameras in every room. Yes, bathroom gets one too.

Thank you @Bjarl for an updated ship description <3

<details><summary>Screenshots:</summary>
<p>

scrnshot1

StrongDMM-2024-09-17 10 39 23

StrongDMM-2024-09-22 10 41 58

</p>
</details> 

## Why It's Good For The Game
Kansatsu definitely needs a refresh.
Maptainers wanted it to be an aspawn.
Also I think the missions are going to be a great addition. Some pre-made gimmicks you can do.

## Changelog

:cl:
tweak: Kansatsu-class Scout Courier has been remapped.
del: Kansatsu-class is now aspawn-only.
del: Removes all the misc mining gear from the Field Agent outfit.
/:cl:

PR #3485, opened 2024-10-06
Reflavors the Li Tieguai-class.

Sadhorizon:
## About The Pull Request
Li Tieg is now more similar to #3323 style-wise, with the difference that it's in Biodynamics teal instead of fully in VS red.

<p>

scrnshot1

StrongDMM-2024-10-17 21 02 51

StrongDMM-2024-10-06 23 02 48

</p>
</details> 

Also tweaks Li Tieg's weaponry:
- Trauma team looses one Ringneck, but gains a Rattlesnake. - Rattlesnake is just cool.
- Captain looses the Viper. It's replaced by a Himehabu and an edagger. - A Himehabu fits CS caps waaaay better than a Viper. Edagger is also CS special.
- Director looses the Ringneck and the Dart Gun. They only get an edagger now. Sorry. - I'd say they really don't need a gun. Edagger is cool though.

## Why It's Good For The Game
Please stop with the carpets and curtains. I beg you.

## Changelog

:cl:
tweak: Reflavored Li Tieguai-class to use Cybersun Biodynamics colors.
balance: Removed the Viper, the Dart Gun and two Ringnecks from the Li Tieguai. Instead, added a Rattlesnake, a Himehabu and two edaggers.
fix: Li Tieguai captain is now an officer.
/:cl:


generalthrax:
I've captained the Li Tieguai a lot. Possibly more than anyone else. There is one major issue (to me) that this could fix

image

The armoury being in the dorms always felt really weird. Felt like it could be swapped places with the bathroom to be in the Trauma Team prep room instead, up here.

image

Apogee-dev:
Oh right, please put the guns in pistol cases. They shouldn’t be spawning loose unless they’re in a gun rack or display case or something like that- iirc we were moving towards using those cases even in job equipment lockers like we see here.

Sadhorizon:
> The armoury being in the dorms always felt really weird. Felt like it could be swapped places with the bathroom to be in the Trauma Team prep room instead, up here.

I feel like it would either make the bathroom hyper squished (bad), or make it cut into dorms (also bad).

I agree that it sucks but I really don't want to fix the layout in here, as, let's be real, most of it sucks.

PR #3502, opened 2024-10-08
E40 Changes (.299 cargo and firerate)

generalthrax:
## About The Pull Request

Brings up the E40 ballistic mode firerate by a fair amount to compensate for the significantly lower damage. Experimenting, mostly. Takes the .299 Caseless box out of the black market and adds it to cargo. Fixes a typo that said it held 75 rounds when it actually held 120. You're gonna need that ammo, it guzzles.

## Why It's Good For The Game

E40 is a Big, Big purchase if you don't already have it and the Raleigh starts with the rifle and needs to Gamble in order to get more ammo. Should make it the legendary, expensive weapon it is And make both components Actually Usable.

## Changelog

:cl:
balance: E40 hybrid rifle ballistic mode now shoots faster
balance: .299 Caseless taken out of the black market, added to cargo
fix: .299 Caseless now correctly states it holds 120 rounds
/:cl:


code/modules/cargo/packs/ammo.dm
```diff
@@ -422,6 +422,11 @@
 	contains = list(/obj/item/ammo_box/a44roum/hp)
 	cost = 500
 
+/datum/supply_pack/ammo/c299
+	name = ".299 Eoehoma Caseless Ammo Box Crate"
+	desc = "Contains a 120-round box of .299 Caseless ammo from the defunct Eoehoma. Used for the E-40 Hybrid Rifle."
+	contains = list(/obj/item/ammo_box/c299)
```
SomeguyManperson: Should probably have a cost defined here I think

generalthrax: Swore I did. Must've forgot

Apogee-dev:
I have concerns here regarding the Raleigh. Part of the reason the Raleigh actually gets an E40 when no other ship gets an assault rifle of any description is actually because of the lower peak damage output than a normal AR. If the damage output gets buffed it might end up too strong to just _give_ somebody. I think the flexibility of being able to swap between laser and ballistic damage as needed makes up for lower peak dps.

DIB-DOG:
I may be wrong, but doesn’t the Atlas class start with an assault rifle?

Still not fullmerged ofc, but just for sake of argument.

Apogee-dev:
> I may be wrong, but doesn’t the Atlas class start with an assault rifle?
> 
> Still not fullmerged ofc, but just for sake of argument.

true actually, they _do_ get one CM-82.... but the atlas is also an actual military ship. and i don't think the E40 is _weaker_ than a ballistic assault rifle anyway, it's much more flexible at the cost of having slightly less dps. the dps is the reason i put ARs right on the cusp of acceptability for a normal ship.

FalloutFalcon (approved):
good code wise, probably doesn't need a tm.

FalloutFalcon (approved):

PR #3518, opened 2024-10-10
Suspicious Duffel Bags now incur movement slowdown

MemeSnorfer:
## About The Pull Request

Suspicious duffel bags now incur movement slowdown like any other duffel bag.

## Why It's Good For The Game

The amount of times I've seen this used to powergame storage is more times than I'd like to see.

## Changelog

:cl:
balance: Suspicious Duffel Bags now incur movement slowdown
/:cl:

PR #3521, opened 2024-10-10
Syndicate duffle bags are slighlty faster then a normal duffel

FalloutFalcon:
## About The Pull Request
"reverts" #3518 which changes the slowdown from 0 to 1. This changes it to 0.75

## Why It's Good For The Game
Prevents the duffel from being a straight upgrade to backpacks and instead makes it a slightly better duffel
Fits the vibe that they are tactical and thus faster then a normal duffel.

## Changelog

:cl:
add: Syndi duffels are slightly faster then a normal dufflel, powergamers rejoice.
/:cl:


Apogee-dev:
I think the full removal was fine. Syndicate-origin gear should not generally be a straight upgrade over normal equipment (this applies to the syndie hand tools too, for example)- making it an upgrade like this means mappers can’t just use it the same way we would a normal duffel bag as gear on ships and player outfits.

meemofcourse:
sorry, no

PR #3529, opened 2024-10-11
Outpost Vendors Cost Money

Erikafox:
## About The Pull Request
Outpost vendors now cost some money (instead of selling everything for free)
also ports https://github.com/tgstation/tgstation/pull/54463

## Why It's Good For The Game
The economy is in shambles

## Changelog

:cl:
balance: Outpost vending machines now cost money.
/:cl:


Sadhorizon:
Issue here is that prices are balanced around stations.
I think it all should be made *very* cheap, as more of a symbolic price than anything. The likes of 10 credits per item maybe. So the outpost doesn't give stuff for free, but also, you can still buy all you really need with the roundstart 150.

Also, make sure it actually works. I tried this before and I couldn't actually buy stuff, probably because of bank cards.

Erikafox:
yeah i made it work and also changed prices across the board. balance is pulled completely out of my ass.

code/modules/vending/_vending.dm
```diff
@@ -194,6 +193,8 @@ IF YOU MODIFY THE PRODUCTS LIST OF A MACHINE, MAKE SURE TO UPDATE ITS RESUPPLY C
 
 	Radio = new /obj/item/radio(src)
 	Radio.listening = 0
+	if(istype(get_area(src.loc), /area/outpost) || istype(get_area(src.loc), /area/ruin))
+		all_items_free = FALSE
```
FalloutFalcon, line 197: I think this would make more sense as a !ship_area but this is also fine.

FalloutFalcon (approved):
If it was me I would also let you hack them to turn the prices on and off.

Erikafox:
> If it was me I would also let you hack them to turn the prices on and off.

I feel like that's a way to completely circumvent what im doing with this

Anticept:
This PR seems to have made ALL vendors cost money, including ones found as loot.

This is a bit odd for places like the singulo lab for example, where the engineering vendors are charging money...

Erikafox:
image

Anticept:
I saw it. That makes this PR's title and its changelog incomplete and misleading since it only says outpost but is being applied to ruins too, which is odd for some vendors, and in a roundabout way, is changing loot available in ruins.

tantspants:
prices are totally ridiculous. 100 cr for lime juice? It should be 1/10 that.

Erikafox:
make a pr to reduce prices?

PR #3553, opened 2024-10-13
Console examine_more viewability change

zimon9:
## About The Pull Request

This PR makes it so that you can view a console, laptop, a wideband log, the screen on an APC, and the gasses measured from an air alarm, from one tile away. You can not interact with the UI unless you are right next to it. To do this, the examine_more function is used.

## Why It's Good For The Game

It makes sense to be able to see screens without having to be right next to them. You will still be unable to interact with them, but you no longer need to be right next to them just to open up the UI. Laptops have also always required holding them in your hands to open up the UI, and this change should address that so that this is no longer necessary.

## Changelog

:cl:
add: Added ability to see screens using the examine_more function
/:cl:


FalloutFalcon (approved):
I like this, one change i might make is only having this work within a view tiles. Its fine for the bigger screens but you really shouldn't be able to read a laptop from a full screen width.
just needs test merges to ensure it works as intended. 
I haven't looked at the code for a while and coulda sworn these would normally auto-close.

zimon9:
> I like this, one change i might make is only having this work within a view tiles. Its fine for the bigger screens but you really shouldn't be able to read a laptop from a full screen width. just needs test merges to ensure it works as intended. I haven't looked at the code for a while and coulda sworn these would normally auto-close.

Ah, it already behaves this way. Example: 

https://github.com/user-attachments/assets/0eddfcf5-ce09-40d5-9876-de56627ce9bf

It also auto-closes when you move sufficiently far away, in this case being two tiles away.

PR #3556, opened 2024-10-14
Syndicate tools and soap are no longer better.

Sadhorizon:
## About The Pull Request
- syndicate soap clean speed swapped from 5 to 27 (1 better than NT soap, because I find that funny). Also it's uses are tripled.
- "suspicious toolbox" renamed to "black and red toolbox", lost it's damage buff.
- syndicate tools have lost the "suspicious" names, their descriptions, and their toolspeed buff.

## Why It's Good For The Game
Cruft. Syndi items were better on stations because of rarity. On shiptest, it feels kinda weird to map those in knowing they are just upgrades. Soap is especially awkward in that regard.

## Changelog

:cl:
tweak: Syndicate soap is now in-line with NT soap.
del: Syndicate tools are no longer faster. They are just reskinned normal tools now.
del: Syndicate toolbox lost it's damage buff.
/:cl:

### DIFF
--- a/code/game/objects/items/clown_items.dm
+++ b/code/game/objects/items/clown_items.dm
@@ -71,7 +71,7 @@
 /obj/item/soap/syndie
 	desc = "An untrustworthy bar of soap made of strong chemical agents that dissolve blood faster."
 	icon_state = "soapsyndie"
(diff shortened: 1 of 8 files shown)


generalthrax:
Make sure to go the full way and remove them from the black market, as well.

PR #3559, opened 2024-10-14
Allows fire extinguishers to be refilled with sinks

zimon9:
## About The Pull Request

This change adds the ability to refill a fire extinguisher with a sink.

## Why It's Good For The Game

I've always found it a little inconsistent how a water dispenser is required in order to refill a fire extinguisher. My hope is that this change will give more options for players to refill their extinguishers.

## Changelog

:cl:
add: Added ability to refill fire extinguishers with sinks
/:cl:

### DIFF
--- a/code/game/objects/items/extinguisher.dm
+++ b/code/game/objects/items/extinguisher.dm
@@ -20,7 +20,7 @@
 	var/chem = /datum/reagent/water
 	var/safety = TRUE
 	var/refilling = FALSE
-	var/tanktype = /obj/structure/reagent_dispensers/watertank
+	var/tanktype = /obj/structure/ //A simple tanktype change won't do it
 	var/sprite_name = "fire_extinguisher"
 	var/power = 5 //Maximum distance launched water will travel
 	var/precision = FALSE //By default, turfs picked from a spray are random, set to 1 to make it always have at least one water effect per row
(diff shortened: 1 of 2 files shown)


Apogee-dev:
It doesn't strike me as inconsistent, how're you getting the requisite water pressure out of a sink?

zimon9:
My idea for that is to include a integrated hand-pump into the fire extinguishers (as part of their description, and not as an actual mechanic). I only felt this way because I always imagined water dispensers as being like, gravity fed, as they don't take power. Sort of like these? 
image

zimon9:
This is just my idea of how dispensers are, though -- I'm liable to be wrong about it.

Apogee-dev:
Yeah, looking at the water dispenser sprites (and the fuel dispenser sprites) they _might_ have a motorized pump+nozzle on them- I've certainly seen other takes on their sprites that lean hard into that angle, and it's one I quite like. From a game mechanics angle I do also like the fact that extinguishers need a water dispenser- every single ship has at least one water dispenser, and you can't just craft them out of thin air like you can do for a sink. Player-constructed sinks as an infinite, on-demand source of water are already a bit of a problem anyway.

zimon9:
> Yeah, looking at the water dispenser sprites (and the fuel dispenser sprites) they _might_ have a motorized pump+nozzle on them- I've certainly seen other takes on their sprites that lean hard into that angle, and it's one I quite like. From a game mechanics angle I do also like the fact that extinguishers need a water dispenser- every single ship has at least one water dispenser, and you can't just craft them out of thin air like you can do for a sink. Player-constructed sinks as an infinite, on-demand source of water are already a bit of a problem anyway.

This is pretty fair! I think having a motorized pump in the dispenser itself is a better way to see them yeah. I agree that this would be kind of balance-breaking, so I'll close this now. Thanks for the input!

PR #3560, opened 2024-10-14
Crate Shelf Tweaks

Erikafox:
## About The Pull Request
Mapped crate shelves now start at a height of 3. You can build them to a max height of 4
Constructed crate shelves start at 1 crate and need to be built up
You can now add shelves to crate shelves (with metal)

## Why It's Good For The Game
it's not, honestly. i think this change is going to be the death of the game. for pr after pr i've made nothing but hastily produced shitcode changes to a game that used to be soulful and thriving, stripping out the identity of what's made it so interesting to people. this is where it comes full circle. im sorry for doing this. but it had to be done.

## Changelog

:cl:
add: You can now add shelves to crate racks
/:cl:


Apogee-dev:
I’m actually not sure what this one accomplishes, ngl (also you didnt mention the maximum shelf count being, if i’m reading this right, 4). it seems like kind of a nothing change to me?

generalthrax:
> it seems like kind of a nothing change to me?

They're on the Elated-Bolide and they look cool.

Orchidthederg:
Feels unnecessary and will make crate organization more of a PITA than it already is. Is there like, a _reason_ it should be reduced?

Erikafox:
> I’m actually not sure what this one accomplishes, ngl (also you didnt mention the maximum shelf count being, if i’m reading this right, 4). it seems like kind of a nothing change to me?

my bad on the max. that's fixed now.

generalthrax:
> Feels unnecessary and will make crate organization more of a PITA than it already is.

Shiptester reading comprehension at an all time low upon being forced to look at files changed (new max isn't at 4 as per 30 seconds ago)

code/game/objects/structures/crateshelf.dm
```diff
@@ -47,12 +50,35 @@
...
 	if (item.tool_behaviour == TOOL_WRENCH && !(flags_1&NODECONSTRUCT_1))
 		item.play_tool_sound(src)
 		if(do_after(user, 3 SECONDS, target = src))
 			deconstruct(TRUE)
 			return TRUE
+	if(istype(item, /obj/item/stack/sheet/metal) && capacity < 4)
+		var/obj/item/stack/sheet/metal/our_sheet = item
+		if(our_sheet.get_amount() >= 2)
+			balloon_alert(user, "adding additional shelf to rack")
+			if(do_after(user, 3 SECONDS, target = src))
```
FalloutFalcon:
```suggestion
			if(do_after(user, 3 SECONDS, src))
```

code/game/objects/structures/crateshelf.dm
```diff
@@ -47,12 +50,35 @@
...
+			stack_layer = ABOVE_MOB_LAYER + (0.02 * i) - 0.01
+		else
+			stack_layer  = BELOW_OBJ_LAYER + (0.02 * i) - 0.01 // Make each shelf piece render above the last, but below the crate that should be on it.
+		stack_offset = DEFAULT_SHELF_VERTICAL_OFFSET * i // Make each shelf piece physically above the last.
+		overlays += image(icon = 'icons/obj/objects.dmi', icon_state = "shelf_stack", layer = stack_layer, pixel_y = stack_offset)
+
 /obj/structure/crate_shelf/attackby(obj/item/item, mob/living/user, params)
 	if (item.tool_behaviour == TOOL_WRENCH && !(flags_1&NODECONSTRUCT_1))
 		item.play_tool_sound(src)
 		if(do_after(user, 3 SECONDS, target = src))
```
FalloutFalcon:
```suggestion
		if(do_after(user, 3 SECONDS, src))
```

FalloutFalcon (dismissed):
This looks good though I would rather the default stay at 3 and var edit smaller ones where needed. Unless you looked through all of the existing maps that use them and really think its a good change.
There is also no validation in add_shelf that its in the legal limit.
Also define or var for max capacity please! Also a small message stating you cant build any higher, and one saying you need more then 1 sheet would be good feedback.
Possibly even how many more layers you can add to it in its examine text if your feeling spicy.

Erikafox:
>amount

I think it's a fine change

Sun-Soaked (approved):
Code looks fine, besides requested changes.

I can get this merged if you return the baseline to 3 and do those do-after changes fallcon requested.

Erikafox:
The do_after changes are done.

I think having the default amount lower is Fine for a few reasons;
Ease of expansion means that it can be circumvented
Layer 3 is where they start to layer weirdly

PR #3566, opened 2024-10-15
Gezena is no longer barefoot

generalthrax:
## About The Pull Request

Someone commented out something to try and see why something was broken and left it. Digi isn't broken either so I'm not sure how this happened.

Also removes partial-insulation from the Gezenan captain's gloves that Apogee happened to forget.

## Why It's Good For The Game

It's too early in the day for the marines to get silly like this.

## Changelog

:cl:
fix: Gezena has shoes again
del: Gezenan captain gloves no longer have partial insulation in parity with other captain gloves.
/:cl:


thgvr:
Unless someone fixed this it causes the boots to not use digi sprites

generalthrax:
I tested it. Digi sprites work. I also have the controversial opinion that everyone in PGF ships should not be barefoot just because of one tiny issue.

thgvr:
huh was it broken? I never noticed

PR #3574, opened 2024-10-16
Ray cats are now a subtype of cats

Erikafox:
## About The Pull Request
as title

## Why It's Good For The Game
I think it's more interesting to make this something rare rather than something all cats just. Do.

## Changelog

:cl:
balance: All cats are no longer radiation indicators - only a specific breed of them.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/friendly/cat.dm
+++ b/code/modules/mob/living/simple_animal/friendly/cat.dm
@@ -1,6 +1,6 @@
 /mob/living/simple_animal/pet/cat
 	name = "cat"
-	desc = "Most modern cats hail from a solarian experimental geneline. The perfect purrtection from rats and radiation."
+	desc = "The perfect purrtection from rats."
 	icon = 'icons/mob/pets.dmi'
 	icon_state = "cat2"
 	icon_living = "cat2"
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/friendly/cat.dm
```diff
@@ -83,6 +76,68 @@
 	mob_size = MOB_SIZE_SMALL
 	collar_type = "kitten"
 
+/mob/living/simple_animal/pet/cat/raycat
+	name = "\improper raycat"
+	desc = "A member of a feline lineage lineage said to trace back to the Night Of Fire. A raycat is an adoring pet for the radiation adverse spacer."
```
goober3: > feline lineage lineage

laugh at this user

Erikafox: true....

rye-rice:
i dont think most cats are going to be thrown into rads, so  you wont see it constantly, i think its just a nice thing to have around as a "huh, I didn't know that" sort of deal

Apogee-dev:
yeah i agree with rye. it was also made a Universal Cat Thing as kind of a nod to both the Solarian penchant for genetic engineering and their history in the Night of Fire, which I found to be a nice touch.

FalloutFalcon:
the whole point of the idea of real ray cats is that they become very common to detect radiation.

FalloutFalcon:
gonna close this as like, not planned!

PR #3575, opened 2024-10-16
Wasteplanets Less Explosive

Erikafox:
## About The Pull Request
removes welding fuel tanks from the wasteplanet spawn pool

## Why It's Good For The Game
you will explode... less..

## Changelog

:cl:
del: Wasteplanets no longer spawn welding fuel tanks.
/:cl:


Apogee-dev:
I wonder if it would be possible to enforce a "safe zone" biome for each planet type in the landing areas that's just the default biome without specific hazards so we can still have things like the tanks without ships getting randomly exploded by fauna or landing in the middle of an acid lake

PR #3584, opened 2024-10-18
Uncomments Machete Cargo Order

generalthrax:
## About The Pull Request

Uncomments the Machete Crate from cargo

## Why It's Good For The Game

They're bulky but they're neat melee options. Beats you to death with machetes

## Changelog

:cl:
add: Added Stamped Steel Machetes, 5 for 500
/:cl:

### DIFF
--- a/code/modules/cargo/packs/sec_supply.dm
+++ b/code/modules/cargo/packs/sec_supply.dm
@@ -77,7 +77,6 @@
 					/obj/item/melee/knife/survival)
 	crate_name = "combat knife crate"
 
-/*
 /datum/supply_pack/sec_supply/machete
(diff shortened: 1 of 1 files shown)


FalloutFalcon (dismissed):
I continue to regret not commenting these out in the melee re-path they were included in.
At present it should be worth more then 500. I would drop it down to two per crate which sounds fine for the time being.
The intent was that they would be very breakable, shown by the integrity, but it did not manifest in the melee path refactor that they were included in.
The melee pr reworking these is my priority PR post missions.

generalthrax:
> I would drop it down to two per crate which sounds fine for the time being

Done. I think two machetes for 500 is reasonable enough to stay, though.

PR #3585, opened 2024-10-18
Angle Grinder Balance

Erikafox:
## About The Pull Request
R-windows take less time to grind
blast doors can now be grinded
grinders slow you down way less now. 
Grinder noise is less grating

## Why It's Good For The Game
It's not

## Changelog

:cl:
balance: Angle grinders can now grind blast doors. reinforced windows are now less bad to grind down.
balance: angle grinder packs now slow you down less
balance: angle grinders now sound less Screechy
balance: angle grinders can now be safely used with a Bowman headset.
/:cl:


firebudgy:
thanks for stopping the tinnitus the angle grinder sound gave me

Erikafox:
Oh right I need to add that to my cl

code/game/machinery/doors/poddoor.dm
```diff
@@ -56,6 +58,17 @@
...
 				deconstruct(TRUE)
 
+	if(W.tool_behaviour == TOOL_DECONSTRUCT)
+		if(!(machine_stat & NOPOWER))
+			do_sparks(5, TRUE, src)
+			electrocute_mob(user, get_area(src), src, 1, TRUE) //zorp
+			close()
+		to_chat(user, span_notice("You start to cut [src] apart"))
+		if (W.use_tool(src, user, 15 SECONDS, volume = 75))
+			deconstruct(TRUE)
```
FalloutFalcon, line 68: should use /deconstruct_act here but also surplus white space

code/game/objects/structures/poddoor_assembly.dm
```diff
@@ -56,6 +56,14 @@
 			return
 		created_name = new_name
 
+	else if(item_used.tool_behaviour == TOOL_DECONSTRUCT)
+		if(!item_used.tool_start_check(user, amount=0))
+			return
+		user.visible_message(span_notice("[user] cuts apart [src]."), span_notice("You start to slice apart [src]..."))
+		if(item_used.use_tool(src, user, 4 SECONDS, volume=50))
+			to_chat(user, span_notice("You disassemble [src]."))
+			deconstruct(TRUE)
```
FalloutFalcon, line 65: /deconstruct_act

FalloutFalcon (dismissed):
I haven't gotten to see the angle grinder a lot so i cant comment too much on the balance aspect.
However the slowdown does seem like a bit of a big boost. Its meant to be very clunky.
I'm fine with test merges to see how it plays out.
I would strongly consider how long it should take relative to busting into a player ship along side its intended use of salvaging.

Erikafox:
> slowdown

A slowdown value of 1 is entirely asinine and entirely part of the reason no one uses angle grinders. Especially when plasma cutters are abundant and do the same thing.

code/game/objects/structures/poddoor_assembly.dm
Erikafox: not defined for this object

code/game/machinery/doors/poddoor.dm
Erikafox: not defined for this object

Apogee-dev:
> > slowdown
> 
> A slowdown value of 1 is entirely asinine and entirely part of the reason no one uses angle grinders. Especially when plasma cutters are abundant and do the same thing.

I mean, plasma cutters should probably be way less abundant than they currently are, honestly, otherwise there’s very little point in even _having_ the angle grinder. It’d be like a third of our ships getting hand drills and jaws of life (and being able to obtain them pretty easily if you dont start with em). why bother with basic tools at all when the upgrade is too easy to get

Erikafox:
> > > slowdown
> > 
> > 
> > A slowdown value of 1 is entirely asinine and entirely part of the reason no one uses angle grinders. Especially when plasma cutters are abundant and do the same thing.
> 
> I mean, plasma cutters should probably be way less abundant than they currently are, honestly, otherwise there’s very little point in even _having_ the angle grinder. It’d be like a third of our ships getting hand drills and jaws of life (and being able to obtain them pretty easily if you dont start with em)

Ill make a separate pr to map grinders in really quick

Apogee-dev:
> > > > slowdown
> > > 
> > > 
> > > A slowdown value of 1 is entirely asinine and entirely part of the reason no one uses angle grinders. Especially when plasma cutters are abundant and do the same thing.
> > 
> > 
> > I mean, plasma cutters should probably be way less abundant than they currently are, honestly, otherwise there’s very little point in even _having_ the angle grinder. It’d be like a third of our ships getting hand drills and jaws of life (and being able to obtain them pretty easily if you dont start with em)
> 
> Ill make a separate pr to map grinders in really quick

Good idea, I’d only leave plasma cutters on the REALLY specialized scrapping ships like the gecko.

PR #3586, opened 2024-10-18
Allows medical smocks to use their suit storage.

Aquidu:
## About The Pull Request
This PR adds a whitelist of medical items to the suit storage slots of Frontiersmen, NGR, Hardliners, and Nanotrasen smocks, as well as expanding the whitelist for the non-factional surgical apron.

Currently, I don't know of any items that work in the smocks aside from oxytanks and a PDA for the NT smock. Whereas other, more generic suits(such as winter coats and labcoats) would be more useful to wear by having a suit slot in them. This is a shame, because the four smock sprites are very pleasing, and compliment their faction uniform well. Pulling from labcoats and medical winter coats, with a few tweakings of my own, I have created a whitelist for the suit storage slot to be used on each smock, including but not limited to surgical tools, medical scanners, chemistry tools, and a few generic items.

NOTE: While working on this, I discovered a few things that may warrant attention. The Hardliner and Frontiersmen smocks are the "base" suit model for all the factional gear, meaning future edits to them that we don't want all their equipment to inherit may be difficult if we don't add a placeholder model. Additionally, in the future, it may be smart to put the faction AFTER the category of item it is, e.g: "/clothing/suit/medical_smock/nanotrasen" instead of what it is now, "/clothing/suit/nanotrasen/medical smock" for easier grouping or cross-faction changes.

## Why It's Good For The Game
Players will feel less inclined to forgo the cool-looking smocks for other clothes simply because they don't have suit storage. Makes medical inventory management slightly easier.

## Changelog

:cl:
add: Added a suit storage whitelist to the Nanotrasen, Hardliner, NGR, and Frontiersmen smocks. Expands the whitelist of the surgical smock.
/:cl:


FalloutFalcon (dismissed):
Put the note in the "about the pull request"
I would also make the list defines if they are used more then twice.
I also do agree with the part on not making the parent type based off of the faction for most instances..

meemofcourse:
fix ur checks pls

Aquidu:
> fix ur checks pls

i don't know what's causing them, i tried asking in the discord when it was happening but no one could figure out the issue, one of them isn't even in a file i edited

code/__DEFINES/clothing.dm
```diff
@@ -37,3 +37,32 @@
...
+    /obj/item/reagent_containers/pill, \
+    /obj/item/storage/pill_bottle, \
+    /obj/item/paper, \
+    /obj/item/melee/classic_baton/telescopic, \
+    /obj/item/toy, \
+    /obj/item/storage/fancy/cigarettes, \
+    /obj/item/lighter, \
+    /obj/item/tank/internals/emergency_oxygen, \
+    /obj/item/tank/internals/plasmaman \
+)
```
SomeguyManperson: Entries in the list look like they should be tabbed once with tabs, these are currently tabbed with spaces

code/__DEFINES/clothing.dm
```diff
@@ -37,3 +37,32 @@
 /// How much integrity does a shirt lose every time we bite it?
 #define MOTH_EATING_CLOTHING_DAMAGE 15
 */
+//Whitelist for the suit storage slot on medical suits
+#define MEDICAL_SUIT_ALLOWED_ITEMS list( \
```
SomeguyManperson: Not sure why this is here instead of in __DEFINES/inventory.dm where the rest are. You _should_ be able to just copy how those work I think

Aquidu:
fixed both

PR #3592, opened 2024-10-18
Simple human armor now draws its values from real armor

SomeguyManperson:
## About The Pull Request

Simple humans now pull armor values from armor, currently set as whatever they appear to be wearing
The armor_base variable has a vv intercept to update armor values automatically for bus

## Why It's Good For The Game

Intuitive that the guy wearing super heavy armor dies to AP rounds and not 3 shotgun shells

## Changelog

:cl:
balance: simple humans now have as much health as normal humans! This makes them weaker!
balance: simple humans now also benefit from their armor as much as normal humans! This makes them stronger!
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/human/human.dm
+++ b/code/modules/mob/living/simple_animal/hostile/human/human.dm
@@ -39,15 +39,37 @@
 	footstep_type = FOOTSTEP_MOB_SHOE
 
 	faction = list("hermit")
-	/// Creates an armor item to compare 
+	/// use this to match our armor values with the suit armor it is set to
 	var/obj/item/clothing/armor_base = null
 
 /mob/living/simple_animal/hostile/human/Initialize()
 	. = ..()
 	if(ispath(armor_base, /obj/item/clothing))
-		armor_base = new armor_base()
+		var/obj/item/clothing/temp = new armor_base()
+		armor = temp.armor.getList()
+		qdel(temp)
 
 /mob/living/simple_animal/hostile/human/getarmor(def_zone, type) //WE CLOWN IN THIS fake carbon/human. GET YOUR INTRINSIC ARMOR BACK TO /mob/living/simple_animal
 	if(istype(armor_base))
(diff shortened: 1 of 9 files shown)


generalthrax:
I got real confused for a second until I realized the old pr was closed for some reason. Huh.

code/modules/mob/living/simple_animal/hostile/human/human.dm
```diff
@@ -39,3 +39,15 @@
...
+
+/mob/living/simple_animal/hostile/human/Initialize()
+	. = ..()
+	if(ispath(armor_base, /obj/item/clothing))
+		armor_base = new armor_base()
+
+/mob/living/simple_animal/hostile/human/getarmor(def_zone, type) //WE CLOWN IN THIS fake carbon/human. GET YOUR INTRINSIC ARMOR BACK TO /mob/living/simple_animal
+	if(istype(armor_base))
+		return armor_base.armor.getRating(type)
+	return FALSE
```
MarkSuckerberg:
```suggestion
	///Steals the armor datum from this type of armor upon init
	var/obj/item/clothing/armor_base
	///Stored reference to the armor_base's armor datum, or null
	var/datum/armor/armor

/mob/living/simple_animal/hostile/human/Initialize()
	. = ..()
	if(ispath(armor_base, /obj/item/clothing))
		//sigh. if only we could get the initial() value of list vars
		var/obj/item/clothing/instance = new armor_base()
		armor = instance.armor
		qdel(instance)

/mob/living/simple_animal/hostile/human/getarmor(def_zone, type) //WE CLOWN IN THIS fake carbon/human. GET YOUR INTRINSIC ARMOR BACK TO /mob/living/simple_animal
	if(armor)
		return armor.getRating(type)
	return FALSE
```

Apogee-dev:
This is a good and necessary PR but we _really_ need to nerf basic Frontiersmen armor before this goes through. Those little shits are wearing bulletproofs with 60 bullet armor.

PR #3603, opened 2024-10-20
scatter laser changes

Jedi-Toothpaste:
## About The Pull Request
WIP: I want some feedback, 'slug' noises are **not final** I want to get the weapon balance right first mostly.

### NEW Laser Slug Mode!
Instead of switching your scatter-laser to a regular TG Laser it is now a ✨ slug ✨ which fires for 30 damage with standard AP.
**UPDATE**: Laser Slug now deals **20** Damage with an AP of 40, this keeps the TTK relatively the same, especially against armoured targets. Imagine it like being hit by a laser-shaped bat.
Has a delay of 0.9 Seconds between firing which gives it an almost identical time to kill (TTK) to the SL-204.
> _(don't question why I want to keep the slug mode similar it's more punishing when you miss okay it's balanced okay we can tweak this)_

Now uses the heavy laser projectile to help distinguish what kind of laser weapon your enemy is using.
> _down the line I can definitely get some unique sprites for the laser itself alongside a new scatter laser sprite maybe_ 🥺 

Also, it finally doesn't just say, "kill" when you switch modes!
image

New laser slug noise- bear in mind this is **not the final version**. _This one is bad, I know._

https://youtu.be/GyuE71jnlxE

> will update this with an edit when I get the final version

### REWORKED Scatter Mode!
Double the pellets, double the fun. With an increased variance for the pellets; hose down those corridors! You _go_ Girl.
With pellets up from **5** to **10**, you become a threat- however, they drop off just as hard as regular buckshot, so shoot carefully.
UPDATE: The pellets only fly 6 tiles forward, but if you're point-blank, it won't matter much.
image

Increased the base damage from 3->5 which makes you extremely scary, if you can get close enough.
> _it's not techshells though, they're still way more insane_

All these pellets come at a cost, however, you get approx. half the shots. (**6** shots with a Regular Cell, and **12** with a Superior Cell)
image
image

## Why It's Good For The Game
Currently, the average shiptester has two reactions to the scatter laser
The First: They do not pick it up.
The Second: They pick it up and switch mode to, 'kill' for a SL-204 called 'scatter laser'
This makes the scatter laser unique, and quite scary in some situations and hopefully not making it relegated to a glorified SL-204.

## Changelog

:cl:
add: new secondary 'slug' mode for the scatter laser (and other scatter lasers sub-types)
add: scatter laser to cargo, for you freaks who want to buy it
change: slug mode now uses cs-fire.ogg as its firing sound
del: removed old secondary mode for scatter lasers (and scatter laser sub-types)
balance: laser scatter mode now fires 10 pellets instead of 5
balance: laser scatter pellets now do 5 damage instead of 3
balance: laser scatter mode variance increased from 25 to 40
balance: doubled laser scatter mode energy cost (6 shots with normal cell, 12 with upgraded)
/:cl:


code/modules/projectiles/guns/energy/laser.dm
```diff
@@ -74,8 +74,8 @@
 
 /obj/item/gun/energy/laser/scatter
 	name = "scatter laser gun"
-	desc = "A laser gun equipped with a refraction kit that spreads bolts."
-	ammo_type = list(/obj/item/ammo_casing/energy/laser/scatter, /obj/item/ammo_casing/energy/laser)
+	desc = "A laser gun equipped with a refraction kit that spreads bolts. Also has a trigger to fire laser 'slugs'."
```
Erikafox:
```suggestion
	desc = "A laser gun mimicking the function of a shotgun, manufactured with an adjustable lens capable of projecting scattershot or slugs."
```

code/modules/projectiles/projectile/beams.dm
```diff
@@ -84,6 +84,11 @@
 	damage = 0
 	nodamage = TRUE
 
+/obj/projectile/beam/laser/slug
+	name = "laser slug"
+	icon_state = "heavylaser"
+	damage = 30
```
Erikafox: imo it should have some AP for being a more focused shot

Jedi-Toothpaste: If it had AP it would again be relegated to always being on kill (or in this case, slug) mode. Not that i'm unwilling to change this with time, but i'd like to see how players use it first. Then maybe buff it if it needs it.
Or I can reduce the damage and give AP which would make it basically the same, idk.

I am willing to try it out with AP & less damage.

Erikafox: AP and less damage would give it another interesting niche imo.

Jedi-Toothpaste: Will give it a shot when I push the damage fall-off to the scatter mode

generalthrax:
You should probably ask a Maintainer or Someone Else to give it a look, and maybe set this to draft. Seems incomplete

Jedi-Toothpaste:
> You should probably ask a Maintainer or Someone Else to give it a look, and maybe set this to draft. Seems incomplete

Yeah, done so now. Though, it is done asides the sound, which can just be put in with a placeholder and made better later, especially since i'm about to dive into a tasty 70 hour work week.

Jedi-Toothpaste:
I kinda wanna just get this Out There, the sound is fine and if needs be can be updated later

Jedi-Toothpaste:
auxmos check

FalloutFalcon (approved):

PR #3607, opened 2024-10-20
Adds inteq survival boxes.

Sadhorizon:
## About The Pull Request
Contents:
- inteq balaclava
- radio
- adv. o2 tank
- medipen

Every inteq role gets this.

## Why It's Good For The Game
Balaclavas are, according to their desc, used to conceal member identities. Kinda weird that most members don't get them!

## Changelog

:cl:
add: Added an inteq survival box.
/:cl:


Apogee-dev:
I wouldn't give every single member of an inteq crew an advanced oxygen tank in their survival box when normally only engineers get those. Frankly, I don't know why the CLIP and syndie boxes provide them either (hell, the CLIP boxes with balaclavas actually have _double tanks.)_ It looks like a pretty cut and dry example of power creep to me.

Also we have _got_ to refactor balaclavas to function like normal breath masks instead of gas masks.

thgvr:
I was gonna bring up they should be the standard if we don't want to increase the time of the blue small tanks

Martinpachu:
i think itd be good if they were standard the small blue ones are so pathetic

Apogee-dev:
i wouldn't be against making yellow tanks standard, but if we're gonna do that we should do it all at once instead of piecemeal (and the double tanks in that one survival box are still kinda ridiculous though)

Sadhorizon:
> It looks like a pretty cut and dry example of power creep to me.

Name a better duo than inteq and suffocating. I think they need it.

Apogee-dev:
I mean, to be fair the default blue emergency tanks in boxes are only intended to last until you can get to a safe area and/or a better source of oxygen, like one of the large tank dispensers every single ship has rather than used as an oxygen supply when disembarking. I suppose one reason we might not want to switch to yellow tanks as standard is it encourages people to use back tanks less often. We have very few items that can even begin to compete with backpacks for that slot as it is (we should make regular jetpacks only function in the actual back slot :clueless: )

PR #3609, opened 2024-10-21
Balaclavas Aren't Gas Masks

Apogee-dev:
## About The Pull Request

Addresses a longstanding pet peeve of mine. As originally implemented military balaclavas were just better gas masks. Now they just function as normal breath masks that can hide your identity, providing no passive gas protection nor hailer functionality. Also rewrites the description for the regular balaclava to no longer be a joke.

**This PR contains repaths, don't TM it alongside map PRs.**

## Why It's Good For The Game

actual gas masks should be useful enough to justify using them instead of balaclavas when gas is about

## Changelog

:cl:
balance: Balaclavas and half masks are now just breath masks, not gas masks
balance: Normal balaclavas can use internals too
fix: Balaclavas don't have a joke description anymore
iconadd: mask-up sprites for NGR balaclavas and masks
/:cl:


FalloutFalcon:
I do agree with this change but like the issue is irl you would just be able to layer these 2 together. I would honestly have a balaclava that has an integrated gas mask as a distinct item.

Apogee-dev:
> I do agree with this change but like the issue is irl you would just be able to layer these 2 together. I would honestly have a balaclava that has an integrated gas mask as a distinct item.

If you mean a half gas mask over a balaclava, sure, it'd work, although I don't want to for reasons elaborated on below. If you mean a full gas mask over a balaclava I’m not sure there’s actually any point since the balaclava would not be visible and gas masks already anonymize the wearer, so they the same core functionality (and more). 
The other issue is I wanted gas masks that actually offer passive gas/smoke protection to be normal-sized items. Stuff you find in armories and fire lockers, not a piece of equipment you’ll keep in your backpack or wear casually 24/7. The mere existence and accessibility of a small gas mask would kind of render those pointless outside of aesthetics, which kind of sucks and substantially reduces the credibility of gas and smoke as actual dangers. 
With this PR that’s actually the case, with the only notable exception being sec hailers…. which are mostly unused (and should be reworked or removed imho, since it’s not like the hailing function is useful in regular gameplay). That, and if you're gasproof you should look the part rather than looking like any other mook.

Apogee-dev:
Don't test merge this PR at the same time as any map file containing balaclavas of any description, or it _will_ cause broken tiles.

PR #3610, opened 2024-10-21
Remove Gunslinger Quirk

Erikafox:
## About The Pull Request
Removes Gunslinger quirk and related recoil behavior - revolvers are now flippable for all users

## Why It's Good For The Game
I don't think it's healthy for gun balance to have to factor in a 0pt quirk.

## Changelog

:cl:
del: Gunslinger trait
/:cl:


generalthrax:
Cutting the revolver fanfire and all the unique interactions of the Hunter's Pride lever actions? That was the big point of using the .44s over semi-autos for ages. The Absolution and Conflagration was hanging on these, as well

Erikafox:
> Cutting the revolver fanfire and all the unique interactions of the Hunter's Pride lever actions? That was the big point of using the .44s over semi-autos for ages. The Absolution and Conflagration was hanging on these, as well

Yeah that kinda blows. However - the balance and functionality of those guns should not be revolving around a quirk.
We can get data and look at buffing them if they underperform.

rye-rice (changes requested):
for the record the only reason this is a 0 point quirk is there is No perk system, and i also despite quriks, making it 0 points means it interacts with the quirks system the absolute least

post pref-rework it can likely made a perk

generalthrax:
Don't forget the Lever Action interactions as well

Erikafox:
> for the record the only reason this is a 0 point quirk is there is No perk system, and i also despite quriks, making it 0 points means it interacts with the quirks system the absolute least
> 
> post pref-rework it can likely made a perk

Coincidentally I don't think gun balance should have to factor in assorted perks. Not without a design document at the least.

Erikafox:
> Don't forget the Lever Action interactions as well

VSC is telling me I got every instance of TRAIT_GUNSLINGER so I'm feeling pretty good

generalthrax:
> VSC is telling me I got every instance of TRAIT_GUNSLINGER so I'm feeling pretty good

I mean making sure the lever actions retain their prior function like fanfire as per Rylie's requested changes
image

Erikafox:
> > VSC is telling me I got every instance of TRAIT_GUNSLINGER so I'm feeling pretty good
> 
> I mean making sure the lever actions retain their prior function like fanfire as per Rylie's requested changes image

This is all gunslinger related code. If we want to add it in at another point git has us covered. Also |where| in the requested changes.

rye-rice:
the purpose was so you can larp as a cowboy from a 70s western, not so much affecting gun balance in any meaningful way, since before HP guns from a pure balanced perspective are the worst weapons in the game, 

if theres no reason to sacrifice using the better guns to use the western guns stylishly, then im not sure if even having the quick rack and fanfire is even worth it at this point

and fanfiring being a thing only certain  characters can do i think has its charm, adds to the space cowboy blorbos. if its something everyone can do i think it ruins the charm of fanfire

ill delegate this to @Apogee-dev

Apogee-dev:
I don't think this is a good idea in the slightest, honestly. Removing it completely kills the viability of the revolvers and leverguns by killing fanfire; Giving _everybody_ fanfire means every single player gets an extremely powerful ability with no downside that dramatically buffs the cheapest guns in the game. The whole tradeoff with gunslinger is making the bad-to-underwhelming guns pretty decent in exchange for hamstringing your ability to use the high-end guns. I think this is genuinely best left unchanged, because both of those options are worse than the current state of things.

Erikafox:
> I don't think this is a good idea in the slightest, honestly. Removing it completely kills the viability of the revolvers and leverguns by killing fanfire; Giving _everybody_ fanfire means every single player gets an extremely powerful ability with no downside that dramatically buffs the cheapest guns in the game. The whole tradeoff with gunslinger is making the bad-to-underwhelming guns pretty decent in exchange for hamstringing your ability to use the high-end guns. I think this is genuinely best left unchanged, because both of those options are worse than the current state of things.

I don't think weapons should be balanced around a quirk

Apogee-dev:
> I don't think weapons should be balanced around a quirk

Your opinion is noted, and provides no counterargument against the negative consequences I listed out nor a better solution. Rye and I both respectfully disagree, and we've been doing most of the design and implementation legwork for gun balance and mechanics for at least the past two years. Gunslinger as a quirk is a patently awkward way of implementing the idea, but a better solution can't happen until the prefs rework. We should at _least_ wait until then to try reworking it.

thgvr:
IDK what a quirk vs perk would be in your eyes, they mean the same thing. The behavior should just be tied to the guns. Having guns just be better because of char creation isn't really something to go for

Erikafox:
image

PR #3629, opened 2024-10-24
Fixes searching with the orbit menu

zimon9:
## About The Pull Request

This PR modifies the search algorithm for points of interests so that the real name of a character, e.g. "Kristina Dryden" is what gets passed to the search function, instead of a character's "full name", e.g. "calm human woman". 

vf7sJJ034D

The sorting algorithm seems to take jobs as the sorting parameter, instead of real names. As of right now, the portion of code that causes this to occur still eludes me.

## Why It's Good For The Game

This felt like a minor oversight to me. It feels clunky to search for a character's name and have the search return no results because the search function had been looking for a character's _attributes_ instead. 

## Changelog

:cl:
fix: fixed the orbit menu search function so that names work again
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/Orbit/helpers.ts
+++ b/tgui/packages/tgui/interfaces/Orbit/helpers.ts
@@ -96,8 +96,8 @@ export const searchFor = (searchText) =>
 export const compareString = (a, b) => (a < b ? -1 : a > b);
 
 export const compareNumberedText = (a, b) => {
-  const aName = a.real_name;
-  const bName = b.real_name;
+  const aName = a.name;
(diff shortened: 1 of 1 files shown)


FalloutFalcon (commented):
hey i'm pretty confident that full_name is both of them but i will look into it
i thought i already fixed this

zimon9:
> hey i'm pretty confident that full_name is both of them but i will look into it i thought i already fixed this

image

From what I've seen using the VSC debugger, it appears that the game saves the full_name variable as the concatenated version of all of the attributes that make up a character, while the name variable saves the _actual_ name of a character.

PR #3632, opened 2024-10-24
Carpentry Tweaks -- Allows easels and cabinets to be crafted, and fixes various bugs with painting frames

zimon9:
## About The Pull Request

This PR adds the ability to craft easels and cabinets, and it fixes the spawned structure type when painting frames are used from being the generic type to the library type. This PR also fixes a bug when placing painting frames, as shown below.

| before | after |
|--------|-------|
|dreamseeker_tpNvZO4I62|dreamseeker_Yep9pLj1oV|

## Why It's Good For The Game

Cabinets and easels _should_ be craftable, so as to keep parity with other materials. It also provides an avenue to pursue more low-tech solutions when one is going for a more rustic aesthetic. 

Additionally, crafted painting frames have had a minor problem with not displaying saved paintings when one is placed in them because the type of frame that typically gets mapped is not the same type that can be crafted. This should fix that, _and_ a bug associated with pixel shifts for the frame structure when it's initialized.

## Changelog

:cl:
add: Added ability to make cabinets and easels
fix: fixed painting frame persistence and initialization bugs
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/sheets/sheet_types.dm
+++ b/code/game/objects/items/stacks/sheets/sheet_types.dm
@@ -127,6 +127,7 @@ GLOBAL_LIST_INIT(wood_recipes, list ( \
 	new/datum/stack_recipe("dresser", /obj/structure/dresser, 10, time = 15, one_per_turf = TRUE, on_floor = TRUE), \
 	new/datum/stack_recipe("picture frame", /obj/item/wallframe/picture, 1, time = 10),\
 	new/datum/stack_recipe("painting frame", /obj/item/wallframe/painting, 1, time = 10),\
+	new/datum/stack_recipe("easel", /obj/structure/easel, 4, time = 10),\
 	new/datum/stack_recipe("display case chassis", /obj/structure/displaycase_chassis, 5, one_per_turf = TRUE, on_floor = TRUE), \
 	new/datum/stack_recipe("wooden buckler", /obj/item/shield/riot/buckler, 20, time = 40), \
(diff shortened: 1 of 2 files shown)


MarkSuckerberg:
I think I fixed the saving in #1722, it's actually more the loading, since paintings that were hand-created are saved to "general" rather than "library". It's a little out of scope perhaps since it'd be a lot of map changes, but I think it'd be nice to just switch everything to be the "general" base type of paintings on maps and such so that we don't lose the paintings we have now if possible. Up to you if you want to try that, though.

zimon9:
> I think I fixed the saving in #1722, it's actually more the loading, since paintings that were hand-created are saved to "general" rather than "library". It's a little out of scope perhaps since it'd be a lot of map changes, but I think it'd be nice to just switch everything to be the "general" base type of paintings on maps and such so that we don't lose the paintings we have now if possible. Up to you if you want to try that, though.

You're entirely right, and I've amended my PR body with regards to what the PR actually does. 

As for switching the type of painting frame on maps, I _do_ like the idea, and I want paintings that might have been saved to "general" to be displayed, but at the same time, there are lots of really nice paintings in the "library" archive that would then end up not being displayed. There's also the problem of any future maps inadvertently using library frames again. 

One way to solve this that I can see is to manually merge the two repositories, whether that be to copy the paintings saved in "general" to "library" or vice versa. Whichever is preferred. 

I don't mind replacing the frames that are currently mapped to the generic type and reverting the change to the crafted frame, but I'd want paintings in the "library" repository moved over alongside this.

PR #3636, opened 2024-10-25
Edits the name of plasma and plastitanium shards

zimon9:
## About The Pull Request

Changes the name of plasmaglass shards from "purple shard" to "plasmaglass shard" and the name of plastitanium glass shards from "beige shard" to "plastitanium glass shard".

## Why It's Good For The Game

It does not fit with the current naming convention for items for these shards to be vaguely named based on their color. This change would help these shards make more sense.

## Changelog

:cl:
add: Added better names for some material shards
/:cl:


Apogee-dev:
plastiglass isn’t beige anyway, it’s orange. I’m starting to think people just have an obsession with calling everything beige

meemofcourse (approved):
works

PR #3642, opened 2024-10-25
removes some awaymission cruft

FalloutFalcon:
## About The Pull Request
tired of seeing it.

## Changelog

:cl:
del: Removed some ancient away mission cruft
/:cl:


Sun-Soaked (approved):
we should just Destroy the awaycode files and repath everything in them but this is a good step in the right direction

FalloutFalcon:
i thought about it. There is like. 3 things in there that are used, i didnt feel like doing it.

PR #3643, opened 2024-10-25
Downgrades python requirements

MarkSuckerberg:
## About The Pull Request
Upgraded these as it was blocking my ability to commit, not realising that the bootstrapped python isn't set up to get a specific version of python on linux hosts. sigh


Sun-Soaked:
waow. I'm so glad it passed checks and marg definitely didn't forcemerge it!!! thanks checks!!!!

PR #3652, opened 2024-10-26
2 small fixes

Erikafox:
:cl:
code: cauterization now checks for pain feeling traits instead of morphine
balance: morphine pills no longer addict and overdose you
/:cl:


code/modules/mob/living/carbon/carbon_defense.dm
```diff
@@ -1,6 +1,6 @@
 /mob/living/carbon/attackby(obj/item/W, mob/user, params)
 	var/obj/item/bodypart/BP = get_bodypart(check_zone(user.zone_selected))
-	var/has_painkillers = reagents.has_reagent(/datum/reagent/medicine/morphine, needs_metabolizing = TRUE)
+	var/feels_pain = (HAS_TRAIT(user, TRAIT_ANALGESIA) || HAS_TRAIT(user, TRAIT_PAIN_RESIST))
```
MarkSuckerberg: I'd probably call this `painless` or something instead, even `no_feel_pain` or something, since it's a bit incorrect as is. you can also just invert it probably but there's not really a point

PR #3694, opened 2024-11-02
Ammo Antics

Erikafox:
## About The Pull Request

Subtypes AP/HP/Rubber ammo for several calibres that didn't have it. adds it to the market in relevant cases.
adds tracker rounds to a few places. makes them work differently.
removes incendiary ammo from everything except shotguns

- [ ] I tested my pr

## Why It's Good For The Game

makes the ammo gameplay stuff available to more weapons. entertainment value

## Changelog

:cl:
add: rifle calibre and pdw calibre rubbershot/armor piercing/hollow point. check da outpost
add: tracker rounds for a few guns. check the black market :3
del: most forms of incendiary ammo
del: A bunch of clown and mime stuff tangentially related to guns

/:cl:


Apogee-dev:
I wonder how viable it would be to make the speed values you defined here modifiers on the cartridge’s base speed value rather than fixed values? I was looking over the code and noticed a few weird things here and there like match revolver bullets being faster than standard DMR bullets. it’d be nice if we could make certain guns in general have slightly faster or slower bullets than others while still having the cleaner, lower-maintenance code you made here for the cartridges

Erikafox:
> I wonder how viable it would be to make the speed values you defined here modifiers on the cartridge’s base speed value rather than fixed values? I was looking over the code and noticed a few weird things here and there like match revolver bullets being faster than standard DMR bullets. it’d be nice if we could make certain guns in general have slightly faster or slower bullets than others.

we'd have to tie it to a modifier like how bullet skins work. not opposed to doing it like such, though.

generalthrax:
A few select bullet types like .22lr hollow / rubber and .300 are half and half normal. Probably a quick fix but something you might wanna look at before We Become Real

generalthrax:
6.5x57mm only has half the ammo and its the wrong ammo type for both normal and trackers.

code/modules/projectiles/boxes_magazines/ammo_stacks/prefab_stacks/premade_sniper_stacks.dm
```diff
@@ -7,3 +7,78 @@
...
+
+/obj/item/ammo_box/magazine/ammo_stack/prefilled/a65clip/trac
+	ammo_type = /obj/item/ammo_casing/a65clip/trac
+	max_ammo = 5
+
+/obj/item/storage/box/ammo/a65clip/trac
+	name = "box of 6.5x57mm CLIP tracker ammo"
+	desc = "A box of standard 6.5x57mm CLIP tracker ammo."
+
+/obj/item/storage/box/ammo/a65clip/PopulateContents()
```
Erikafox:
```suggestion
/obj/item/storage/box/ammo/a65clip/trac/PopulateContents()
```

Sun-Soaked (changes requested):
document cruft removals in changelog

code/modules/projectiles/boxes_magazines/ammo_stacks/prefab_stacks/premade_smg_stacks.dm
```diff
@@ -1,24 +1,45 @@
...
+/obj/item/ammo_box/magazine/ammo_stack/prefilled/c46x30mm/hp
+	ammo_type = /obj/item/ammo_casing/c46x30mm/hp
+	max_ammo = 20
+
+/obj/item/storage/box/ammo/c46x30mm/hp
+	name = "box of 4.6x30mm HP ammo"
+	desc = "A box of standard 4.6x30mm HP ammo."
+	icon_state = "46x30mmbox"
+
+/obj/item/storage/box/ammo/c46x30mm/ap/PopulateContents()
```
generalthrax: This should be hp instead of ap

Sun-Soaked (approved):
looks good

Erikafox:
yaaaaay

PR #3697, opened 2024-11-02
Legion Core Fluff Change + Cloneloss

Erikafox:
## About The Pull Request

rewrites some regenerative core fluff

regen cores now do some cloneloss on the end of the effect

## Why It's Good For The Game

I think this is more flavorful than free heals

## Changelog

:cl:
balance: legion cores now slowly build up cloneloss when you use them.
/:cl:


code/datums/status_effects/buffs.dm
```diff
@@ -414,6 +414,8 @@
 
 /datum/status_effect/regenerative_core/on_remove()
 	REMOVE_TRAIT(owner, TRAIT_IGNOREDAMAGESLOWDOWN, id)
+	to_chat(owner, span_warning("The tendrils of the regenerative core sink into your flesh, leaving dark markings where they dive."))
+	owner.adjustCloneLoss(10)
```
thgvr: this is pretty much unhealable. not a fan

Zevotech: 5 clone seems more reasonable but also ground puce crystals heal cloneloss + ghettoing cloneoxadone isnt that hard. burn medkit has rhigoxane which can activate it easily.

Apogee-dev: Player knowledge of the ways we do have to heal clone damage is just low because clone damage is normally a complete non-factor. if it were to become relevant (as in this PR), they’d pick it up quickly enough.

Sun-Soaked (approved):
I was hoping for like. Low effect for using a single one and Escalating consequences for using a lot of them in tandem. but this is good for now

PR #3708, opened 2024-11-05
Replaces Rubbershot on the Kilo and Scarab with Buckshot

generalthrax:
## About The Pull Request

Says on the tin.

## Why It's Good For The Game

Every round on both these ships I've seen the boxes either go unused or recycled. What am I gonna shoot them at? Goliaths? My own crew?

## Changelog

:cl:
balance: Replaced rubbershot on the Kilo and Scarab with buckshot
/:cl:


Jedi-Toothpaste:
can we add meteorshot instead

Sadhorizon:
Kilo sure, but on the scarab, I'd rather you don't do this.
It's not exactly meant to get a lot of ammo roundstart, and I think a non-lethal option is kind of neat for them, even if unused - you know, just in case.

I'd rather you remove it completely tbh.

generalthrax:
I'm just saying its a mining ship, for drilling and killing fauna. I don't think it makes a lot of sense to have it. I could probably just get rid of the box, I think theres a spare anyways.

PR #3713, opened 2024-11-06
Adds more trauma for your blorbos

SomeguyManperson:
## About The Pull Request

Adds trauma to organic limbs, which increases the damage of any attack to that limb by up to 150%
Trauma is gained through fast and/or cheap healing methods: patches, tend wounds surgery, and legion cores, as well as the super fast healing stacks found on the black market. Current numbers are around 1:2 trauma per health, with tend wounds being 4:5. Numbers are VERY likely to change

Currently, trauma is reduced by waiting, with 1 point being lost per 5 seconds. Eating enough healthy food doubles this, while starving enough will halve it. I want to add some fancy surgical equipment or machine that significantly boosts recovery for medships at some point, too, if not anything else. Oh right you can also just. remove the limb. and get a new one. I look forward to seeing the moth of theseus.

## Why It's Good For The Game

Attempts to balance out fast/slow and expensive/cheap healing by making things that instantly or quickly heal a lot of damage have a downside. You _can_ go from 0 to 100 in short order, but you'll be sent from 100 to 0 in even shorter order after.

## Changelog

:cl:
add: characters can now take a long-term form of damage to their limbs called trauma. A traumatized limb takes bonus damage to reflect vulnerability while it is recovering.
add: trauma is gained when healed with fast-acting or cheap equipment: tend wounds surgeries, patches, legion cores, and brute packs/ointment. Trauma will naturally decay over time. Eating healthy food can increase the decay speed.
balance: legion cores now have a short delay for use. This delay is halved if you are applying it to someone else. Friends do have benefits.
balance: using several legion cores while moderately traumatized will cause an infection, which is bad for you.
/:cl:


tantspants:
when you say infection, do you mean the legion skull infection or is it something new? Either way, seems like a cool idea.

SomeguyManperson:
> when you say infection, do you mean the legion skull infection or is it something new? Either way, seems like a cool idea.

Skull infection

Jedi-Toothpaste:
You should also check for the medibot. Since it does it's own tend-wound surgery technically.

PR #3717, opened 2024-11-08
You can no longer see clothes through Blood Red Hardsuits (& others)

generalthrax:
## About The Pull Request

Used to be able to see what someone was wearing underneath their hardsuit 

## Why It's Good For The Game

Oversight probably. I don't think the NGR / Hardliner / Ramzi hardsuits are supposed to have this var applied to them to begin with

## Changelog

:cl:
fix: You can no longer see through Marauder hardsuits
/:cl:


thgvr:
Leftover from scarlet suit maybe?

PR #3720, opened 2024-11-08
Severely Reduces Fishing Missions

MarkSuckerberg:
## About The Pull Request

Judging by the stats, nobody seems to think fishing missions are terribly worthwhile as they are now for multiple reasons. They're extremely common and don't pay that well, despite the setup and time required to do them. I feel like leaving a rarer, better-paid one would be fine, so that's what I did. I drastically reduced the total weight of all fishing missions from 11 to 2, so they should take up a lot less slots now. I also only kept the non-specific fish mission, as the specific fish ones seemed to be mostly based on RNG, let alone the rarer ones.

## Why It's Good For The Game

It's still not a very interesting mission objective, but it might inspire people to go try out fishing when they're bored, or something. Getting rid of annoying, rarely-done missions means people have to sift through less to get to missions they actually want.

## Changelog

:cl:
del: Most fishing missions. Now only the cooking one, that takes 1-3 of any fish, has been left.
tweak: The fish cooking mission now pays twice as much as it used to, 500 credits per fish. It's also less likely to appear as a mission.
/:cl:

### DIFF
--- a/code/modules/overmap/missions.dm
+++ b/code/modules/overmap/missions.dm
@@ -1,9 +1,12 @@
 /datum/mission
 	var/name = "Mission"
 	var/desc = "Do something for me."
-	var/value = 1000 /// The mission's payout.
-	var/duration = 30 MINUTES /// The amount of time in which to complete the mission.
-	var/weight = 0 /// The relative probability of this mission being selected. 0-weight missions are never selected.
+	/// The mission's payout.
+	var/value = 1000
+	/// The amount of time in which to complete the mission.
+	var/duration = 30 MINUTES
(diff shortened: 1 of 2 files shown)


Anticept:
The other reason people don't touch fishing missions is that sometimes it STILL won't let you actually catch fish. It's very rare now but still happens.

100% agree on not taking them out, it's neat as a little fluff.

Joel-682:
I think it may be better to have mission "categories", and have one mission generate per category instead of having them all fight for presence on the server. Really sad when your deep drilling ship can't deep drill for money because space RNGesus doesn't like you

Sun-Soaked (approved):
everyone say thank you marg

PR #3726, opened 2024-11-10
NGR Turret ammo swap

Erikafox:
:cl:
balance: medium and light ngr turrets swapped ammo. mediums should be more effective now.
/:cl:


Sun-Soaked (approved):
more like new stupid republi[THIS USER IS UNDER INVESTIGATION BY THE REVOLUTIONARY GUARD]

PR #3728, opened 2024-11-10
Legion virus feedback

SomeguyManperson:
## About The Pull Request

Makes the legion virus stage messages player-facing instead of functionally nonexistent since my dumbass somehow didn't realize they were tied to the stage increase probability. which is zero. yeah.
Also getting infected now shoves a message into chat which you will probably miss anyways since it's surrounded by getting bitten and having your bones melt. You WILL be able to tell if you have a new friend though.

## Why It's Good For The Game

I thought people were dying because the messages weren't providing enough feedback or something and it turns out they were providing a lot more than not enough feedback since they were providing NONE!

## Changelog

:cl:
code: legion virus stage messages now properly display themselves. instead of not doing that.
code: legion infesting people now give them a direct chat message. Say hello to your new friend! They might be sticking around your body longer than you will.
/:cl:


Sun-Soaked (approved):
it should probably make a unique noise or smth 
but this is good for now

PR #3727, opened 2024-11-10
increase plasmacutter price. reduce anglegrinder price

Erikafox:
:cl:
balance: plasma cutters r more expensive.
balance: angle grinders are cheaper, and come with a bowman headset.
/:cl:

PR #3737, opened 2024-11-11
Angle Grinders are more Efficient than Plasmacutters

generalthrax:
## About The Pull Request

Makes angle grinders overall faster than plasmacutters at taking down walls and structures, and taking a fair bit less charge to boot. Slightly decreases the time to deconstruct certain objects like safes and blast doors by a negligible amount.

## Why It's Good For The Game

Bulky as hell and no better than plasmacutters other than cost (as discussed in https://github.com/shiptest-ss13/Shiptest/pull/3727). I think it'd be interesting that angle grinders being a very efficient and quick method of shipbreaking (which has little to no in game benefit, mind you) compared to the plasmacutter being a compact, specialist tool for breaching would make buying either worthwhile.

## Changelog

:cl:
balance: Angle Grinders now take less charge and are much quicker at deconstructing walls
balance: Plasmacutters are now marginally slower, hardly noticeable. 
balance: Blast doors and safes are now very slightly faster to cut open.
/:cl:

### DIFF
--- a/code/game/machinery/doors/poddoor.dm
+++ b/code/game/machinery/doors/poddoor.dm
@@ -64,7 +64,7 @@
 			electrocute_mob(user, get_area(src), src, 1, TRUE) //zorp
 			close()
 		to_chat(user, span_notice("You start to cut [src] apart"))
-		if (W.use_tool(src, user, 15 SECONDS, volume = 75))
+		if (W.use_tool(src, user, 10 SECONDS, volume = 75))
 			deconstruct(TRUE)
 
 /obj/machinery/door/poddoor/examine(mob/user)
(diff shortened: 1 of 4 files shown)


Apogee-dev:
eh, i liked the idea of plasmacutters being a straight upgrade limited by availability more than the two being sidegrades- it means salvage specialists like the Derecho can get a really good tool that everyone else would have to work their ass off to buy. should just move plasmacutters to the black market and explain it by them just being not very widely available technology that requires special import (or thievery from the few governments and corporations that have them) or something.

generalthrax:
I'm not sure we should have people work their asses off to buy something with a minimal payoff in the gameplay loop. Scrapping / Shipbreaking gives you not a single credit for your work.

Thera-Pissed:
Could keep plasma cutters as a sidegrade and have advanced plasma cutters be a straight upgrade, maybe?

Apogee-dev:
I dunno, shipbreaking _is_ the only way to get resources in the field if you dont have a refinery. That’s probably worth something.

PR #3738, opened 2024-11-11
Frontiersman Vests Repathed to be like Everyone Else (Hits Simplemobs Too)

generalthrax:
## About The Pull Request

Repaths Frontiersmen armour vests to be normal armour vests (35 bullet armour) instead of bulletproof vests (60 bullet) armour. I'm 90% sure nothing testmerged has Frontie armour in it so this should be fine. 

Does not repath Frontier X11s because all X11s are bulletproof too.

Decreases the price of the blackmarket Frontie armour to reflect this.

## Why It's Good For The Game

This is more so intended for Frontiersmen armoured simplemobs that are fully armoured (frontier body-wide armour is 75 compared to this 60), meaning you do not have to fight 8 full-body armoured 60 bullet armour Monsters on every ruin (that most likely have assault rifles or shotguns that They Do Not Drop). 

It's also a little funny to me that a worn-out pirate vest is significantly better than the faction vests of every single faction with a stake in the Frontier.

## Changelog

:cl:
balance: Frontiersmen vests now count as normal armour vests and not bulletproof
balance: Frontiersmen simplemobs count this for their armour values
balance: Frontie armour vest in the blackmarket cost decreased
/:cl:


Sun-Soaked (approved):
should be good

PR #3739, opened 2024-11-11
Full Auto Mech Weaponry and Refresh

Gristlebee:
## About The Pull Request

Adds a component to facilitate full auto for certain mech weapons.
Ballistic mech weapons can now be set to eject casings.
Some general sanity changes for mech firecode.

The UMG mech machine gun is now full auto instead of burst fire. Max ammo capacity was cut in half to 600. Ammo boxes now hold 200 and cost 500 at the outpost.
Mech Scattershot now holds 12 rounds with a max ammo capacity of 72. Now fires 7 pellets up from 4, but is subject to damage falloff. Ammo boxes hold 24 rounds.
Mech disabler and light laser merged together into the Downpour energy carbine, which is fully automatic and can swap modes from the mech menu.
Mech heavy laser now fires an emitter blast instead.

Video Demonstration: 

https://github.com/user-attachments/assets/caf9acde-fa0d-4cee-8e59-4ff8446ca02c

## Why It's Good For The Game

Mech weapons aren't terribly impressive for how big of an investment getting a combat mech is. These additions and changes should make them more fun to use and interesting to see. And really. Big automatic mech machine gun is cool as hell.
## Changelog

:cl:
add: Mech Full Auto
add: Mech scattergun now has 7 pellets, and damage falloff.
add: Mech guns can now eject casings.
add: Solaris fires an emitter blast.
add: Downpour energy carbine
del: Immolator laser, Peacemaker disabler
balance: Ammo capacity for UMG and Scattergun mech weapons. UMG ammo now costs 500 at the outpost.
/:cl:


Anticept:
The UMG was prior discussed with apogee.

That thing is EXTREMELY strong. If you're allowing full auto, you really should reduce the damage.

It's basically an I win button.

generalthrax:
If you look closer I think it's actually a lower ttk than the burst. Since it's three bullets hitting in quick succession compared to this.

Gristlebee:
> The UMG was prior discussed with apogee.
> 
> That thing is EXTREMELY strong. If you're allowing full auto, you really should reduce the damage.
> 
> It's basically an I win button.

When you're putting down anywhere from 15-20k~ to buy and equip a mech, it can stand to be a bit stronger.

And IMO, it's too early to say for certain. I'd like to see how it performs on live with a TM first. If it does end up being too strong and an instant win in practice, I'm more than happy to tone it down.

Apogee-dev:
The large bullet casing is a bit unnecessary, in that the exosuit ballistic weapons fire normal ammo- the UMG-2 is described as a light machine gun and the LBX-10 could be like.... 8ga or something. Neither's actually something normal weapons couldn't use, and explicit cartridges are just about always better than abstract.

honestly not changing that part of the descriptions when i renamed the exosuit weapons was a mistake

(also also, never use the word "mech" in descs and names, always use "exosuit")

Anticept:
> If you look closer I think it's actually a lower ttk than the burst. Since it's three bullets hitting in quick succession compared to this.

The damage and armor pen haven't changed, and the ROF looks faster than burst to me. From his video, he was shooting center mass too on armored targets.

/obj/projectile/bullet/lmg
	damage = 25
	armour_penetration = 40

That's the damage values.

Now that said, I agree that the umg did need a little bit of a buff, apogee was thinking of it being okay to add a couple rounds to the burst, but that was the concern with that high armor pen. Two rounds will kill a limb and you're not going to find enough armor to prevent it.

Let's see what happens in TMs!

code/game/mecha/equipment/weapons/weapons.dm
```diff
@@ -326,31 +368,43 @@
...
-	projectiles = 40
-	projectiles_cache = 40
-	projectiles_cache_max = 160
-	projectiles_per_shot = 4
+	equip_cooldown = 10
+	projectile = /obj/projectile/bullet/pellet/scattershot
+	projectiles = 12
+	projectiles_cache = 24
+	projectiles_cache_max = 72
+	projectiles_per_shot = 7
```
SomeguyManperson: Each shot spends projectiles_per_shot so you'll get 1 shot with 7 rounds and 1 shot with 5 if you don't reload every time you fire

Gristlebee: The one casing var accounts for that. So instead of consuming 7 bullets per trigger pull it consumes 1 since that's a bit more logical since it comes out from a single casing.
image

FalloutFalcon (approved):
Looks good for what its trying to do!

PR #3751, opened 2024-11-13
Fixes runtimes from mobs smashing walls

Gristlebee:
## About The Pull Request

Title.

## Why It's Good For The Game

Runtimes are bad.
## Changelog

:cl:
fix: Fixes mobs breaking walls causing runtimes.
/:cl:

### DIFF
--- a/code/game/turfs/closed/_closed.dm
+++ b/code/game/turfs/closed/_closed.dm
@@ -319,7 +319,8 @@
 /turf/closed/attack_animal(mob/living/simple_animal/M)
 	M.changeNext_move(CLICK_CD_MELEE)
 	M.do_attack_animation(src)
-	if((M.environment_smash & mob_smash_flags))
-		playsound(src, 'sound/effects/meteorimpact.ogg', 100, TRUE)
-		alter_integrity(-400)
-		return
+	if(isclosedturf(src))
+		if((M.environment_smash & mob_smash_flags))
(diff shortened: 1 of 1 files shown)


code/game/turfs/closed/_closed.dm
```diff
@@ -319,7 +319,8 @@
 /turf/closed/attack_animal(mob/living/simple_animal/M)
 	M.changeNext_move(CLICK_CD_MELEE)
 	M.do_attack_animation(src)
-	if((M.environment_smash & mob_smash_flags))
-		playsound(src, 'sound/effects/meteorimpact.ogg', 100, TRUE)
-		alter_integrity(-400)
-		return
+	if(isclosedturf(src))
+		if((M.environment_smash & mob_smash_flags))
```
Anticept: Are the double parentheses needed?

Gristlebee: no, they arent. Good catch.

MarkSuckerberg:
You were on the right track, but the issue went a little deeper than this. [I fixed it here](https://github.com/shiptest-ss13/Shiptest/commit/685ee38568a82be2b3d82790680482990e6953f6#diff-188c259fb70a7777da917998c58e45813cdefca45a3a6a841bdd1ffa6fa645caR137-R143), where the reason this happens is that the turf gets destroyed while the proc is still running, which means that it still runs code even though it's now a floor instead of a wall. It's very weird that we have to deal with it, but hey :P

PR #3777, opened 2024-11-16
Fixes the Listening Post Server Monitor + Replaces a Frontier Suit with a Softsuit

generalthrax:
## About The Pull Request

The Telecomms Monitor board on the listening post doesn't actually work. It's just a maintenance computer. The intended thing to properly use the listening setup is a Server Monitor, that lets you see the contents of the server.

Replaced the singular suit on the ruin with a Soft suit

## Why It's Good For The Game

Fixes good

## Changelog

:cl:
add: Added a soft suit to the Listening Post
fix: Fixed the Listening Post Comms Monitor to be the right type
/:cl:

### DIFF
--- a/_maps/RandomRuins/LavaRuins/lavaland_abandonedlisteningpost.dmm
+++ b/_maps/RandomRuins/LavaRuins/lavaland_abandonedlisteningpost.dmm
@@ -1232,7 +1232,9 @@
 /turf/open/floor/plasteel/dark,
 /area/ruin/unpowered/listening_post/commons)
 "vJ" = (
-/obj/machinery/suit_storage_unit/mining/eva,
+/obj/machinery/suit_storage_unit/inherit,
+/obj/item/clothing/suit/space/syndicate,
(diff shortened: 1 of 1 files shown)


firebudgy:
This ruin isn't actually supposed to be Ramzi owned, the suit in the cabinet is supposed to be from the other crewmember that died and got buried outside during the ICW.

generalthrax:
RIP. My pirate fantasies go unfulfilled another day. The consoles finally fixed though. Only took like 3 months.

Burning02:
> This ruin isn't actually supposed to be Ramzi owned, the suit in the cabinet is supposed to be from the other crewmember that died and got buried outside during the ICW.

Wouldn't it make more sense for it to be a softsuit instead then? Perhaps even the surplus one?

firebudgy:
I was trying to figure out github UI. Yeah, could you change it to a syndicate softsuit? It being a frontier suit was something I overlooked.

generalthrax:
Yeah I'll do that

PR #3784, opened 2024-11-16
Bandage lifespan no longer scales with bleeding speed

SomeguyManperson:
## About The Pull Request

Bandage lifespan is now a flat 1 point = 1 tick of bleed staunching instead of also being reduced by the current bloodloss speed

## Why It's Good For The Game

Didn't scale well with higher speeds of bleeding, effectively making them unmanageable since the rate of bleed healing is balanced around casual bloodloss and not ranked and competitive bloodloss

## Changelog

:cl:
balance: bandages will no longer fall off faster with more bleeding
/:cl:

### DIFF
--- a/code/datums/components/bandage.dm
+++ b/code/datums/components/bandage.dm
@@ -41,7 +41,7 @@
 	SIGNAL_HANDLER
 
 	var/obj/item/bodypart/heal_target = parent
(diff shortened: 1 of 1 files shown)


MarkSuckerberg (approved):
I always wondered why bandaging just didn't work sometimes

PR #3787, opened 2024-11-16
New Ruin: Tesla Lab

Erikafox:
## About The Pull Request
Adds a new ruin to the iceplanet ruin pool - Tesla Lab
<details><summary>Screencaps</summary>
<p>

image
StrongDMM-2024-11-16 15 39 07
StrongDMM-2024-11-16 15 39 08

</p>
</details> 

## Why It's Good For The Game
This is an very hard ruin filled with nefarious tricks and a decent bit of loot. I want it to challenge people.
## Changelog

:cl:
add: CLIP has lost contact with a facility in the system. Go find it. Go see what happened. I dare you.
/:cl:


Gristlebee (changes requested):
- since these are secure containment areas, I think it'd make sense if they had blastdoors for the windows. The tesla observation room also has no lighting.
image
image
- these are all active turfs. Use planetary atmos subtypes please. The room also has no lighting.
image
- no lighting in the jani closet or toilet.
image
- there's no lighting in this washroom as well.
image
- the holofields in the subshuttle are not anchored. Use the roundstart subtype.
image
- Exterior windows should have shutters I feel.
image
- same issue with the compound entrance holofields.
image
- the railing overlays on top of the medkits in the subshuttle medbay
image
- missing a wire here between subshuttle holofields and the SMES
image
image
- the compound holofields aren't properly wired either.
image
  - The compound entrance holofields and subshuttle portside holofields share the same IDs.
- the office has no buttons for the door and windows shutters.
  image
- the turrets on the subshuttle shoot the frontiersmen simplemobs as they dont have the frontiersmen faction set.
- the turrets wont target heads of staff.
image
- the tapes in here don't play back properly. They start, but never begin playing back the recording.
image
- this should be a proper cycle airlock if you're planning on making the tesla room airless.
image
- the tesla anomaly engine is cool, but doesn't actually generate much in the way of power, and with such a large ruin it tends to burn through all it's stored power fast which is a concern. Maybe put some more SMESes with extra charge, though if you have another solution in mind you want to go with that's fine.
image
- the blood freezer is empty. I notice you set the populate to false, so maybe toss around some empty blood bags to imply they used up the blood so there's a bit more substance too it.
image
- Another empty crate here in the cargo bay. Put something in, or have it start open to imply it's been looted like the other crates.
image
- the desynchronizers are fun and it's nice to see them actually used. I'd say cut down the amount mapped to one or two though.
image

Gristlebee (changes requested):
- You added shutters to the tesla room like I asked in the previous review, but not the static anomaly room. Is that intentional?
image
- You sealed off the room, which does solve the issue of active turfs, but does limit the amount of the ways you can enter the ruin and makes it a lot more linear. I do think it would be more interesting if you had it be open like before but with planetary turfs. There's also still no lights from what I can see.
image
- Shutters and buttons in the office also have the wrong ID.
image
image
- Stray light in the SMES room
image
- ~~Cycle airlock here as I asked for in the previous review unless you have a reason as for otherwise.~~ I'm blind, there is one.
image
- You don't seem to have changed the faction on the turrets on the subshuttle from what I can tell so they're still gonna target the frontiersmen simple mobs around.
- Doing a check on the loot, the amount of frontiersmen that will drop their guns is concerning. I'd say set 4 of the shotgunners, 3 of the mosins, 1 of the SKM guys to neuter so they dont drop guns and the rest can drop their guns. I do get you want it to be a ruin with good loot for the challenge, but this is a lot of guns as is right now.

thearbiber:
Please keep in mind those claymores are capable of one shotting players and deleting their gear, while also being very small, hard to see, and not able to defuse/prevent like landmines.

Erikafox:
#3817

thearbiber:
waow

Ratvarr:
> Please keep in mind those claymores are capable of one shotting players and deleting their gear, while also being very small, hard to see, and not able to defuse/prevent like landmines.

defusal _is_ possible, I couldn't tell you how though. It's in the crate you buy them from on a page

Erikafox:
yeah apparently you can screwdriver them

Burning02:
On top of that, they quite literally break your gun/s. You'll be unable to load it, fire it, or even remove/add attachments.

Ratvarr:
also: the Static Syncroniser is not a good item probably to have on this because it allows someone to literally delete themselves temporarily. also it has it's **own** bugs and issues potentionally

Erikafox:
It's a cool item that's appearing in like two spots on one ruin. (That gives you brain damage for using it)

Gristlebee (dismissed):
Feedback after testing on live:
---

- This hallway has too many anomalies. Keep the one in the center of the hall and cut the other two.
image
- I'd say either cut this static, move it further in the hall into the lab, or reduce it's effect range. It feels redundant to me already so close to the other anomaly in containment, and the range brain damage effect is enough to get through the firelocks, which you have to cut through to get to the anomaly, which means nearly guaranteed brain damage.
image

- Increase the spacing on the turrets. Players running through the ruin found them a bigger threat than the actual mobs. They're well dug in, which is fine. But the clustering on them makes them difficult to approach. These two right are really close and a particular offender.
image
- In the same vein, cut the sandbags here which were originally meant for more turret emplacements. The area is already being covered by the subshuttle turrets, which does the job.
image
- I think it would be good to have a button on the exterior of the subshuttle doors from a level flow aspect, so there's a natural way in and players dont have to cut through.
image
- The desynchronzers need a much shorter max duration. I notice you already set them a max of a minute at 600 max duration, but I feel that's still too long. Set the max duration to 15 seconds.
- I still think this ruin needs a culling on the amount of loot after seeing it run through on live, particularly the amount of guns. All the Frontiersmen don't need to drop their guns when there's already some strong weapons mapped in. One SKM, one shotgun and three mosins can be dropped from the frontiersmen and the rest should be set to neuter. The mauler and flamers dropped from the admiral and flametroopers can be left as is and don't need to be touched.
- Cut the claymores in the armory down to one if you're going to have the Hammer here.
image

MarkSuckerberg:
justice for doctor pills

Latentish (dismissed):
image
You have an open parentheses somewhere.

FalloutFalcon (approved):

PR #3807, opened 2024-11-19
Minor Miskilamo Redecaling + Other Stuff

generalthrax:
## About The Pull Request

Minorly re-decals the Miskilamo line of ships. Mostly clearing out that sharp white against the grey / black that doesn't look that good. Kept it as consistent as I could throughout the three ships. I also did a few extra, smaller decaling in certain areas like crew quarters or areas with those weird grimy tiles.

Combines cryo and crewquarters into one area and combines the bathroom, washing machine room, and central hallway into one area as well for the Mudskipper.

Replaced the circular saw (angle grinder) with a plasmacutter.

Replaced the decaying armour vests with normal ones on the Kilo and Shetland.

Axes the airless turfs on the Kilo, and streamlines the big 2x3 airlocks on the Mudskipper and Shetland.

image
image
image
image
image

## Why It's Good For The Game

This pr does very little other than getting rid of those white decals which I felt weird to me. Most everything else was unchanged.

Mudskipper area stuff was cut down a bit because the thing chugged as much power as a Riggs at half the size, on one pacman to boot. I don't think a room needs a APC for a washing machine and one lightbulb. It's a salvage ship so its probably fair to have a plasmacutter and angle grinder.

The presence of the Shitty armour vests felt like an odd choice. They just sit there eternally, worn by no one. May as well just have normal ones. Not like they're that big of an advantage in the grand scheme of things.

## Changelog

:cl:
del: Removed the white decaling from Miskilamo ships (Mudskipper, Kilo, Shetland)
balance: Mudskipper is now more compact on areas and has a plasmacutter
balance: Kilo and Shetland now have normal armour vests
fix: Kilo no longer has airless tiles under the prechargers, and airlocks should Work Better
/:cl:

### DIFF
--- a/_maps/shuttles/independent/independent_kilo.dmm
+++ b/_maps/shuttles/independent/independent_kilo.dmm
@@ -39,7 +39,7 @@
 /obj/structure/table_frame,
 /obj/item/shard,
 /obj/effect/decal/cleanable/glass,
-/turf/open/floor/plasteel/tech,
+/turf/open/floor/plating,
 /area/ship/engineering)
 "ay" = (
 /obj/effect/decal/cleanable/glass,
@@ -71,7 +71,7 @@
 /obj/structure/railing{
 	dir = 8
 	},
-/obj/effect/turf_decal/corner/opaque/neutral/half,
+/obj/effect/turf_decal/corner/opaque/grey/half,
 /turf/open/floor/plasteel/dark,
 /area/ship/hallway/port)
 "aS" = (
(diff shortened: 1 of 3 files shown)


Apogee-dev:
Add ship numbers to the Mudskipper and Kilo while you’re at it- same spot as the shetland (aft starboard near the thrusters) if you can. Might need to shrink the kilo’s engineering window by a tile to fit.

Apogee-dev:
You accidentally placed some double catwalks on the kilo- fix that and we should be good

generalthrax:
think i got it

_maps/shuttles/independent/independent_kilo.dmm
```diff
@@ -179,7 +181,10 @@
 /obj/machinery/atmospherics/pipe/simple/supply/hidden/layer2{
 	dir = 5
 	},
-/turf/open/floor/plasteel/dark,
+/obj/effect/turf_decal/corner_techfloor_grid{
+	dir = 10
+	},
+/turf/open/floor/hangar/plasteel,
```
Sadhorizon: You seem to have accidently used hangar tiles on all three ships. Those have planetary atmos and probably shouldn't be on the ships.

PR #3813, opened 2024-11-20
Saw off guns with Angle Grinder

generalthrax:
## Why It's Good For The Game

Makes sense and you can do it with the dingy little circular saw

## Changelog

:cl:
add: You can saw off guns with the Angle Grinder
/:cl:


code/modules/projectiles/gun.dm
```diff
@@ -1120,6 +1120,7 @@
 GLOBAL_LIST_INIT(gun_saw_types, typecacheof(list(
```
Erikafox, line 1120: \>can't do it with a chainsaw
why even live

SomeguyManperson: This is the snowflake list chainsaws work because they count as saws

PR #3846, opened 2024-11-27
Ship renaming now properly abides by the renaming timer

meemofcourse:
## About The Pull Request
When you rename a ship, but there's a cooldown, the ship properly identifies this in the code for the overmap datum and for the helm console, but the latter doesn't actually do anything about it, which leads to a mismatch in names.

## Changelog

:cl:
fix: Ships now properly follow the rename cooldown
/:cl:


MarkSuckerberg:
wait huh. it shouldn't matter that the UI data gets updated because the ship's name never actually changes

MarkSuckerberg:
what the HECKLL the cooldown doesn't even work properly. Grahhhhhh

meemofcourse:
mark fixed this in another PR so this is moot

PR #3867, opened 2024-12-02
Reverts Hydrogen pipes in hangars, adds hydrogen pumps to nt ice and indie outpost

Thera-Pissed:
## About The Pull Request
Reverts hydrogen pipes and exchange terminal in hangars
Adds hydrogen pump and exchange to main outpost level
image
indiechanges

known bug, if a second outpost is spawned with admin tools, the volume pump breaks and needs to be re-wrenched. no idea how to fix it. This used to occur on all hangars so this should be an improvement?

## Why It's Good For The Game
Atmospherics bugs make pumps not work in the most recent hangar, among other jank. Also, the fuel pipelines weren't great either.
Hydrogen needs more work still, but at least it'll be obtainable without having to work around obscure bugs.

## Changelog

:cl:
add: added hydrogen pump and exchange to ntice and indie outposts
del: reverts hangars, removing hydrogen pump, pipes, and exchange from indie and ntice hangars
/:cl:

### DIFF
--- a/_maps/outpost/hangar/indie_space_20x20.dmm
+++ b/_maps/outpost/hangar/indie_space_20x20.dmm
@@ -66,11 +66,10 @@
 /turf/closed/indestructible/reinforced,
 /area/hangar)
 "az" = (
-/obj/structure/catwalk/over/plated_catwalk,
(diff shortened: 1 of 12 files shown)


thgvr:
Looks good, needed for a while.

PR #3871, opened 2024-12-03
SKM-24u LMG Modification on the Black Market

firebudgy:
## About The Pull Request

Adds the SKM-24u LMG Modification to the black market, for on average about 1000 above standard SKM sale price. Given the statistical similarities to the regular SKM minus it's having of a bipod and the non-guarantee of drum magazines, this felt right.

Also fixes bipods to actually work at all after the code for them was just outright removed in a PR I can't care to find.

## Why It's Good For The Game

New rare weapons variety to the black market, and a defensive option in the form of a bipod-attached LMG should make for some interesting situations.

## Changelog

:cl:
add: An intercepted shipment of CM-40 barrels has led to an influx of SKM-24u LMGs to the black market. Report sightings to your nearest CLIP representative.
fix: Bipods actually work now.
/:cl:


rye-rice (approved):
holy shit THANK YOU for making the lmg bipods work again

and yes i know bipods should be an attachment but also i do like bipods working so this is fine

PR #3881, opened 2024-12-04
fixes max weight on surgery cases

Jedi-Toothpaste:
## About The Pull Request
Makes it so the Circ-Saw and the 7th item can fit in by defining a max combined weight to the item
(I am aware it's infinity but there is a cap to the amount of items that can be held. If someone wants to use cases to smuggle 7 circ saws at a time then be my guest.)

## Why It's Good For The Game
So surgery cases work as intended.

## Changelog

:cl:
fix: surgery cases now hold all the items
/:cl:

### DIFF
--- a/code/game/objects/items/storage/cases.dm
+++ b/code/game/objects/items/storage/cases.dm
@@ -13,6 +13,7 @@
 	throw_speed = 3
 	throw_range = 7
 	var/empty = FALSE
+	var/max_items = 10
 	w_class = WEIGHT_CLASS_BULKY
 
 /obj/item/storage/case/ComponentInitialize()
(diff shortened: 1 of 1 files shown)


code/game/objects/items/storage/cases.dm
```diff
@@ -13,6 +13,7 @@
 	throw_speed = 3
 	throw_range = 7
 	var/empty = FALSE
+	var/max_items = 10
```
MarkSuckerberg: did you mean to use this somewhere?

Jedi-Toothpaste: tl;dr it has no use right now
However, I may make it into something for the future but right now it's just a generic small-item storage which is bulky. 
In the future I could make it printable in the lathe for storing small things like circuitboards? Hell, I could add that onto this PR if you like since it'd mesh well with missions to gather small items like puce and chilly peppers
(though I would go and make sure it can ONLY store small items)

MarkSuckerberg: OH I meant that var in particular, the max_items thing

Jedi-Toothpaste: it has no use rn. I just set it to ten in a vain attempt to try and get it's child type (the surgery case) to hold enough items then didn't set it back.
We can blow up this line if you'd like

MarkSuckerberg: yee blow it up and I'll merge

Jedi-Toothpaste: all done

PR #3888, opened 2024-12-06
Hivebot Spawner Faction

Erikafox:
...is now the same as hivebots, so they don't infight
:cl:
fix: hivebots should no longer attack hivebot spawners
/:cl:


MarkSuckerberg (approved):
lol

PR #3512, opened 2024-10-09
Attempts to Fix Temperature + Makes It More Realistic in the Face of Extreme Temperatures

rye-rice:
## About The Pull Request

Zym's temperture PR apparently doesn't work as intended, this is not because of a coding error but simply because SS13 humans are extremely fucking durable somehow

This attemps to fix it by making humands much more squishy and easier for players to notice cold/warm areas without it being in the danger zone.

This also adds a fun mechanic where staying hydrated reduces burn damage taken from hot air.

And in order to fine tune some planet's tempertures,  rockplanet and beachplanet's gases have been randomized.

ports https://github.com/tgstation/tgstation/pull/82336 as well, a nice cold breath effect!

## Why It's Good For The Game

you should no longer be fine in -50 C nor 100 C, this may be frustrating, but holy shit you should NOT be fine at 100C. Hardsuit negates this completely

## Changelog

:cl:
add: Melbert, rye-rice
balance: you will no longer be fine at 100 c, nor 50 c, you should also NOT be fine at -100 c either. clothing helps.
fix: should be able to tell slighlty chilly tempertures
qol: Breathing cold air now has a particle effect associated, careful not to let your glasses fog up.
qol: Breathing cold air will now occasionally make your spaceman shiver. Brrr.
/:cl:

PR #3895, opened 2024-12-08
Partial Tempchanges Revert

Erikafox:
## About The Pull Request
Partial revert of the temp changes introduced in #3512. 

Temperature is an unfortunately horrific system that needs more work than value fudging to produce a good effect, and I am of the opinion that dealing with all the edge cases introduced by #3512 outweights the benefits it's produced

## Changelog

:cl:
balance: Temperature Mechanics have been partially reverted. Please report any weirdness
balance: Temperature now defaults to respecting insulation instead of ignoring it.
/:cl:


Erikafox:
needs a testmerge to make sure that I have not caused any issues.

FalloutFalcon (commented):
I really cant parse much from the numbers without seeing it ingame myself, I havent played a ton recently to feel the temp effects. But i agree with the change for use_insulation default being true. Test merge required

PR #3910, opened 2024-12-16
Arke-Class Surveying Ship | CLIP-LAND

Ical92:
## About The Pull Request
Adds a new specialist ship to CLIP's roster, a small CLIP-LAND ship for surveying planets and relaying information to friendly ships. Has a telecommunications setup and basic equipment, as well as a small armory for the Captain to distribute. 

image

<details><summary>SDMM Screenshots</summary>
<p>

image

image

</p>
</details> 

<details><summary>More In-Game Screenshots</summary>
<p>

image
image
image
image

</p>
</details> 
Also adds a tcomms relay to the Atlas, located in the port side of engineering, as well as giving CLIP headsets to start with.

## Why It's Good For The Game
Adds a new small-crew ship and a new ship for CLIP that isn't part of CMM. Gives telecommunications access to CLIP ships without the need for a larger command ship.

## Changelog

:cl:
add: Arke-Class Surveying Ship
add: The Atlas-Class now has a Telecomms Relay
/:cl:

### DIFF
--- a/code/modules/clothing/outfits/factions/minutemen.dm
+++ b/code/modules/clothing/outfits/factions/minutemen.dm
@@ -11,7 +11,8 @@
 	box = /obj/item/storage/box/survival/clip
 	ears = /obj/item/radio/headset
 
-	satchel = /obj/item/storage/backpack/security/clip
+	backpack = /obj/item/storage/backpack/security/clip
+	satchel = /obj/item/storage/backpack/satchel/sec/clip
 
 // 	var/list/selectable_alt_titles = list()
 
--- a/code/modules/clothing/outfits/factions/minutemen.dm
+++ b/code/modules/clothing/outfits/factions/minutemen.dm
@@ -11,6 +11,8 @@
 	box = /obj/item/storage/box/survival/clip
 	ears = /obj/item/radio/headset
 
+	satchel = /obj/item/storage/backpack/security/clip
+
 // 	var/list/selectable_alt_titles = list()
 
 /datum/outfit/job/clip/post_equip(mob/living/carbon/human/H, visualsOnly)
@@ -43,6 +45,7 @@
 	job_icon = "scribe"
 
 	suit = /obj/item/clothing/suit/hazardvest
+	head = /obj/item/clothing/head/soft/utility_navy
 	gloves = /obj/item/clothing/gloves/color/black
 	accessory = /obj/item/clothing/accessory/pocketprotector/full
 	shoes = /obj/item/clothing/shoes/workboots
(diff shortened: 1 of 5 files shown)


meemofcourse (changes requested):
imagen

change the wooden table, i dont think it fits

imagen

i would prefer if the toolboxes were on a table, or a shelf

meemofcourse (approved):
yea

Jedi-Toothpaste:
Hi, bit of feedback on the ship
firstly, some qol I wouldn't mind seeing.
Could the button for the shutters be moved closer to the pilot's seat?
image
Here is a potential place for it

To my knowledge the windoors to the guns are accessible by the Deckhands, would it be possible to keep this to elevated access as it's a Government vessel, not everyone would be weapon's trained!

I don't think the ship needs a doctor slot, given it doesn't have a dedicated doctor's space, I think the slot might be better either gone, or another engineer of some kind? 
Or even contracted IRMG security like you see on old ships like the Beluga (i'd go over this with a maptainer however)

The comms room starts vented, instead of cold, if you want to keep it vented, can you map in a vent set to siphon which injects back outside, like in this diagram for example?
image

Would you be able to move the light-switch in the engine room so it can be flicked from the stairs?
image

Also, a lightswitch in the cryo-room would be nice, but not required.
image



Secondly, to bugs, there aren't that many, to my knowledge.
First off, this lightswitch in the comms room seems misaligned
image

Only the captain spawns with a radio headset, and since there is no way of obtaining more radio keys they just have to use a custom frequency instead of the Minutemen one you created.

Ical92:
@Jedi-Toothpaste 
> Could the button for the shutters be moved closer to the pilot's seat?

nodnods, definitely, 

> To my knowledge the windoors to the guns are accessible by the Deckhands

It *shouldn't* be, so I'll double check the access on those, thanks for the catch (Update, definitely *not* supposed to happen, but not a map issue as far as I know)

> I don't think the ship needs a doctor slot, given it doesn't have a dedicated doctor's space

Definitely not a full "doctor" but I'd say some sort of designated medical person is helpful, may rephrase the job title somehow though to clarify that they won't be getting a proper medbay. 

> The comms room starts vented, instead of cold

Not *quite* vented, just low pressure

Onto the smaller things
* I'll probably leave the light switch on that same tile, since that's quite a distance to shift it
* Can definitely add a switch to the cryo room
* Tcomms switch is definitely in an awkward spot, I'll try and center it a bit more to see if it helps
* Talking with maptainers to check about headsets, but should be able to get starting headsets in, and already added a couple spare headsets in the crew locker.

code/modules/clothing/outfits/factions/minutemen.dm
```diff
@@ -155,6 +169,28 @@
 	duffelbag = /obj/item/storage/backpack/duffelbag/med
 	courierbag = /obj/item/storage/backpack/messenger/med
 
+/datum/outfit/job/clip/medtech
```
Sadhorizon, line 173: This is defined twice, once as `/medtech`, once as `/minutemen/paramedic`. Same outfit twice seems weird, I'd remove the old unused one (the /paramedic one)
Also, afaik, this outfit is cruft from the Asclepius. You should check up with someone on how to update it I think, because it uses essentially indie gear.

Ical92: `/medtech` was meant to be the non-military version of `/minutemen/paramedic` (hence why it doesn't have an armored vest. Though I agree this definitely could use some more explicit CLIP identification than just *being very blue*.

code/modules/clothing/outfits/factions/minutemen.dm
```diff
@@ -9,6 +9,9 @@
 	faction_icon = "bg_clip"
 
 	box = /obj/item/storage/box/survival/clip
+	ears = /obj/item/radio/headset
+
+	satchel = /obj/item/storage/backpack/security/clip
```
Sadhorizon: This seems odd? There exists an actual CLIP satchel, if you want to use it.

Burning02:
Under this specific airlock in the... airlock, theres a missing tile, leading to space
image

meemofcourse (changes requested):
conducted a very short test

imagen

i'm not sure if like this double decaling

imagen

that's not good

code/modules/clothing/outfits/factions/minutemen.dm
Ical92: Mistyped, put the backpack and satchel in properly

meemofcourse (approved):
ships tested

PR #3915, opened 2024-12-17
The Great Shiptest IRS Audit

Geoengi:
## About The Pull Request

- Removes the ability to place material coins into slot machines by changing the pay mode; they only work with cash bundles now. Thank @flopppi for this wonderful edit.
- The roulette table no longer gets stuck with spinning numbers and now references bank cards over ID cards. 
- Vendors no longer ask for ID cards when no account is presented and instead bank cards.
## Why It's Good For The Game
So turns out that if you place coins in a slot machine and go to withdraw them or deconstruct the machine, it literally tries to throw infinite coins at your head and kills the server. Obviously, it should not do this.
## Changelog
:cl:
fix: Partially fixed persistent overlay on roulette tables.
tweak: Replaced references to ID cards with bank cards where applicable.
code: Slot machines no longer accept material coins due to them nuking the server if withdrawn.
/:cl:


Ratvarr:
I think you messed up your commits or. something. your merging your master not branch

Geoengi:
woops - will use branch

PR #3919, opened 2024-12-19
Minor mapcut fix

Burning02:
## About The Pull Request

As is in the title, fixes a small error in the maptainer mapcut PR

## Why It's Good For The Game

Instead of corporate mining being removed in the ruin datum, bigderelict1 was instead removed in said datum, causing corporate mining to be spawned instead with nothing. Fixes that so derelict1 exists again

## Changelog

:cl:
fix: Bigderelict1 now properly spawns instead of Corporate Mining
/:cl:

### DIFF
--- a/code/datums/ruins/space.dm
+++ b/code/datums/ruins/space.dm
@@ -6,12 +6,13 @@
 	allow_duplicates = FALSE
 	ruin_type = RUINTYPE_SPACE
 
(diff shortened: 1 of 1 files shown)


meemofcourse:
the haters were right. good on the haters

Burning02:
> the haters were right. good on the haters

What haters??

PR #3934, opened 2024-12-25
Fixes skirt preferences for Inteq uniforms

Ical92:
## About The Pull Request
Because of how Inteq clothing was pathed, uniforms were failing to get the subtype for skirts meaning they could not be properly selected. Ex. expecting `.../corpsman/skirt` when it was actually `.../skirt/corpsman`. This fixes the paths for skirts and updates them on used maps.

## Why It's Good For The Game
Lets players actually use Inteq skirtlenecks, rather than failing to retrieve the subtype and falling back to a regular turtleneck.

## Changelog

:cl:
fix: skirt preferences now work for inteq uniforms
/:cl:


meemofcourse:
i would prefer something similar to alt uniforms where the skirt preference can be defined in the outfit datum but this is good too

Ical92:
Definitely thought that would be a cleaner solution while fixing this, may work on that next but this works in the meantime

Ical92:
Bleh, something went wrong with the update script, will fix that (Fixed now)

code/modules/clothing/under/syndicate.dm
```diff
@@ -240,7 +240,7 @@
 	icon_state = "inteqeng"
 	supports_variations = KEPORI_VARIATION | VOX_VARIATION | DIGITIGRADE_VARIATION
 
-/obj/item/clothing/under/syndicate/inteq/skirt/artificer
+/obj/item/clothing/under/syndicate/inteq/artificer/skirt
```
Sadhorizon, line 243: Base inteq skirt has `body_parts_covered = CHEST|GROIN|ARMS`, so it doesn't cover legs. Should probably copy it over, just so it's still the same.

Ical92:
Flaky tests my beloathed

PR #3945, opened 2024-12-29
Makes chem/soda/booze dispensers take ore/refined bluespace crystals instead of sheet polycrystals.

Thera-Pissed:
## About The Pull Request
Currently it takes polycrystals to make chem dispensers, which is jank because you cant turn the ore/refined into the sheet type, while you can turn the sheet into the ore/refined. its weird and dumb.

tested on local it works wowie

## Why It's Good For The Game
cant get polycrystals from refinery afaik

## Changelog

:cl:
fix: chem dispensers take refined bluespace instead of polycrystal
/:cl:


Burning02:
You _can_ turn them back into their ore form by clicking it with your other empty hand, but nowhere does it tell you this to my knowledge.

PR #3946, opened 2024-12-29
Renames and rewrites the Odysseus and Gygax

Apogee-dev:
## About The Pull Request

Changes the Odysseus and Gygax to be manufactured by Cybersun. Finally, the company that specializes in robotics and cybernetics actually has a presence in the exosuit market.
- Gygax is now the 500 Series, represented by the 501p (Gygax) and 515 EX (Dark Gygax).
- Odysseus is now the 200 Series, represented by the 202r, and potentially more in the future.
- 515 EX has better power efficiency when running in overcharge mode. The armor is still the same as the 501p, but you can run around like 30% more without zapping your batteries to dust.

Originally, this was going to be packaged with a resprite/redesign of the exosuits in question and some tweaks to their performance, but that will have to wait until later because spriting exosuits is hell on earth.

....I'll change the wide variety of paths associated with them Later(tm)

## Why It's Good For The Game

cybersun actually gets a presence in the exosuit market and we can finally clean up some of the old outdated scraps of lore that were still hanging around in relation to these two exosuits

## Changelog

:cl:
code: Renamed the Gygax and Odysseus and updated their lore
add: Dark Gygax (now 515 EX) has better efficiency when overloading its legs.
/:cl:


thgvr:
ehh.. having names for them is better than a number identifier, this is less cool

Apogee-dev:
> ehh.. having names for them is better than a number identifier, this is less cool

i'm specifically going for BMW or Tesla style model names here because it fits with Cybersun's corporate personality and is extremely distinctive compared to most of the existing content (particularly exosuits), which do indeed tend to have "cool" names (usually askance references to media or myth/fantasy names). I'm trying to make them stand out and make it hard to confuse them for something more generic (read: built by nanotrasen).

in-universe, cybersun isn't trying to sound cool, it's trying to sound elite. the product doesn't need a pithy nickname, it speaks for itself- or so the marketers want you to think anyway.

FalloutFalcon (approved):

Valorium-RB:
Lore on the descriptions seems good, but I agree with Thgvr that switching from proper names to numbers is less interesting. It'll be less memorable for players to identify and it doesn't sound much more 'elite' to have the same name scheme as a Texas Instruments graphing calculator. Additionally, existing lore laid out in our exosuit documentation provides a dramatically different name schema for Cybersun exosuits than presented here.

Apogee-dev:
> Lore on the descriptions seems good, but I agree with Thgvr that switching from proper names to numbers is less interesting. It'll be less memorable for players to identify and it doesn't sound much more 'elite' to have the same name scheme as a Texas Instruments graphing calculator. Additionally, existing lore laid out in our exosuit documentation provides a dramatically different name schema for Cybersun exosuits than presented here.

The only Cybersun mechs mentioned in the lore by name are the Obliterator and Devastator... both of which are giant mecha, not exosuits, and would have been made purely for combat use anyway. I really don't see cybersun naming a medical exosuit and a light scout/security exosuit, both destined for the civilian market, anything remotely like either of those. 
I still wholeheartedly disagree on the "coolness" argument too- Everything having a cool and/or edgy name is just plain boring because there's no variety, in my opinion. Players can and will think of nicknames on their own. If I'm changing the names on these again I'm still going for a luxury car angle, even if it isnt the BMW style. It suits Cybersun's personality, product types, and the way they're written a lot better than anything else.

PR #2610, opened 2023-12-24
re:lathe machines

rye-rice:
## About The Pull Request

resprites most 'production' machines. most of these are related to research but research is fucking lame so i didn't name the pr that 

also redoes the handheld radio sprite because might as fucking well
made on behalf of @Imaginos16 

## Why It's Good For The Game

image

## Changelog

:cl: Imaginos16
add: Resprites most production machines!
add: Resprites handheld radios
/:cl:

PR #3952, opened 2025-01-01
Camera Console Recalibration and Bodycameras

zimon9:
## About The Pull Request

This change modifies camera consoles so that they can take a custom network input. Camera consoles were also modified to get closer in parity to the SecurEye app. Tablets were given the ability to download securEye.

| Camera Consoles |image|
|-----------------|---|
| SecurEye        |image|

Static cameras were modified so that using a multitool on one with its panel open now presents several new options. One can copy a network that a camera is set to onto a multitool, transfer a saved network into a camera, or even set a new network entirely.
image
They were also given a new variable, can_transmit_across_z_levels, which defaults to false. This variable is for varedits, and a potential future addition that may enable transmission across Z-levels for some cost.

A new portable type of camera was also added, the body camera. It can be activated or deactivated by being alt-clicked, and its tag or network settings can be modified if a multitool is used on it.
They can be purchased from the outpost cargo market at a rate of two units for 250 credits.

Bodycameras can be worn in your pockets, your coat/armor, security belts or webbings, or a helmet, but they get obscured when put into a bag, a box, or your boots.

They have a view range of 5, which is 2 less than camera structures.

https://github.com/user-attachments/assets/9c69cd3a-59c4-40a7-94fd-2be9932ec15f

https://github.com/user-attachments/assets/75655b88-583f-44a6-93ef-e59e159826ce

This PR by Timberpoes proved instrumental in solving a major problem during the development process:
https://github.com/tgstation/tgstation/pull/52767

The bodycamera sprite is modeled after the handheld radio sprite made by @rye-rice in this PR:
https://github.com/shiptest-ss13/Shiptest/pull/2610
image
I'm not a good pixel artist, so any feedback here is appreciated, on top of feedback with regards to anything else about this PR.

## Why It's Good For The Game

It would be nice to be able to track your crew more effectively, be it if you were to send them out to scrap a derelict, or to secure contested territory. This is something that Command or Foremen that want to oversee operations might find useful, as it can be kind of boring to not be able to see your crew when they're at work. This should help that issue.

## Changelog

:cl: Rye-Rice, Timberpoes
add: Added bodycameras
add: Added the ability to set custom networks on cameras and camera consoles
code: Made camera consoles and SecurEye have better parity
/:cl:


code/game/objects/items/bodycamera.dm
```diff
@@ -0,0 +1,128 @@
...
+/obj/item/bodycamera/Destroy()
+	if(can_use())
+		toggle_cam(null, 0) //kick anyone viewing out and remove from the camera chunks
+	GLOB.cameranet.cameras -= src
+	return ..()
+
+/obj/item/bodycamera/examine(mob/user)
+	. += ..()
+	. += "The body camera is currently [status ? "<span class='green'>ON</span>" : "<span class='red'>OFF</span>"]. Ctrl-Click to toggle its status."
+	. += "The body camera is set to a nametag of '[c_tag]'. Use a multitool on [src] in order to set a new nametag."
```
Erikafox: I would put the nametag/network stuff in the examine_more.

zimon9: I think that'd be better, yea
I've made that change now

zimon9: I'm a little conflicted on this, actually. From what I understand, no other item has an examine-more description, so this kind of breaks tradition, and I'm loathe to make things that aren't consistent with other things in the game. Inconsistency is one of the things that annoys me about older things in the game.

Erikafox: some objects use examine_mores. I think its fine to expand on it, especially since I shouldn't be able to tell the technical details on it from a cursory glance.

zimon9: The items that do that seem to make it imply that examining more is less of something required, and more of a bit of added flavor. That, or it's a way for someone to interact with an object without having to click on it. As for being able to see technical details, I liken it to how you can see a radio's frequency setting by examining it, in that the details of the network setting and the tag setting are on a small display on the body of the device. However, it may be good to make it less of something that can be seen quite as easily, yeah. From a standpoint of someone that might be newer to the game, though, it might be a little confusing to leave it like this, though.

In any case, this isn't really a big deal or anything, but it's something I've been thinking about. I do think there is a lot of room for improvement with the design of this.

MassiveMen:
Small issue. The bodycams dont have a distance check. You can turn them on as long as you see them.

Ical92:
> Small issue. The bodycams dont have a distance check. You can turn them on as long as you see them.

The grinch returns ..

zimon9:
> Small issue. The bodycams dont have a distance check. You can turn them on as long as you see them.

Bluetooth activation. Lovely.

code/game/machinery/computer/camera.dm
```diff
@@ -7,14 +7,18 @@
 	light_color = COLOR_SOFT_RED
 
 	var/list/network = list("ss13")
+	var/tempNetwork = list("")
```
FalloutFalcon: im a little sleepy but everything about this var is silly. For now- camel_case please.

FalloutFalcon (changes requested):

code/game/machinery/computer/camera.dm
zimon9: Yeaa, you're right, let me fix that, thank you

FalloutFalcon (approved):

PR #3955, opened 2025-01-03
Cargo Security Tab Adjustments

MemeSnorfer:
## About The Pull Request

Removes some crufty items from the Security tab in cargo, as well as putting some new things in. Also adjusts some prices.

image

## Why It's Good For The Game

Makes some of these crates more worth purchasing. This is really just a spite PR because I only want flashbangs from cargo and nothing else.

## Changelog

:cl:
del: Security Supplies crate
del: Chemical Implants crate
del: Tracking Implants crate
del: Wall Mounted Flash crate
del: Syringe Gun crate
del: DRAGnet crate
add: Flashbangs crate
add: Teargas Grenades crate
add: Camera Console crate
add: Pepper Spray crate
balance: Lowered Security Barrier Grenades crate cost to 500 (from 2000)
balance: Lowered Stingbang Grenades crate cost to 1500 (from 7500)
balance: Lowered Stun Baton crate cost to 2500 (from 3000)
/:cl:

### DIFF
--- a/code/modules/cargo/packs/sec_supply.dm
+++ b/code/modules/cargo/packs/sec_supply.dm
@@ -132,7 +132,8 @@
 
 /datum/supply_pack/sec_supply/baton
 	name = "Stun Baton Crate"
-	desc = "Arm your vessel security with a stun baton. Batteries included."	cost = 1500
+	desc = "Arm your vessel security with a stun baton. Batteries included."
+	cost = 1500
 	contains = list(/obj/item/melee/baton/loaded)
 	crate_name = "stun baton crate"
 
--- a/code/modules/cargo/packs/sec_supply.dm
+++ b/code/modules/cargo/packs/sec_supply.dm
@@ -12,28 +12,14 @@
 	contains = list(/obj/item/clothing/accessory/holster)
 	crate_name = "holster crate"
 
-/datum/supply_pack/sec_supply/chemimp
-	name = "Chemical Implants Crate"
-	desc = "Contains five remote chemical implants."
-	cost = 1000
-	contains = list(/obj/item/storage/box/chemimp)
-	crate_name = "chemical implant crate"
-
-/datum/supply_pack/sec_supply/trackingimp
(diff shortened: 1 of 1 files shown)


Erikafox:
IMO the syringe gun does have a reason to exist. It just should be in something like medical instead

MemeSnorfer:
> IMO the syringe gun does have a reason to exist. It just should be in something like medical instead

With a price tag of 8000, there is no way anybody is ever going to buy it outside of extreme/gimmicky scenarios. It only existed in cargo for the sake of keeping it around

code/modules/cargo/packs/sec_supply.dm
```diff
@@ -122,38 +109,34 @@
...
-	cost = 8000
-	contains = list(/obj/item/gun/syringe)
-	crate_name = "syringe gun crate"
-
 /datum/supply_pack/sec_supply/baton
-	name = "Stun Batons Crate"
+	name = "Stun Baton Crate"
 	desc = "Arm your vessel security with a stun baton. Batteries included."
-	cost = 3000
+	cost = 1500
```
FalloutFalcon: I honestly still think this is a fair price saying its one of the best melees (sadly) and this is about the price of a higher tier melee
```suggestion
	cost = 2500
```

FalloutFalcon (changes requested):

code/modules/cargo/packs/sec_supply.dm
MemeSnorfer: You're right actually. As much as I've found it ridiculous that stun batons cost the same as submachine guns, stun batons still suck.

firebudgy: honestly you could just remove them from cargo outright. nobody would miss them

Martinpachu: you could delete them from the game

FalloutFalcon (approved):

PR #3968, opened 2025-01-05
Libertatia 2

Apogee-dev:
## About The Pull Request
falmouth_ingame
> "A small, cheap cargo ship most remarkable for its very high speed and maneuverability for a ship of its type. Marketed as a do-it-all light ship suitable for short-range runs and courier work, the Falmouth is extremely popular among smugglers and the occasional ambitious pirate; Falmouths operating on legitimate business are downright rare in some areas, particularly on the Frontier."

Replaces the Libertatia with the Falmouth-class Light Freighter, a ship with a similar design, but a slightly different focus.

falmouth
falmouth_areas

The Falmouth shifts the Libertatia's "combat" focus (which had been devastated by power creep anyway) to lean more into the flavor of being a smuggling ship rather than an outright pirate ship. While not a combat ship, the Falmouth _does_ nevertheless come with an eclectic collection of weapons:

- One CM357 Heavy Pistol (stolen)
- One Model 11 Buckmaster Shotgun
- One Candor Pistol
- One E-10 Laser Pistol
- One E-11 Heap Of Garbage

Crew complement consists of a captain, a first mate, a security detail, and two deckhands for a total of 5 crew.

My hope is that the Falmouth can take the broad pattern and flavor of the old Libertatia and make it a much more liveable, balanced, and _playable_ ship.

## Why It's Good For The Game

The Libertatia sucks. This, hopefully, doesn't.

## Changelog

:cl:
add: added Miskilamo Falmouth-class Light Freighter
del: removed Libertatia-class Hauler
/:cl:


Apogee-dev:
chat i'm so tired

Zevotech (dismissed):
SMES needs the wire under it to be top right instead of right in order to power the ship. The grav gen should also start anchored.
image

The exterior thrusters aren't wired to their prechargers.
image

This injector needs to be on piping_layer 1 in order to connect to the grey pipe and output gas.
image

The safe starts open due to the number of tumblers being 0. Unsure if this is intentional.
image

3 of the SSU's start empty. Use the standard_unit type if you want them to spawn with normal EVA suits.
image

The Security Detail cannot access the ammunition locker. The Candor in the weapon closet has no magazines anywhere to use for it.
image


Other than that, Ship looks fine. Can't wait to test it out on live. The doggirls are gonna love this one.

Apogee-dev:
>wiring

Fixed.

>pipe layer

It's already set to layer 1.

>safe

gonna try changing to 3 tumblers, I was under the impression number of tumblers was only used to generate the randomized ones

>SSUs

fixed

>ammo locker

Technically intended. the gun locker is ALSO supposed to be locked, but that code seems to be broken. these are both set to use bridge access. that being said i did that when i was still planning to put some armor and other goodies in that locker, so it can change to secoff access

Erikafox:
>tumblers

add an if statement to the tumbler init - if(!tumblers.len()), do the generation of tumblers, and then it'll work with the number set.

meemofcourse (approved):
liberal tail a so good they made a second one

PR #3970, opened 2025-01-06
Crates load onto shelves when mapped

Ical92:
## About The Pull Request
Crates mapped on top of shelves now are loaded onto the shelf automatically, additionally, adds a new mapping helper to load crates onto nearby shelves.

image
(Example on the Talos-class, loading crates onto shelves and nearby crates loaded using the mapping helper)

## Why It's Good For The Game
Players wont need to load crates onto shelves at the start of each round, and mappers can be more creative with their use of shelves.

## Changelog

:cl:
add: Outpost staff have hired professional crate shelvers
/:cl:


code/modules/mapping/mapping_helpers.dm
```diff
@@ -490,6 +490,25 @@ INITIALIZE_IMMEDIATE(/obj/effect/mapping_helpers/no_lava)
 	icon_cache[url] = I
 	return I
 
+/obj/effect/mapping_helpers/crate_shelve
```
FalloutFalcon, line 493: this helper doesn't really seem needed when you can just varedit the shelf.

code/game/objects/structures/crateshelf.dm
```diff
@@ -15,6 +16,7 @@
 	var/capacity = DEFAULT_SHELF_CAPACITY
 	var/max_capacity = DEFAULT_SHELF_MAX_CAPACITY
 	var/use_delay = DEFAULT_SHELF_USE_DELAY
+	var/pickup = DEFAULT_SHELF_PICKUP
```
FalloutFalcon: This doesn't really warrant a define and could use a slightly more clear variable name
```suggestion
	var/pickup_crates = TRUE
```

code/game/objects/structures/crateshelf.dm
```diff
@@ -114,7 +124,7 @@
 	if(!next_free) // If we don't find an empty slot, return early.
 		balloon_alert(user, "shelf full!")
 		return FALSE
-	if(do_after(user, use_delay, target = crate))
+	if(!user || do_after(user, use_delay, target = crate))
```
FalloutFalcon: this might have a chance to runtime... i am a little eepy so i might be wrong.

code/game/objects/structures/crateshelf.dm
```diff
@@ -179,6 +189,11 @@
 		transfer_fingerprints_to(new_parts)
 	return ..()
 
+/obj/structure/crate_shelf/proc/load_crates(atom/movable/holder)
+	for(var/obj/structure/closet/crate/crate in loc)
+		if(crate != src && !load(crate))
```
FalloutFalcon: I dont think crate != src does anything in practice here.

FalloutFalcon (changes requested):

code/game/objects/structures/crateshelf.dm
Ical92: True! I think that's leftover from the previous version of it I made

code/game/objects/structures/crateshelf.dm
Ical92: Also true, will remove the define

code/modules/mapping/mapping_helpers.dm
Ical92: The helper is for use on crates that aren't mapped directly on top of the shelf, the shelf only picks up crates that are on the same tile as it on it's own.

Ical92:
Oh god that's so many , failed checks what happened .

FalloutFalcon (approved):

code/modules/mapping/mapping_helpers.dm
```diff
@@ -490,6 +490,25 @@ INITIALIZE_IMMEDIATE(/obj/effect/mapping_helpers/no_lava)
...
+	var/obj/structure/closet/crate/crate = locate(/obj/structure/closet/crate) in loc
+	if(!crate)
+		log_mapping("[src] failed to find a crate at [AREACOORD(src)]")
+	else
+		shelve(crate)
+
+/obj/effect/mapping_helpers/crate_shelve/proc/shelve(crate)
+	var/obj/structure/crate_shelf/shelf = locate(/obj/structure/crate_shelf) in range(range, crate)
+	if(!shelf.load(crate))
+		log_mapping("[src] failed to shelve a crate at [AREACOORD(src)]")
```
Erikafox, line 510: should indicate shelf capacity as a potential avenue for failure

Ical92: Not sure if this is exactly what you meant, but load() logs when a shelf tries to shelve something while full.
```dm	
if(!next_free) // If we don't find an empty slot, return early.
	if(ismob(user))
		balloon_alert(user, "shelf full!")
	else
		log_mapping("[src] attempted to shelve a crate when full at [AREACOORD(src)]")
	return FALSE

Erikafox: I would've put the failure on the helper myself, but that should work

PR #3972, opened 2025-01-07
Adds a Shaft Miner to the Harrier, removes the Janitor (Role)

firebudgy:
## About The Pull Request

Adds a Shaft Miner, and minorly adjusts the N+S cargo office to accommodate. In return to keep the crew size identical, the Janitor's slot has been removed and the room lowered to general access. Assistants can take up the role if they would like to be a janitor.
image 
image

## Why It's Good For The Game

Inherently, the Harrier has an issue with integration into the gameplay loop. A lack of combat equipment means that ruin raiding is not much of an option, and most missions are either ruin raiding... or mining. But the Harrier is unable to properly perform mining as almost all of it's staff are civilian in nature. What's left for it is it's medical bay, which is worse than dedicated medships and most would rather interact with an Atoll than a Harrier.

Now, you may say, what about the Cargo Techs?
Cargo Technicians are wage-slaves to N+S, but their role is *not* excavation. Nor is the Supply Director. These are warehouse workers. There is a specialized role for that. Having spoken with some of the most active Harrier captains, it's pretty clear that most of the time the Cargo Technician is sent out anyways for mining since someone is _needed_ to keep the ship running and supplied.

## Changelog

:cl:
add: N+S has deployed a new batch of Shaft Miners to Harrier-class vessels.
del: Janitors have been reassigned from the Harrier for retraining.
/:cl:


Apogee-dev:
Replace one cargo tech with a miner, don’t cull the janitor. Harrier is the only ship in the game with a disposals setup and a decent stock of janitor supplies, I’d rather that weren’t rendered pointless. He’s basically just a reflavored assistant and that’s fine.

firebudgy:
The Janitor's closet isn't pointless, but compared to Assistants the Janitor's job is only really relevant in the first 20 minutes of the round when the ship is dirty. Plus, as you said, it's basically a reflavored assistant. They've got access to the closet and the gear is all still there.

Apogee-dev:
I’d still rather you keep the janitor. It’s a reflavored assistant, and that’s better than a flavorless assistant. Three cargo jobs is also just too much, especially since the supply director tends to handle the cargo personally anyway, leaving the CT with nothing to do now that theres a miner. You also overgeared the miner- he doesn’t need an engineering belt and a toolbox, or a drill for that matter (most ships only provide a pickaxe even when they have miner slots).

Erikafox:
admittedly a drill is a good way to show that NT has actually invested in this sort of thing, and the only mechanical differences are tool speed of 0.4 vs 0.5 and "cannot be worn on back"

Apogee-dev:
Frankly I just dont think the Harrier should have much more gear than it already starts with, because it’s already a jack of all trades that can perform competently in several categories. It doesn’t need the kind of personal gear usually found on actual mining ships to top it off.

firebudgy:
I will pull the toolbox and belt out, that one was an oversight

But I do disagree significantly, it doesn't perform competently at all in several categories because it's just not equipped to actually engage in most of them.

Gristlebee:
I can see either the janitor or cargo tech going in favour of the miner, but I do think the janitor would be more suited to go. It's a much more niche role responsibility wise, and a bespoke janitor is pigeonholed on what they can do vs say the more generalized assistant. Janitors in general I feel are more suited to a station based gameplay loop more than anything.

As for the miner's gear, non-dedicated mining ships have gotten drills in the past such as the ~~Paracelsus~~ and the Inkwell. N+S is a dedicated mining company, and the basic drill is only marginally better than the pickaxe, so I don't feel it's too egregious. (I was thinking of the Chronicle, not the Paracelsus, that one was removed)

Aquidu:
As the most regular Harrier captain, I wholeheartedly support this change. Whenever I have a janitor on the ship, it's extremely difficult to get them any job after the ship is cleaned the first time. I think all but one of the janitor's I've ever had have all cryoed before the end of the round. It really is a job anyone can do, and has very little depth to it. Janitor just simply works better on station servers, where there is more space for things to get dirty, and more chances for dirt to appear.

Another common issue I run into is using roles in a reasonable IC way. The only two roles suited best for away teams are the paramedic and security officer, with the rest being a bit forced. Shiptest's gameplay loop requires you to interact with planets, contracts, and/or ruins to progress and proceed, barring a few exceptions like medships or barRP ships, which in of themselves have their own issues. On the topic of IC roles, the lack of a miner makes it very difficult to obtain ore(vital for gameplay) in a reasonable fashion. I usually end up sending the cargo tech to mine(if I have one) or I simply do it myself, neither of which feel thematically appropriate.

Janitor has never been relevant for more than the first five minutes of the ship's life, and the lack of a miner is a constant thorn in the side of trying to interact with the gameplay loop.

Apogee-dev:
The cargo tech is still a bit redundant with a miner around, and the miner’s gear does make the lower area of cargo one long, awful 1x1 passageway. Replace the cargo tech with the miner and _also_ cull the janitor, if you must. It’s not like Harriers ever get filled to capacity anyway.

firebudgy:
I did not add or remove any moveable tiles in the cargo space. It was a one-tile hallway when I got there
image

Apogee-dev:
You moved the disposal chute one tile to the left (assuming i’m not misremembering the lriginal layout), which made everything up to the cargo tech locker 1 wide.

firebudgy:
> You moved the disposal chute one tile to the left, which made everything up to the cargo tech locker 1 wide.

image
Nope

Apogee-dev:
Okay, wait, altering the silhouette of the ship to add a single locker is definitely worse. Now that I’m looking more closely you didnt even use the corner wall type. Just replace the cargo tech and his locker.

firebudgy:
> Okay, wait, altering the silhouette of the ship to add a single locker is definitely worse.

I moved a single piece of hull one tile. Corner type is my mistake, I can fix it.

Apogee-dev:
I mean, it also threw off the symmetry and the slope on that side of the ship and everything. I _strongly_ advise against altering ship exteriors like that. And all the points I previously made about the cargo tech still stand, he won’t have anything to do with the miner doing all the mining and the supply director hanging out in cargo all the time like he should be.

Aquidu:
I disagree with removing the cargo tech. Cargo tech was never designed to do mining, so to remove the cargo tech because a miner was added seems like a non sequitur. I've never gotten a supply director on the Harrier, so I can't speak on their redundancies, but every proposed interaction they would've had within my ship's forum post is positive. Additionally, cargo techs are far more common to have than supply directors, as I believe the job is easier to build a character for. I can easily think of several jobs I could give a cargo tech, even with a supply director present too. I cannot say the same for the Janitor. If having three members of the supply department causes issues, I would vastly prefer whatever issue is caused by that than I would with the current issues with the Harrier.

Apogee-dev:
Just find a way to add the shaft miner’s locker without editing the hull of the ship. The other changes are fine.

Martinpachu:
i think there being a spare janitor's uniform so that any old assistant can pick up and leave the job as they wish would be pretty great. have the janitor vibes without being locked into that forever. good change

Aquidu:
Screenshot (3349)
Why not add the locker where that rack with the pickaxe is now? After all, the locker already has a drill inside of it.
If not, the fuel tank is nice but isn't vital, since there's another tank in Engineering.
On an unrelated note, maybe swap the engineering scanner goggles for simple mesons so they aren't as powerful?

PrismMind:
I main on Aquidu's Harrier. I 100% support janitors conversion to shaft miner but I feel like cargo tech is separate enough that I would want it kept.

firebudgy:
> Screenshot (3349) Why not add the locker where that rack with the pickaxe is now? After all, the locker already has a drill inside of it. If not, the fuel tank is nice but isn't vital, since there's another tank in Engineering. On an unrelated note, maybe swap the engineering scanner goggles for simple mesons so they aren't as powerful?

Moved the locker there. Thanks for the suggestion

Apogee-dev:
Good spot for the locker, yeah. Sounds good to me. I should be able to review the map file properly later this week and approve it if there’s nothing else it needs.

Apogee-dev (approved):

Imaginos16:
LGTM

PR #3976, opened 2025-01-08
New Ship: Nimbus-Class Caravel

Erikafox:
## About The Pull Request

Adds a new Cybersun ship

image
image
image

The Nimbus-class is a small generalist ship designed for Cybersun. It's supposed to be a flighty little run around and shoot trouble ship.

- bad med
- decent living space
- central design
- good equipment
- okay engineering

## Why It's Good For The Game

Small ship that people can play on with a hook into the gameplay loop? Exquisite.

## Changelog

:cl:
add: Cybersun's newest doohickey, the Nimbus-Class Caravel, is now on sale in the Frontier.
/:cl:

### DIFF
--- a/_maps/configs/cybersun_nimbus.json
+++ b/_maps/configs/cybersun_nimbus.json
@@ -1,6 +1,6 @@
 {
 	"$schema": "https://raw.githubusercontent.com/shiptest-ss13/Shiptest/master/_maps/ship_config_schema.json",
-	"map_name": "Nimbus-class Caraval",
+	"map_name": "Nimbus-class Caravel",
 	"faction": "/datum/faction/syndicate/cybersun",
 	"prefix": "CSSV",
 	"manufacturer": "ISF Spacecraft",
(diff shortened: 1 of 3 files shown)


Gristlebee (dismissed):
- The laundry room has no lights. 
image
- The SMES has no terminal. 
image
- Move the cell charger and table it's on out somewhere else, or get rid of it since you already have on in the main engine room. You have to climb over it in order to reach the thruster.
image
- I think the number of tumblers need to be 3. The safe immediately cracks on every turn of the dial.
image
image
- Hatched wooden floors need to pried up with a crowbar, not a screwdriver due to how their pathed. You can either put a crowbar in on the bridge for now, or repath the tiles if you want, but you don't have to since that's a bit out of scope.
image
- Having connectors for the airfeed on top of the thrusters terminal doesn't look great to me. Similar issue to having a table in front of the thrusters in the other engine bay as it makes the thrusters hard to reach if you need a tank mounted for air.
image
- The ship doesn't appear to have a grav gen from what I can tell.
- The eswords in the armory are 30 force weapons, and pretty strong. Replace them with survival knifes, or use the e-bayonets
- I'd suggest using the syndicate space suits rather than indie sec hardsuits.
image
- Is this cabinet intentionally meant to be empty?
image

meemofcourse:
epic fail
fix your checks girl

PR #3986, opened 2025-01-11
.45-70 now comes in 24-round boxes

firebudgy:
## About The Pull Request

.45-70 now comes in 24 round boxes. Price reduced to 300 credits.
.45-70 match reduced to 400 credits. 800 credits is a scam.

## Why It's Good For The Game

For only being used in two guns, these rounds are prohibitively expensive and it's simply just economically better to run literally any other gun. This should make it a little bit more reasonable to purchase, and at the very least you can properly load an Ashhand 4 times with a .45-70 box. It did not divide nicely beforehand.

## Changelog

:cl:
balance: .45-70 comes in 24 round boxes.
balance: .45-70 is cheaper.
/:cl:


code/modules/cargo/packs/ammo.dm
```diff
@@ -246,15 +246,15 @@
...
+	desc = "Contains a 24-round box containing devastatingly powerful .45-70 caliber ammunition."
 	contains = list(/obj/item/storage/box/ammo/a4570)
 	cost = 400
 
 /datum/supply_pack/ammo/a4570_box/match
 	name = ".45-70 Match Crate"
-	desc = "Contains a 20-round box containing devastatingly powerful .45-70 caliber ammunition, that travels faster, pierces armour better, and ricochets off targets."
+	desc = "Contains a 24-round box containing devastatingly powerful .45-70 caliber ammunition, that travels faster, pierces armour better, and ricochets off targets."
 	contains = list(/obj/item/storage/box/ammo/a4570_match)
-	cost = 800
```
Erikafox, line 257: special ammo types are double the price of normal ammo

firebudgy: oh is that a standard? my bad. i can drop it

Apogee-dev:
I'm cautious about the price reduction here- 24 rounds at 300 credits makes for 12.5 credits per shot- cheaper than .308 (16.6) and the sniper cartridges (20), and a little less than twice as expensive as 8mm Illestren (why is 8mm a third as expensive as .308 anyway? it's a more powerful cartridge). It seems like an okay place for price efficiency, but we'll have to see in practice.


...also either 8mm needs a price nerf or .308 needs a price buff in another PR, or both and we split the difference idk

PR #3989, opened 2025-01-12
Factional Cargo: Magazines, I Love War Edition

firebudgy:
## About The Pull Request

Adds the Firestorm Pan Magazine (apogee request), Bulldog Drum Magazine (12 rounds over 8), Boomslang 10rnd Magazine and SKM Extended Magazine (effectively two magazines in one) to factional cargo.

## Why It's Good For The Game

Adds another way to upgrade without totally breaking the balance, as most of these are only small increases in ammo or the equivalent of two magazines in one, and have been priced appropriately. This also accounts for the cost it would take to refill these magazines, which effectively doubles the cost in some cases.

Most of these being as expensive as they are should also make sure they're thought about before purchasing.

## Changelog

:cl:
add: SKM Extended Magazine, Bulldog Drum Magazine, Boomslang 10rnd Magazine and Firestorm Pan Magazine to factional cargo.
/:cl:


Apogee-dev:
I'd actually price these at slightly higher than just double the cost of the smaller size (or whatever's appropriate depending on how much larger the expanded magazine is). Otherwise there's little reason to ever buy the smaller ones- your money is literally just as well spent on the much more convenient large magazine. Maybe make it more like two-and-a-half. That's the price of convenience (of not being shot when you have to reload)

firebudgy:
> I'd actually price these at slightly higher than just double the cost of the smaller size (or whatever's appropriate depending on how much larger the expanded magazine is). Otherwise there's little reason to ever buy the smaller ones- your money is literally just as well spent on the much more convenient large magazine. Maybe make it more like two-and-a-half. That's the price of convenience (of not being shot when you have to reload)

I agree honestly. I'm gonna do that.

code/modules/cargo/packs/magazines.dm
```diff
@@ -158,6 +165,14 @@
 	cost = 750
 	faction = /datum/faction/syndicate/scarborough_arms
 
+/datum/supply_pack/magazine/boomslang_mag_extended
+	name = "Boomslang-10 Magazine Crate"
```
Apogee-dev: I actually don't know why this and the gun crate refer to the military Boomslang as the Boomslang-10. The actual name here should be "MSR-90 'Boomslang'" - the Asp and Viper crates got the scheme right. Scarborough uses a different pattern for their civilian and military guns.

code/modules/cargo/packs/magazines.dm
```diff
@@ -203,6 +218,15 @@
 	contains = list(/obj/item/ammo_box/magazine/skm_762_40/empty)
 	cost = 500
 
+/datum/supply_pack/magazine/skm_ammo_extended
+	name = "SKM Extended Magazine Crate"
+	desc = "Contains a 7.62x40mm magazine for the SKM rifles, with a capacity of fourty rounds."
+	contains = list(/obj/item/ammo_box/magazine/skm_762_40/extended/empty)
+	cost = 1000
```
Apogee-dev: as I mentioned in a comment I'd make this more like 2.5 or 2.25x the price of the normal magazine so it's slightly less cost-efficient (but a lot more convenient). Same for the other extended mags.

Apogee-dev (changes requested):

code/modules/cargo/packs/magazines.dm
firebudgy: oh yeah no don't worry i'm already making the changes

code/modules/cargo/packs/magazines.dm
firebudgy: i... don't know where that came from actually. I can fix that real quick in this PR too
good catch!

code/modules/cargo/packs/magazines.dm
firebudgy: done

firebudgy: worth factoring in that all these magazines also come empty and must be filled with ammo.

Apogee-dev (approved):

PR #3994, opened 2025-01-13
Partially ports boxed messages, fieldset message blocks, and improves vote visibility

zimon9:
## About The Pull Request

This PR partially ports https://github.com/tgstation/tgstation/pull/88678 and implements the boxed_message and fieldset_legend blocks that Aylong designed for votes.

| Light Mode |image|
|------------|---|
| Dark Mode  |image|

Examine blocks were swapped with the boxed message class that Aylong designed, as shown below.
| Old |image|
|------------|---|
| New |image|

This PR also adds a blinking red background to the vote icon that is displayed in the top left corner of the screen whenever a vote is called.
dreamseeker_R8TtiRICIA

## Why It's Good For The Game

There have been instances where a player has been distracted, or may have missed the vote notification sound. These visibility changes may help increase voter turnout in some of these instances.

## Changelog

:cl: Aylong
add: Added a blinking background icon to vote action buttons
add: Added boxed message blocks and fieldset message blocks
add: Added a boxed message block for votes and a fieldset message block for vote results
/:cl:


Aquidu:
goated pr
this-peaks-memes

code/__DEFINES/chat.dm
```diff
@@ -37,3 +37,9 @@
 #define debug2_world_log(msg) if (GLOB.Debug2) log_world("DEBUG: [msg]")
 /// Adds a generic box around whatever message you're sending in chat. Really makes things stand out.
 #define examine_block(str) ("<div class='examine_block'>" + str + "</div>")
```
thgvr: This and boxed message should not both exist at the same time. Rename one or the other, whatever's easiest for you.

zimon9: Alright, I'll swap instances of the examine block with the boxed message block

zimon9: Should be in now. I was worried about breaking something, so I opted to just rename the boxed_message function to the examine_block function. I'll update the PR body with the current changes. Thank you for the review, and let me know if there's anything else that needs to be changed.

code/__DEFINES/chat.dm
```diff
@@ -36,4 +36,8 @@
 /// Used for debug messages to the server
 #define debug2_world_log(msg) if (GLOB.Debug2) log_world("DEBUG: [msg]")
 /// Adds a generic box around whatever message you're sending in chat. Really makes things stand out.
-#define examine_block(str) ("<div class='examine_block'>" + str + "</div>")
+#define examine_block(str) ("<div class='boxed_message'>" + str + "</div>")
+/// Adds a box around whatever message you're sending in chat. Can apply color and/or additional classes. Available colors: red, green, blue, purple. Use it like red_box
+#define custom_boxed_message(classes, str) ("<div class='boxed_message " + classes + "'>" + str + "</div>")
```
thgvr, line 41: `custom_examine_block` instead, same for the div class here and in the .scss file

zimon9: Perhaps I shouldn't have been lazy and I should have properly replaced the examine_block instances in the code with boxed_message blocks, haha

I can do that instead, potentially, if that may be preferred? I can rename everything here too, but I feel like it makes a little more sense to leave the name as boxed_message and custom_boxed_message because they're used for things other than just examines

thgvr: Then you could rename them all, there aren't many instances. It just needs to be consistent

zimon9: I was originally worried about how tgui-panel.bundle.css looked like it depended on how examine blocks were defined the .scss file, but on further examination now, it looks more like this file merely stores the way that the class should render, so my original fears were moot

zimon9: Alright! I'll do that now, thank you

PR #3996, opened 2025-01-13
Reduces the chance of getting hurt from a broken bone if one is walking

zimon9:
## About The Pull Request

This PR reduces the probability to cause damage to a player from 5% per tile walked to 1% per tile walked if the player is in the walk intent mode. If they are in the run intent mode, the probability remains at 5%.

## Why It's Good For The Game

I thought it might be a little flavorful to have it so that the chance to get hurt from a broken bone is reduced if one is walking instead of running, as it would imply that a character is attempting to avoid jostling their broken bone(s) around. I'm not sure if the probabilities should be tweaked a little, however.

## Changelog

:cl:
balance: The bone breakage damage proc now depends on the movement intent of a carbon. Walking reduces the proc call probability
/:cl:


code/modules/surgery/bodyparts/bodyparts.dm
```diff
@@ -802,7 +802,7 @@
 	if(bone_status != BONE_FLAG_BROKEN || !owner || istype(owner?.buckled, /obj/structure/bed/roller))
 		return
 
-	if(prob(5))
+	if((prob(5) && (owner.m_intent == MOVE_INTENT_RUN)) || (prob(1) && (owner.m_intent == MOVE_INTENT_WALK)))
```
west3436: Someone with more Byond knowledge should weigh in but I believe a switch with owner.m_intent is faster than this which is important since this is called every time a mob moves.

SomeguyManperson:
```suggestion
	if(prob(owner.m_intent == MOVE_INTENT_RUN ? 5 : 1))
```

zimon9: I like this, it's far cleaner. Thank you.

zimon9: I'm actually uncertain of this myself, and it is certainly worth evaluating. I know that in javascript, at least, switches can be more performant, especially when there are a lot of cases

PR #4002, opened 2025-01-14
Gathering now checks if the item can be reached

Ical92:
## About The Pull Request

You can no longer gather (e.g. with a plant bag or trash bag) from any distance/through walls/through cameras/boundless thievery.

## Why It's Good For The Game
Unfortunately it is funny but I think you shouldn't be able to steal things by looking at it through a camera.

## Changelog

:cl:
fix: gathering no longer has infinite reach
/:cl:


Ratvarr:
how the grinch stole shiptest. 
another feature removed. another Christmas saved.

code/datums/components/storage/storage.dm
```diff
@@ -52,6 +52,7 @@
 	var/allow_quick_gather = FALSE					//allow toggle mob verb which toggles collecting all items from a tile.
 
 	var/collection_mode = COLLECT_EVERYTHING
+	var/collection_range = 1 						//range that things can be picked up from, in view
```
FalloutFalcon: useless var.

PR #4040, opened 2025-01-19
Attachment Update + A few new attachments

Gristlebee:
## About The Pull Request

Adds two new lists to help handle initialization for valid attachments on gun's attachment holder lists, weapon_class_attachments (for attachments specific to a gun type, eg scopes for rifles) and unique_attachments (for attachments specific to that gun eg. folding stocks on the saber) and refused_attachments (attachments the gun can't take.)

Adds secondary actions to items, currently bound to shift-space by default. Using secondary action on a gun will cycle through it's firemodes, if any.

Underbarrel guns now support hand unloading and magazines.

Underbarrel guns now come in guncases.

Underbarrel shotgun is now pumped, holds two shots, and had it's price increased to 1200.

Adds the Hognose underbarrel pistol, alof tube attachment and underbarrel energy cannon.

The Hognose and alof can be bought in outpost cargo, and the underbarrel energy cannon can be bought in the black market.

Fixes being unable to load underbarrel guns while weapons with bolts like illestren were closed.

Energy guns have a muzzle slot so they can take bayonets properly.

Pistols can no longer take underbarrel guns.

Fixes a few typos

## Why It's Good For The Game

Adding new attachments to the game isn't great right now since it operates off a single list. If a gun needs to have a different list of attachments it can take, it needs a snowflaked valid attachment list that needs to updated along with everything else whenever a new attachment is added which adds a lot of work. Organizing the lists into different categories should hopefully make things less messy to work with.

Underbarrel guns now come in cases for consistency.

The firemode hotkey was added due to feedback about how awkward it was to quickly swap firemodes to use an underbarrel gun, especially weapons with multiple firemodes since you had to moouse up and click on a HUD element. Having a hotkey should be easier.

The underbarrel shotgun was changed to make it feel better to use after feedback.

New attachments are cool I think, and work as proof of concept for the new mechanics in the update.
## Changelog

:cl:
add: Hognose, alof, underbarrel energy cannon
add:  Underbarrel guns come in cases.
add: Secondary actions and hotkey.
add: Muzzle slot for eguns
balance: Pistols can no longer take underbarrel guns.
balance: Underbarrel shotgun holds two rounds, price increased to 1200
fix: Fixes being unable to load underbarrel guns while gun bolts were closed.
fix: slings being funky
spellcheck: typos
code: new attachment lists, weapon_class_attachments, unique_attachments, refused_attachments
/:cl:


code/modules/projectiles/gun.dm
```diff
@@ -268,8 +268,14 @@
...
 */
-	///The types of attachments allowed, a list of types. SUBTYPES OF AN ALLOWED TYPE ARE ALSO ALLOWED
+	///The types of attachments allowed, a list of types. SUBTYPES OF AN ALLOWED TYPE ARE ALSO ALLOWED.
 	var/list/valid_attachments = list()
+	///The types of attachments allowed for this type of gun. Eg. Pistols/Rifles/Shotguns/etc.
+	var/list/weapon_class_attachments = list()
+	///The types of attachments that are unique to this gun. Adds it to the base valid_attachments list. So if this gun takes a special stock, add it here.
+	var/list/unique_attachments = list()
+	///The types of attachments that aren't allowed. Removes it from the base valid_attachments list.
+	var/list/refused_attachments
```
FalloutFalcon: im not a big fan of this as its 3 lists that to the same thing just for different parts of the item hierarchy. If we have weapon_class attachments and unique_attachments i don't see a need for valid attachments. I get your goal is the reduce on repeated lists but i would argue it makes it harder to work with attachments.  At the very least drop unique attachments and use valid for that purpose.

FalloutFalcon (changes requested):

FalloutFalcon (approved):

Erikafox:
I would like pistols to refuse underbarrel attachments.

PR #4071, opened 2025-01-24
Increase Drill Mission Value

Erikafox:
## About The Pull Request

As title. Pairs with #4069 

## Why It's Good For The Game

avoid the problem of creating a money sink.

## Changelog

:cl:
balance: drill missions now pay more money.
/:cl:


zimon9:
I'm wondering if it would be worth increasing the payout of class 1 drills, as well? When this was testmerged, I noticed that people tended to ignore class 1 missions, letting them clutter up the board (and understandably so, as it wouldn't have been a good time investment to pursue them)

PR #4080, opened 2025-01-25
Makes indies, inteq and SRM use combustion thrusters instead of plasmas.

Sadhorizon:
## About The Pull Request
Alongside replacing the thrusters I also:
- **Kilo, Mudskipper, Talos, Elder** - replaced plasma canisters with fuel canisters.
- **Riggs, Colossus** - replaced plasma pressure tanks with fuel canisters (~~as for Riggs- I couldn't really find the room to add fuel mixers, though it should get them I think - will think a bit more about it~~ yeah I dunno)
- **Shetland** - I went a little overboard and made the TEG run on a hydrogen/oxygen mix, from which the thrusters feed. Should work about the same though. I hope so anyways.

Also rotates the AACs on all of the above cuz idk, I'm touching them already, might as well do it.
Also does like really minor Kilo fixes that were annoying me for a while. (removes a wall decal, rotates beds)

## Why It's Good For The Game
I've been told by @Apogee-dev that plasmas should be only used on NT ships. This fixes that.

## Changelog

:cl:
tweak: Kilo, Mudskipper, Riggs, Shetland, Colossus, Talos and Elder now use combustion thrusters instead of plasma thrusters.
tweak: Shetland's TEG now runs on hydrogen instead of plasma.
fix: Rotated the AACs on all of the above ships.
/:cl:


MemeSnorfer:
Random thing about the Talos, actually. It _does_ actually have plasma canisters in the maintenance wings, ~~one on each wing, for each thruster~~ (so it turns out one of the canisters were disappeared in the maints remap for the Talos). It's a small thing, but I imagine the thrusters were the whole reasoning for their existence in maints.

Sadhorizon:
> Random thing about the Talos, actually. It _does_ actually have plasma canisters in the maintenance wings, ~one on each wing, for each thruster~ (so it turns out one of the canisters were disappeared in the maints remap for the Talos). It's a small thing, but I imagine the thrusters were the whole reasoning for their existence in maints.

Accidently removed one in a previous PR- I'll just... give it two fuel ones again.

Sadhorizon:
Oh and uh- please make sure the Shetland was tested before it's merged, I'm not an expert with atmos and I totally could have done it incorrectly/missed something important.

SomeguyManperson:
Thrusters on the riggs are still plasmas

Sadhorizon:
> Thrusters on the riggs are still plasmas

I... probably ctrl+z'd it at some point by accident. Oops!

PR #4093, opened 2025-01-26
Reverts Frontier Armour Vests back to being Bulletproof + Slightly increases Rusted Red armour

generalthrax:
## About The Pull Request

Frontie armour vests are bulletproof again and Rusted Reds have slightly increased armour values.

Adds a variant of frontie vests that the simple mobs drop, and the mobs use this variant for armor_base

## Why It's Good For The Game

It occurred to me that pirates should probably have decent armour after seeing Rylie switch out their normal vests for bulletproof during a big convoy event two months ago. Rusted Red suits have increased armour for the same reason (theyre still slow) and this will make the ramzi simplemobs a bit tougher (not nearly as much as when they had 50 so i might adjust it to be slightly lower) so We'll See

## Changelog

:cl:
balance: Frontier armour vests are bulletproof again.
balance: Simple mobs drop a version of the vest that has normal armour vest values, player pirates and the black market version are unaffected. Frontie mobs inherit this armour.
balance: The Rusted Red hardsuit has increased armour. Ramzi mobs are tougher as a result.
fix: Revert some of the health changes from Dynamic Mobs
/:cl:


Apogee-dev:
I feel like there's a way better solution for this than having mob armor values not match their visuals and drops: Add a higher-quality Frontie vest (or, alternatively, a lower quality Frontie vest) and give mobs the weak one and player pirates the better one. Even if that just means using the same sprites for both vests until a dedicated sprite can be whipped up, it’d be an improvement since stats would actually line up between what's going on behind the scenes and what players actually can see.

rye-rice (approved):

Zevotech:
> Add a higher-quality Frontie vest (or, alternatively, a lower quality Frontie vest) and give mobs the weak one and player pirates the better one. 

Thissss. Plus it would discourage players from wearing pirate armor looted off of corpses/ruins so the "its better than what i had" excuse doesnt work.

Apogee-dev:
> > Add a higher-quality Frontie vest (or, alternatively, a lower quality Frontie vest) and give mobs the weak one and player pirates the better one.
> 
> Thissss. Plus it would discourage players from wearing pirate armor looted off of corpses/ruins so the "its better than what i had" excuse doesnt work.

It gives admins more discretion in which version to give player-controlled pirates too. There’s definitely situations in which you dont actually want to give everybody bulletproofs.

generalthrax:
Completely forgot I was gonna do this. Should be done. Also reversed a few health changes that Dynamic mobs did for Some Reason

PR #4099, opened 2025-01-28
Atmospheres Refactor Take 2

MarkSuckerberg:
## About The Pull Request

Re-attempt of #3878.

Basically makes atmosphere gas string generation a lot better. The previous method was baffling to me and it bugged me to no end. This was intended to be part of a bigger PR but I might as well PR this as it is now.

Calculations of randomized atmos mixes **WILL** be different now, but I tried to make it as similar to the old method as I could without bending over backwards. We'll get used to it, unless it's bugged.

Also gets rid of `initial_temperature` on turfs, because it literally did nothing. Specify temperature as part of `initial_gas_mix`.

## Why It's Good For The Game

Simplifies an EXTREMELY needlessly complicated process that did something as simple as generating a "random" gas string for planets to use. Allows me to feel less bad when I generate a lot more unique atmosphere types.

## Changelog

:cl:
/:cl:


MarkSuckerberg:
Also updated and fixed.

MarkSuckerberg:
still ready

PR #4100, opened 2025-01-28
Fixes AI Movement Runtime

MarkSuckerberg:
## About The Pull Request
Should fix the AI movement runtime that apparently is happening an absolute ton. AI controllers never removed themselves from the movement processing lists when deleted, causing them to get hard deleted and replaced with nulls in the list. This fixes that.

Supersedes #4082

## Why It's Good For The Game
wowie that's a lotta runtimes

## Changelog

:cl:
/:cl:


FalloutFalcon (approved):
augh so true.

PR #4103, opened 2025-01-29
Fixes a runtime with SecurEye and updates bodycamera descriptions

zimon9:
## About The Pull Request

This PR should fix a runtime that had been popping up with bodycameras during their testmerges. It should also fix part of the description for them that I'd erroneously made bold.

## Why It's Good For The Game

Runtimes are not good, and the description was a little confusing in what words were made bold. A single missing character causes untold havoc yet again.

## Changelog

:cl:
fix: fixed securEye runtime
fix: fixed bodycamera description typo
/:cl:

### DIFF
--- a/code/modules/modular_computers/file_system/programs/secureye.dm
+++ b/code/modules/modular_computers/file_system/programs/secureye.dm
@@ -173,7 +173,11 @@
 
 	var/list/visible_turfs = list()
 
-	if(!active_camera.loc)
+	if(!active_camera)
+		show_camera_static()
+		return
+	else if (active_camera.loc == null)
+		show_camera_static()
 		return
 
 	var/cam_location = active_camera.loc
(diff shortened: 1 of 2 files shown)


code/modules/modular_computers/file_system/programs/secureye.dm
```diff
@@ -173,7 +173,11 @@
 
 	var/list/visible_turfs = list()
 
-	if(!active_camera.loc)
+	if(!active_camera)
+		show_camera_static()
+		return
+	else if (active_camera.loc == null)
+		show_camera_static()
```
FalloutFalcon, line 180: any reason not to slap an OR here instead?

zimon9: The reason it's staggered like this is because if no active camera is selected, the reference would be set to null, and this was causing runtimes when active_camera was null (because null.loc would throw a runtime)

This should check if active_camera is not null first, and _then_ evaluate if the location is null

Truth be told, I could've made lines 179 through 181 their own if function instead of chaining it up with the if function starting at line 176, but the behavior is functionally the same

FalloutFalcon: ok gotcha sounds fine

PR #4119, opened 2025-02-01
Gun misfires now target only the leg again

firebudgy:
## About The Pull Request

Gun misfires now only target the left leg as they were originally coded. 

## Why It's Good For The Game

Originally, the purpose of misfires was to encourage through gameplay mechanics that people perform gun safety and targeting the leg was so that it was able to be destructive but not cause things that are incredibly difficult to treat or in some cases instantaneously kill you.

This brings it back to that original purpose of being a mechanic-encouraged gun safety feature and not a "gotcha" for stepping on an ice tile with an unsafed firearm.

## Changelog

:cl:
balance: Misfires only go off in the your leg again.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -1002,10 +1002,9 @@
 			if(at_risk == human_holder.s_store)
 				chance_to_fire = round(GUN_NO_SAFETY_MALFUNCTION_CHANCE_LOW * at_risk.safety_multiplier)
 		if(at_risk.safety == FALSE && prob(chance_to_fire))
-			var/bodyzone = pick(BODY_ZONE_HEAD, BODY_ZONE_CHEST, BODY_ZONE_L_ARM, BODY_ZONE_R_ARM, BODY_ZONE_L_LEG,BODY_ZONE_R_LEG)
-			if(at_risk.process_fire(src,src,FALSE, null, bodyzone) == TRUE)
+			if(at_risk.process_fire(src,src,FALSE, null,  pick(BODY_ZONE_L_LEG,BODY_ZONE_R_LEG)) == TRUE)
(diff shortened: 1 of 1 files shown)


Erikafox:
>mechanic-encouraged gun safety feature

And how do we do this?
perhaps with a 'gotcha' for not turning your safety on?

generalthrax:
There's plenty of insta-kill gotchas already from things like 5 pixel claymores on certain ruins, getting a leg broken is significantly preferable to trying to argue semantics on how having safeties off meaning you have a 6% chance to rotate your gun 270 degrees and firing a Trabuco into your skull killing both you and your entire crew instantly is mechanically interesting in any scenario.

Erikafox:
You're entirely right firing a trabuco into your leg, killing you and your entire crew instantly is much more mechanically rich

generalthrax:
Point stands.

thgvr:
As a solution, how about #4120 's implementation with the odds of hitting anything but leg lowered to around 1% and head even lower than that. Personally, I think having some "horrible accident" sort of mechanics that are incredibly rare (rolling a 1 on a dnd skillcheck) add a lot to the experience but I can see how some might disagree.

Regardless of that as of now the odds in #4120 are way too high and this pr is preferable

kaylexis:
I've always been at odds with the goof factor of gun safety, and i do think changing the chances of it hitting anywhere but the legs is good, but a lot of the hate towards that could be fixed by having a way to temporarily fix chest and head fractures with a splint or gauze. because currently there's no way to hold out with a chest/head fracture other than not moving at all

FalloutFalcon:
closing in favor of https://github.com/shiptest-ss13/Shiptest/pull/4120 .. sorry for closing two of your prs in like 10 minutes.

PR #4146, opened 2025-02-06
Scrapping Charges in Cargo

firebudgy:
## About The Pull Request

Adds a C-4 duffel to cargo (10 charges at 100 a charge), at the cost of 1000 credits.
C-4 itself is surprisingly not as destructive as one might think, and in the things one performs in shiptest it could be particularly useful. Breaching a wall in a ruin is a big one, or blowing open safes (did you know you could do that with C4?)

C-4 is already available in a few places and even on certain military ships roundstart, so it wouldn't be too out of place.

This PR also makes it so you only need one C-4 charge to blow the lock on a safe, to make them another method to bypass the lock. Three charges on what was effectively rare to find explosives just to pop open a safe was a bit much. (Refactoring was done by @FalloutFalcon, thanks Fallcon!)

## Why It's Good For The Game

This adds additional gameplay options for clearing ruins, excavating, breaking open safes, and a number of other things. With a do_after to place them, a 10 second timer and a very visible overlay on items, it's hard to use them stealthily either. 

## Changelog

:cl: Cloudbreak, FalloutFalcon
add: Demolitions charges to Cargo
balance: Safes only take one C-4 charge to explode
/:cl:

### DIFF
--- a/code/game/objects/structures/safe.dm
+++ b/code/game/objects/structures/safe.dm
@@ -7,7 +7,7 @@ FLOOR SAFES
 /// Chance for a sound clue
 #define SOUND_CHANCE 10
 /// Explosion number threshold for opening safe
-#define BROKEN_THRESHOLD 3
+#define BROKEN_THRESHOLD 1
 
 //SAFES
 /obj/structure/safe
@@ -92,10 +92,6 @@ FLOOR SAFES
 		explosion_count++
 		switch(explosion_count)
 			if(1)
-				desc = initial(desc) + "\nIt looks a little banged up."
-			if(2)
-				desc = initial(desc) + "\nIt's pretty heavily damaged."
-			if(3)
 				desc = initial(desc) + "\nThe lock seems to be broken."
 
 /obj/structure/safe/ui_assets(mob/user)
--- a/code/game/objects/structures/safe.dm
+++ b/code/game/objects/structures/safe.dm
@@ -6,8 +6,6 @@ FLOOR SAFES
 
 /// Chance for a sound clue
 #define SOUND_CHANCE 10
-/// Explosion number threshold for opening safe
-#define BROKEN_THRESHOLD 1
 
 //SAFES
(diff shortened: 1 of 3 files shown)


FalloutFalcon (changes requested):
At that rate I would just refactor BROKEN_THRESHOLD to a bool.

firebudgy:
> At that rate I would just refactor BROKEN_THRESHOLD to a bool.

i honestly have no idea how. i can drop it from the PR if it's an issue i'm not married to it

code/game/objects/structures/safe.dm
```diff
@@ -7,7 +7,7 @@ FLOOR SAFES
 /// Chance for a sound clue
 #define SOUND_CHANCE 10
 /// Explosion number threshold for opening safe
-#define BROKEN_THRESHOLD 3
+#define BROKEN_THRESHOLD 1
```
FalloutFalcon: Remove this

code/game/objects/structures/safe.dm
```diff
@@ -92,10 +92,6 @@ FLOOR SAFES
 		explosion_count++
```
FalloutFalcon: `if(((severity == 2 && target == src) || severity == 1) && !safe_broken)
		safe_broken = TRUE
		if(safe_broken)
			desc = initial(desc) + "\nThe lock seems to be broken."`

firebudgy:
> > At that rate I would just refactor BROKEN_THRESHOLD to a bool.
> 
> i honestly have no idea how. i can drop it from the PR if it's an issue i'm not married to it

Got help in dev-gen, should be good now.

firebudgy:
Review is addressed by the way.

FalloutFalcon (approved):

code/game/objects/structures/safe.dm
FalloutFalcon: Oops didnt realize i had pending comments

PR #4149, opened 2025-02-07
Change Mob Neutering

Erikafox:
## About The Pull Request

changes simplehuman drop neutering to support a prob chance instead of being binary y/n

## Changelog

:cl:
code: neutered var on .../hostile/human removed in favor of weapon_drop_chance
balance: default chance of a simple human dropping their gun is now 30%
/:cl:


Apogee-dev:
honestly if we could figure out a way for this to support neutering armor drops by chance too and good drop rates for that it’d be peam

FalloutFalcon:
> honestly if we could figure out a way for this to support neutering armor drops by chance too and good drop rates for that it’d be peam

Would be a tiny bit harder cause you would have to prevent the vest spawning on the corpse

code/modules/mob/living/simple_animal/hostile/human/human.dm
```diff
@@ -46,8 +46,8 @@
 	var/r_hand
 	/// Path of the left hand held item we give to the mob's visuals.
 	var/l_hand
-	// If we drop l and r hand loot
-	var/neutered = FALSE
+	// Prob of us dropping l/r hand loot.
+	var/weapon_drop_chance = 30
```
FalloutFalcon, line 50: Pretty big balance change. Document it or make it 100.

FalloutFalcon (changes requested):

FalloutFalcon:
I would say we should phase out netured mobs and just balance our mobs around there built in drop chance however its out of scope of this pr

FalloutFalcon (approved):

PR #4160, opened 2025-02-09
Changes Some Recipes

Erikafox:
thanks obama

## Changelog

:cl:

fix: updates recipes for dimorlin (so you can make it), shoalmix (so you can make it) and tramal (so it works as intended)
/:cl:


Zevotech:
thanks kiirv-waha

PR #4172, opened 2025-02-11
Contract list now scales with spawned vessels

Burning02:
## About The Pull Request

As it says in the title, outpost contracts will now scale per vessel in-sector by an addition of 5, with some coding help from Zimon (I did something incredibly stupid)

## Why It's Good For The Game

You remember those rounds you had, where you had to share 15 contracts with close to 10 or more vessels in-round? This makes it so you won't have worry about having to share 15 contracts for a round with a high vessel count.

## Changelog

:cl:
balance: The outpost contract limit will now scale with the amount of vessels in-round
/:cl:


FalloutFalcon:
I think this is probably too many to display during high pop rounds with the current UI but also I will prob fix it in dynamic missions so its whatever.

FalloutFalcon (approved):

FalloutFalcon:
I also just realized the mission count config does nothing

code/modules/overmap/objects/outpost/outpost.dm
```diff
@@ -141,6 +141,7 @@
 	return "[person_name] [pick(GLOB.station_suffixes)]"
 
 /datum/overmap/outpost/proc/fill_missions()
+	max_missions = 15 + (SSovermap.controlled_ships.len * 5)
```
FalloutFalcon:
```suggestion
	max_missions = 10 + (SSovermap.controlled_ships.len * 5)
```

FalloutFalcon (changes requested):

code/modules/overmap/objects/outpost/outpost.dm
Burning02: Oh, didn't even see this lol, sure I'll do that instead

FalloutFalcon (approved):

PR #4182, opened 2025-02-12
Magazine Changes in Cargo

firebudgy:
## About The Pull Request

Slightly lowers the prices of several extended magazines in cargo, given that their sheer price has made them simply not economic to use. Some of these were a bit crazy given how highly priced they were and after asking a few people, they avoided buying them due to their high price.

Also removes the Woodman 10-round magazine from Cargo.
## Why It's Good For The Game

In regards to the Woodsman Magazine Removal, as a semi-automatic rifle firing one of the highest-damaging rounds in the game that can take a scope, it effectively invalidates the Illestren in it's niche. The Illestren's high-power is specifically balanced by it's ammo capacity, so this helps put them a bit more on a level playing field. This does NOT remove it from the game, however, for use in whereever it might be needed. 

Other magazine changes were made to make them options worth considering rather than just packing two more magazines every time.

## Changelog

:cl:
balance: Extended magazine prices reduced slightly.
balance: Woodsman magazine removed from cargo.
/:cl:

### DIFF
--- a/code/modules/cargo/packs/magazines.dm
+++ b/code/modules/cargo/packs/magazines.dm
@@ -61,7 +61,7 @@
 	name = "Firestorm Pan Magazine Crate"
 	desc = "Contains a 40-round pan magazine for the Hunter's Pride Firestorm SMG."
 	contains = list(/obj/item/ammo_box/magazine/c44_firestorm_mag/pan/empty)
-	cost = 1000
+	cost = 700
 	faction = /datum/faction/srm
 
 
@@ -89,12 +89,6 @@
 	name = "Woodsman Magazine Crate"
 	desc = "Contains an 8x50mmR magazine for the Woodsman Rifle, with a capacity of five rounds."
 	contains = list(/obj/item/ammo_box/magazine/m23/empty)
-	cost = 200
-
-/datum/supply_pack/magazine/woodsman_mag_extended
-	name = "Woodsman Magazine Crate"
-	desc = "Contains an 8x50mmR magazine for the Woodsman Rifle, with a capacity of ten rounds."
-	contains = list(/obj/item/ammo_box/magazine/m23/extended/empty)
 	cost = 500
 
 /datum/supply_pack/magazine/m20_auto_elite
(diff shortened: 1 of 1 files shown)


code/modules/cargo/packs/magazines.dm
```diff
@@ -89,12 +89,6 @@
 	name = "Woodsman Magazine Crate"
 	desc = "Contains an 8x50mmR magazine for the Woodsman Rifle, with a capacity of five rounds."
 	contains = list(/obj/item/ammo_box/magazine/m23/empty)
-	cost = 200
```
Erikafox, line 92: is this intended?

Erikafox (commented):
I don't agree with the logic here.

I think expensive magazines are expensive for good reason - it increases how long you can function in combat without needing to resupply. If you don't exploit that, then they increase the amount of other equipment you can take relative to the same load of ammo. 

They're less a cost efficiency buy and a more space efficiency buy.

Especially in regards to the SKM until apogee and I go through the ammo prices (hopefully soon). 

(also please specific *which* woodsman magazine is being removed - although I'd prefer looking at its ammo capacity and how it actually performs, as the woodsman and illestren fill different niches).

Apogee-dev:
The extended mags being uneconomic to use is intended. When they were first added to cargo we generally priced them at double the price of normal mags (depending on how much the extra capacity actually is) plus 25-50% extra- If it were just double price or god forbid _less,_ which is what this PR does in most cases, there would be no reason to buy small mags in the first place. The extra is the price of convenience and saving inventory space. 

Extended mags Woodsman being just plain better than the Illestren is also intended- because using it means paying several times more for the rifle and mags than you would for an Illestren. It’s like saying a Cobra is better than a Candor- it is, and purposefully so. Illestren, like the Flaming Arrow, is a low-tier budget gun that isn’t meant to compete with pricier options. It is instead meant to allow dirt-poor players access to a weapon with good armor penetration on a shoestring budget.

That being said, if Woodsman exmags are just plain too strong all around, nerf them to 8 rounds instead of removing them from cargo.

firebudgy:
Sure, yeah, I'll drop the price changes. But I do think the Woodsman's extended magazine should be at the same price point as other marksman rifles' extended magazines.

PR #4191, opened 2025-02-13
buckshot ranging test pr

Erikafox:
:cl:
balance: buckshot now loses ap as it travels. starts at -20 ap
/:cl:


Apogee-dev:
Starting at 0 ap is a _huge_ buff honestly. I don’t think point blank buckshot should be as  capable against armor as that honestly- exosuits and high grade armor like elite suits and bulletproof vests should still invalidate it tbh. Also isnt that effectively a stealth slugs nerf? Starting at -20 ap (the same as handguns) would be more fair i think

Erikafox:
I think they still do - they'll just invalidate it less like this at point blank.
This is a test anyways so we'll see how it feels lol.

I could adjust slugs but the interest here is how the pellets feel.

Apogee-dev:
> I think they still do - they'll just invalidate it less like this at point blank. This is a test anyways so we'll see how it feels lol.
> 
> I could adjust slugs but the interest here is how the pellets feel.

I just think starting at -20 (the same as handguns) would be more fair and make sense tbh. Shotgun pellets aren’t going to outpenetrate 9mm even at point blank.

FalloutFalcon:
thoughts on the pr?
seems fine to me 
image

PR #4195, opened 2025-02-15
Saber Magazines no longer faction locked

firebudgy:
## About The Pull Request

9mm SMGs (often used as ruin loot) no longer have their magazines faction-locked. 

## Why It's Good For The Game

The basic magazines of a weapon probably shouldn't be locked behind a faction, especially when that weapon isn't super powerful.

## Changelog

:cl:
balance: SMG magazines for the Saber are now available.
/:cl:

### DIFF
--- a/code/modules/cargo/packs/magazines.dm
+++ b/code/modules/cargo/packs/magazines.dm
@@ -18,11 +18,8 @@
 	name = "9mm SMG Magazine Crate"
 	desc = "Contains a 9mm magazine for the Vector and Saber SMGs, with a capacity of thirty rounds."
 	contains = list(/obj/item/ammo_box/magazine/smgm9mm/empty)
-	cost = 300
+	cost = 250
 	faction = /datum/faction/nt
-	faction_discount = 0
-	faction_locked = TRUE
-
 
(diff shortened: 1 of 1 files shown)


Apogee-dev:
eh… just because the saber appears in ruins doesn’t make it any less of a faction weapon than a Mongrel or CM-5 is. yeah getting more mags is tough but it’s also a random gun you picked up off the ground, not something you can just go out and buy. that kind of comes with the territory. you can’t buy mags for any of the _other_ faction loot guns either, with pretty much the sole exception being the cobra because there’a a nerfed civvie version on the market.

Erikafox:
you can buy most faction weapon magazines now, actually.

Apogee-dev:
Buy on the open market or buy in faction cargo? If most are open market already then sure, but as far as I knew that was only the case for guns that are already open market or have open market relatives, like the cobra and boomslang.

Erikafox:
open cargo. extended/special mags are faction.

Apogee-dev:
> open cargo. extended/special mags are faction.

We _could_ code in a 20 round mag for the saber and make the fullsize one faction locked…

PR #4213, opened 2025-02-17
Fixes mobdrop item names + frontiersmen are injured on death + Frontiersman shredder mob

Erikafox:
## Changelog

:cl:
fix: mob items no longer show the path when dropping
add: the frontiersman lmg-man now carries a shredder instead of a laser minigun. prepare yourself.
/:cl:

### DIFF
--- a/_maps/RandomRuins/IceRuins/icemoon_ice_lodge.dmm
+++ b/_maps/RandomRuins/IceRuins/icemoon_ice_lodge.dmm
@@ -1161,6 +1161,14 @@
 /obj/effect/decal/cleanable/dirt,
 /turf/open/floor/wood/maple,
 /area/ruin/powered/icemoon/lodge/mainhall)
+"rc" = (
+/mob/living/simple_animal/hostile/human/frontier/ranged/trooper/heavy/internals/neutered{
+	aggro_vision_range = 12
+	},
+/turf/open/floor/wood/icemoon{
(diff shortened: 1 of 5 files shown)


Apogee-dev:
Shouldn’t we just replace the laser minigun on these mobs with the shredder? it has the feel of being a placeholder to me and the laser minigun is getting replaced with sharplite’s laser lmg later down the line

Erikafox:
yeah sure

Erikafox:
done

Apogee-dev:
Second question: Should shredder NPCs use slugs or buckshot? Both are horrific, but Shredders issued to players for events generally come with slugs by default, and the damage profile should be a bit more similar to the laser minigun it’s replacing. I’m also not sure if player armor would neuter buckshot too hard to make them the horrifying threat they should be. due to scatter and limb damage maybe it wouldn’t. Probably worth testing both honestly and going with the more consistently dangerous one.

FalloutFalcon (approved):

generalthrax:
I will also add that the Frontiersmen Lodge ruin will probably have to be edited slightly since there is a laser gunner behind a room with glass that is Significantly, Significantly less dangerous since he cant actually hit people coming up

Probably either move him outside or keep him as the Only laser gunner Ever

You probably want to replace the ruin frontie heavies with neutered ones too since player accessible shredders sounds Bad

PR #4246, opened 2025-02-22
EHF Point Beacons (Planet Preservation)

Ical92:
## About The Pull Request
This PR adds beacons for planetary use, to preserve, rename, and send a distress beacon from planets. The beacon needs to be secured and power similarly to a floodlight to be used. An inserted ID is required to send a distress beacon, and an ID from an officer is needed to rename a planet.

Also added: IDs now have a boolean if it belongs to an officier job, and overmap objects now have a proc to change if it's preserved. Also, overmap tokens now properly show if the planet is being preserved.

image

image

image

## Why It's Good For The Game
Allows people to do more on planets than just mining and ruin clearing. Setting up outposts, colonies, or any other semi-permanent structure, while previously possible was difficult because of planets being unloaded when nobody is on them. Meaning crews would either need to have someone stay behind, or prepare everything they may need before landing. This PR lets crews preserve planets after some initial setup (requiring some amount of power, though not neccessarily an APC), as well as additional features benefitting construction on planets.

## Changelog

:cl:
add: EHF point beacons, which can be used to preserve planets
/:cl:


kaylexis:
monumental for colony gameplay, itd be cool to get a docking port creator too. but that'd need to wait for OM5 because of the lack of a docking selector currently

Apogee-dev:
i was under the impression that we weren't going to support colony gameplay, what with the fact that they tie up a planet spawn slot and still suffer from only having 2 landing spots

Imaginos16:
> i was under the impression that we weren't going to support colony gameplay, what with the fact that they tie up a planet spawn slot and still suffer from only having 2 landing spots

I think it would be a cool idea since I know the playerbase does enjoy colonizing every so often. While we aren't supporting colonizing in the sense of giving them an RnD suite and the sort, it's not like we couldn't do this system already with just a person staying in the planet while everyone else goes out traveling. The beacon would remove that awkward barrier of entry.

Apogee-dev:
> > i was under the impression that we weren't going to support colony gameplay, what with the fact that they tie up a planet spawn slot and still suffer from only having 2 landing spots
> 
> I think it would be a cool idea since I know the playerbase does enjoy colonizing every so often. While we aren't supporting colonizing in the sense of giving them an RnD suite and the sort, it's not like we couldn't do this system already with just a person staying in the planet while everyone else goes out traveling. The beacon would remove that awkward barrier of entry.

I still feel like this should be the purview of admin events, or at least take admin support to acquire one. What happens if two or three ships do construction gimmicks and nobody’s actually flying around? I feel like it’s more likely to curb player interaction than it is to encourage it.

Imaginos16:
> I still feel like this should be the purview of admin events, or at least take admin support to acquire one. What happens if two or three ships do construction gimmicks and nobody’s actually flying around? I feel like it’s more likely to curb player interaction than it is to encourage it.

I am of the opinion that is an extreme case and if anything this'll probably just be a gimmick if nothing else. I suppose an argument could be made of how much content this could actually bring to the table, but I don't think the extreme of three ships doing construction gimmicks teeters more on the territory of unlikely.

code/modules/cargo/packs/machinery.dm
```diff
@@ -207,6 +207,15 @@
 	crate_name = "laser mining drill crate"
 	crate_type = /obj/structure/closet/crate/engineering
 
+/datum/supply_pack/machinery/ehf_beacon
+	name = "EHF point beacon"
+	desc = "A crate containing an EHF point beacon, used to mark points of interest and semi-permanent constructions."
+	cost = 2000 // This sounds reasonable? You'll still need everything else if you want to make an outpost
```
thgvr, line 213: Niche equipment - not sure if this is too expensive or not but 1000 would feel better to me.

code/game/machinery/planet_beacon.dm
```diff
@@ -0,0 +1,216 @@
...
+	if(inserted_id)
+		. += mutable_appearance(icon, "overlay_card")
+
+/obj/machinery/power/planet_beacon/proc/send_distress()
+	if(!inserted_id)
+		return FALSE
+
+	var/datum/overmap/dynamic/OE = get_overmap_location()
+
+	var/text = "A distress beacon has been activated by [inserted_id.registered_name] at local sector co-ordinates [OE.x]/[OE.y]. No further information available."
```
thgvr, line 198: If possible, please convert this to use `proc/create_distress_beacon`. The original proc may need some editing but it should be used here.

code/game/machinery/planet_beacon.dm
```diff
@@ -0,0 +1,216 @@
...
+	priority_announce(text, null, 'sound/effects/alert.ogg', sender_override = "Outpost Distress Beacon System", zlevel = 0)
```
thgvr: If possible, please convert this to use `proc/create_distress_beacon`. The original proc may need some editing but it should be used here.

code/modules/cargo/packs/machinery.dm
thgvr: Niche equipment - not sure if this is too expensive or not but 1000 would feel better to me.

code/game/machinery/planet_beacon.dm
```diff
@@ -0,0 +1,216 @@
...
+			try_put_in_hand(inserted_id, user)
+			inserted_id = null
+			update_icon()
+			playsound(src, 'sound/machines/terminal_insert_disc.ogg', 50, FALSE)
+			return
+
+/obj/machinery/power/planet_beacon/examine(mob/user)
+	. = ..()
+	if(inserted_id)
+		. += "<span class='notice'>Alt-click to eject the ID card.</span>"
```
thgvr:
```suggestion
		. += span_notice("Alt-click to eject the ID card.")
```
Please use the span macros.

This goes for every instance.

Ical92:
Bwuh, where'd they go

code/game/machinery/planet_beacon.dm
```diff
@@ -0,0 +1,216 @@
...
+			add_load(active_power_usage)
+		else
+			change_state(BEACON_ANCHORED)
+	else if(state == BEACON_IDLE) // If idle, use idle_power_usage
+		if(avail(idle_power_usage))
+			add_load(idle_power_usage)
+		else
+			change_state(BEACON_ANCHORED)
+
+/obj/machinery/power/planet_beacon/attackby(obj/item/O, mob/user, params)
```
thgvr: Please do not use single letter vars - `obj/item/O`.

This goes for the whole PR.

Ical92:
Ghost reviews.. I'll get them both fixed though

code/game/machinery/planet_beacon.dm
```diff
@@ -0,0 +1,216 @@
+#define BEACON_UNANCHORED 1
```
thgvr, line 1: Remember to `#undef` your defines at the end of the file if they are not used anywhere else.

code/game/machinery/planet_beacon.dm
```diff
@@ -0,0 +1,216 @@
...
+	desc = "A long-range radio beacon for marking planetoids and points of interest for revisiting."
+	density = TRUE
+	anchored = FALSE
+	clicksound = "terminal_type"
+	use_power = NO_POWER_USE
+	idle_power_usage = IDLE_DRAW_MINIMAL
+	active_power_usage = ACTIVE_DRAW_MINIMAL
+	light_power = 1.5
+	var/state = BEACON_UNANCHORED
+	var/preserving = FALSE // Separate from state to prevent it being unpreserved on distress
```
thgvr:
```suggestion
  /// Separate from state to prevent it being unpreserved on distress
	var/preserving = FALSE
```

Please see the autodoc guide at: https://github.com/shiptest-ss13/Shiptest/blob/383cba0099e9a2528dcce331a40c4262e9d5cf1d/.github/AUTODOC_GUIDE.md?plain=1#L85

thgvr (changes requested):
Sorry for the delay on a review, but this looks good and I'm interested enough to add it.

thgvr (approved):

FalloutFalcon (approved):
Cool concept. Even if it turns out to not be a good choice for cargo it should still be a thing for adminbus.

Ical92:
woe, landmine test ,, I should fix that sometime

PR #4248, opened 2025-02-23
Reduces spawn limit for Mudskipper and Kilo

Apogee-dev:
## About The Pull Request

Mudskipper and Kilo were the only ships with a spawn limit higher than 1, and this was to help ensure smaller ships would be available for players to spawn. Now that several small independent ships are being added to the game, this is less of a concern. Also it might reduce the number of solo ships clogging up the manifest, marginally.

## Why It's Good For The Game

maybe a little less soloshipping would be good for the game. you WILL interact with other players in the multiplayer game

## Changelog

:cl:
balance: Reduced Mudskipper and Kilo spawn limit to 1 each
/:cl:


Ratvarr:
their greed is sickening. they do not need two mudskippers to soloship

Madtankdriver:
Really wish there were like really tiny shuttles you could solo ship so that you would not steal an entire shuttle with more than 3 player slots.

Apogee-dev:
> Really wish there were like really tiny shuttles you could solo ship so that you would not steal an entire shuttle with more than 3 player slots.

we aren’t bringing the pill back, chat
the whole point is to discourage solo shipping so we dont just have like 6+ players just farming ruins alone and then dying

FalloutFalcon (approved):
soloshiping bad. we just added like 2-3 small crew ships. we dont need 2 mudskippers

PR #4264, opened 2025-02-24
DRM and floppy disk installers

FalloutFalcon:
## About The Pull Request
DRM and floppy disk installers for Lifeline and Secureye. Removes them from the net.

## Why It's Good For The Game
Lets us regulate certain programs better and lets us put modcomps and tablets on more ships without having to worry about balance.

## Changelog

:cl:
add: Floppy disk installers with free DRM on all installations of secureye and lifeline.
add: A cargo pack for secureye and lifeline both priced at 250 credits.
del: Removed secureye and lifeline from the net.
/:cl:


ritorizo:
I don't see why secureeye would have DRM. Having it cause no balance issue since the real cost is in having a camera network and lorewise something as antiquated as a camera network would be free software.

FalloutFalcon:
> I don't see why secureeye would have DRM. Having it cause no balance issue since the real cost is in having a camera network and lorewise something as antiquated as a camera network would be free software.

The main reason im locking them up behind DRM is because they circumvent using mapped in consoles which maptainers want to use to enforce ship identity and all that.

MassiveMen:
Why even put it in cargo for dirt cheap. It just as easily acquired as before but now it's slightly inconvenient to get. This has changed nothing balance wise.

FalloutFalcon:
I would consider the 250 creds enough to make a choice to purchase it. Ive seen low mission rolls for 1-2k credits so its not NOTHING but its a low barrier for people who actually want it. This is more to encourage/allow mappers to use the single use console. If someone really wants a modcomp or tablet to come with lifeline or secureye they can also make a preset with it preinstalled.

thgvr:
I'd like some more thoughts on this before merging from whoever

code/modules/cargo/packs/civilian.dm
```diff
@@ -253,6 +253,18 @@
 	crate_name = "spare jumpsuits crate"
 	crate_type = /obj/structure/closet/crate/wooden
 
+/datum/supply_pack/civilian/lifeline
+	name = "A portable drive containing a new install of lifeline"
```
Erikafox:
```suggestion
	name = "Lifeline Disc"
```

code/modules/cargo/packs/civilian.dm
```diff
@@ -253,6 +253,18 @@
 	crate_name = "spare jumpsuits crate"
 	crate_type = /obj/structure/closet/crate/wooden
 
+/datum/supply_pack/civilian/lifeline
```
Erikafox, line 335: add programs to their own cargo category rather than jamming them into civilian

code/modules/cargo/packs/civilian.dm
```diff
@@ -253,6 +253,18 @@
...
+	name = "A portable drive containing a new install of lifeline"
+	desc = "Contains a highly resistant digital rights managment..."
```
Erikafox:
```suggestion
	desc = "Contains a copy of the Lifeline program, and all the associated authorization keys to install and run it."
```

code/modules/cargo/packs/civilian.dm
```diff
@@ -253,6 +253,18 @@
...
+	cost = 250
+	contains = list(/obj/item/computer_hardware/hard_drive/portable/installer/lifeline)
+
+/datum/supply_pack/civilian/secureye
+	name = "A portable drive containing a new install of secureye"
```
Erikafox:
```suggestion
	name = "SecurEye Disc"
```

code/modules/cargo/packs/civilian.dm
```diff
@@ -253,6 +253,18 @@
...
+	desc = "Contains a highly resistant digital rights managment..."
```
Erikafox:
```suggestion
	desc = "Contains a copy of the SecurEye camera monitoring program, and all the associated authorization keys to run it."
```

code/modules/modular_computers/hardware/portable_disk.dm
```diff
@@ -26,3 +26,21 @@
 	power_usage = 40
 	icon_state = "datadisk3"
 	max_capacity = 256
+
+/obj/item/computer_hardware/hard_drive/portable/installer
+	name = "data disk program installer"
+	desc = "Contains a highly resistant digital rights managment..."
```
Erikafox:
```suggestion
/obj/item/computer_hardware/hard_drive/portable/installer
	name = "program installation disc"
	desc = "A small laser-etched disc loaded with a piece of software and the associated authorization keys to operate it."
```

thearbiber:
seems kinda strange to put another buy-in on secureye when it already has buy-ins. Lifelines an odd case in general, we have tools like crew pinpointers that serve the function, but they're not mapped in anywhere.

generalthrax:
Don't think it adds much, personally. Feel like it hinders tools a captain that may be using to lead from the ship (most often on ones with big crews). For example its another layer of extra purchases you need to make to ensure bodycams can be used or if you want to use cameras at all aboard your ship.

I've personally never seen a mapper bring up modconsoles or tablets as a balance concern, myself.

Erikafox:
not entirely sure about this, at least with the specific programs.
Lifeline I can see being a limited thing (as an alternative to a specific tool), but camera systems already have a buy-in

FalloutFalcon:
> seems kinda strange to put another buy-in on secureye when it already has buy-ins. Lifelines an odd case in general, we have tools like crew pinpointers that serve the function, but they're not mapped in anywhere.

Wdym, it has pretty much no buy ins, you can spawn with a laptop that you can install it on for free.

code/modules/cargo/packs/civilian.dm
FalloutFalcon: Doesn't seem right when we only have two programs, Def if we add much more.

Erikafox: better to start now imo

thearbiber:
It has buy-ins in the case of needing to actually buy cameras. Otherwise its completely useless.

thgvr:
I could see making body cameras come in a 2 or 3 pack that includes a disk with the app. (as well as separately)

generalthrax:
Pretty sure above above arguments still stand, for anyone planning on reviewing this

FalloutFalcon:
> I could see making body cameras come in a 2 or 3 pack that includes a disk with the app. (as well as separately)

With cargo becoming increasingly more atomized/granulized I realize that since its already sold separately, i don't see why i would also bundle it into body cameras.

thgvr:
not sure if the way is rn, modcomps need some improvements first

PR #4265, opened 2025-02-25
Reduces minimum timer on C4 from 10 to 0

generalthrax:
## About The Pull Request

Reduces the minimum timer that C4 needs to be attached to something to explode. Do_after to apply it remains unchanged.

## Why It's Good For The Game

Usually imagine C4 as a quick surprise breaching tool compared to slower ones like Grinding but its gotta sit there for 10 seconds and everyone can see it and knows

Still gotta spend time to apply it so it shouldn't cause any problems

## Changelog

:cl:
balance: Changed C4 minimum timer from 10 to 0
/:cl:


Apogee-dev:
couldnt the visibility issue be fixed by just placing the c4 on the tile in front like how we handle posters. and then offsetting the explosion. that way it isnt magically visible from the other side of the wall and we dont need to wrestle with the implications of Instant Boom

generalthrax:
Sounds jank and I want to see how C4 being a quick breaching tool would be like (since its both non-reusable and just Straight Slower Than The Rest, doesnt even destroy doors in one go). Maybe will experiment with it later and see

PR #4266, opened 2025-02-25
Helmet lights and goggles consistency

DIB-DOG:
## About The Pull Request
More helmets can use helmet lights. Helmet lights can be attached and detatched without the use of a screwdriver. Helmets can now take NGR factional goggles and mesons, for sake of consistency.

## Why It's Good For The Game
Noticed that for a long time, the only people using helmet lights were Inteq, and it turned out to be, because their helmets were some of the only ones which **could** do it. This aims to remedy that by increasing flashlight compatability to include all the X11 and M10 variants, as well as a few others. As for the removal of the need for a screwdriver, it didn't quite make sense that it could be attached without a tool but to remove it, you suddenly needed a tool. Overall it's just more convenient and logical. Lastly, for some reason helmets couldn't take NGR or Hardliner goggles? I assume it was oversight, so now our brave Gorlex warriors can carry their goggles on their helmet like the rest of the helmeted folks on the frontier.

## Changelog

:cl:
balance: Helmet lights attach more logically
fix: NGR and Hardliner goggles can fit on helmets
/:cl:


Apogee-dev:
I wouldve gone the opposite direction on the tools side, personally- require a screwdriver to secure the light before you can actually use it. Feels like a bit of a trend to remove tools from equipment modification/use.

DIB-DOG:
> I wouldve gone the opposite direction on the tools side, personally- require a screwdriver to secure the light before you can actually use it. Feels like a bit of a trend to remove tools from equipment modification/use.


Bit of a secondary reason for the Alt-Click change. ATM if you alt click your helmet, you’ll draw the light from its inventory but the light is technically still attached to the helmet, which creates this ghost remote control flashlight situation. Without re-coding how helmet lights work entirely the fix for that boils down to; removing helmet lights all together, removing helmet pockets for goggles and the like, or doing this with the alt click, which disconnects the light as you draw it. 

I had looked into finding a way to prevent the light from being drawn from the helmet once it was placed inside, unless activated with a screwdriver but had no luck on the matter. If someone else feels more up to it, I’d welcome that approach, but in the meantime, this **does** work, and addresses the bug.

PR #4269, opened 2025-02-25
Adds an angle grinder to the Nail

DIB-DOG:
## About The Pull Request
Adds an angle grinding pack, bowman headset, and a pair of welding goggles to the Nail-class boarding shuttle

## Why It's Good For The Game
There had been talk in months past about offering the Crying Sun some ability to have some form of breaching tool but at the time they were not yet coded or implemented. Now that they're done, and have been in the game for some time, this is adds one such tool. There was also an un-used crate just begging to have a purpose so it worked out pretty nice.

As the Nail-class is intended for boarding, among other things, it seems reasonable that it would have some tool by which to facilitate this purpose. It was debated between offering a plasma cutter or breaching explosives but ultimately decided that the angle grinder pack would be the best option. requiring more work and prep-time to utilize.

Screenshot 2025-02-25 150800

## Changelog

:cl:
add: Added a grinder and PPE to the Nail-class boarding shuttle
/:cl:


Gristlebee:
I agree with the idea behind the change. 

Though it was discussed among the maptainer team, and we felt it would be more thematic for the Nail to get C4 as breaching charges. The grinder is more of a salvage tool really. Could you replace the grinder kit with a bag of C4?

thgvr:
@DIB-DOG please address maptainer concerns

Rainrun445:
Having some breaching tools on the Nail would be nice but not placed right into the Nail.

Everyone has access to the nail so some random crewmen would walk off with the breaching tools. Would suggest the tools be added to the Crying Sun armory to be safe and so it isn't always breach ready for when an event just needs the Nail.

thgvr:
This can be changed if a breaching charge of some sort is added, c4 is a bit too destructive

PR #4270, opened 2025-02-26
Adds painkiller gameplay to surgery

SomeguyManperson:
## About The Pull Request

Surgeries now have a chance (default 50%) to trigger a flinch effect, which causes extra damage and bleeding
This can be mitigated by medium levels of drunkenness, or neutralized entirely with a painkiller, including miner's salve
Being asleep in general reduces the chance of triggering a flinch effect by 20%, and strangling the patient reduces the chance by 40% (you get one of these, not both)
This does not affect mechanical limbs since they are likely built significantly better for interior access

## Why It's Good For The Game

Good????? lol

## Changelog

:cl:
add: surgeries to organic limbs that are done without anesthetic or painkillers now have a chance to cause extra damage
balance: medical cyborgs now have access to morphine in their hypospray
/:cl:


DIB-DOG:
Strangled meaning in a neck grab or actually suffocating and strangling? Because I love the idea of "We don't have painkillers, I need you to hold them down." but I don't think you'd really need to be actually suffocating them in order to do that.

SomeguyManperson:
> Strangled meaning in a neck grab or actually suffocating and strangling? Because I love the idea of "We don't have painkillers, I need you to hold them down." but I don't think you'd really need to be actually suffocating them in order to do that.

having someone held down/restrained now reduces the damage too
"strangling" meant "unconscious due to high oxygen damage" but now you can do both

code/modules/reagents/chemistry/reagents/medicine_reagents.dm
```diff
@@ -910,14 +914,14 @@
 
 /datum/reagent/medicine/dimorlin/on_mob_metabolize(mob/living/L)
 	..()
-	ADD_TRAIT(L, TRAIT_ANALGESIA, type)
+	ADD_TRAIT(L, TRAIT_PAIN_RESIST, type)
```
Erikafox: dimorlin is written to be a Strong Painkiller, which is why it uses analgesia instead of pain_resist.

SomeguyManperson: shit I thought that was one of the ones I added oops

FalloutFalcon (approved):

MarkSuckerberg (approved):
sick as hell

PR #4283, opened 2025-02-27
Significantly Reduces Price of Stockparts + Specific Boxes and Volume

generalthrax:
## About The Pull Request

Reduces the price of T2 parts and T3 parts by half. T1 parts price reduced by 400 because I'm not sure why they're even in the market.

Specific boxes for each type of part, 10 parts each (normal box still gets 2 of each).

## Why It's Good For The Game

Stockparts have been in the market for a while but they were too expensive and in too few number to use for actual projects. You could upgrade like one machine and have to spend another 3000 for the next. Maybe you want to upgrade thrusters but you don't need scanning modules? Buy a shitton of lasers. Maybe you still need that small selection of varied parts for like an exosuit or something? Buy the selection.

## Changelog

:cl:
add: Added specific parts boxes for T2 and T3 parts, 10 parts each
balance: Reduced the price of T2 parts by 750 and the price of T3 parts by 1500
/:cl:


DIB-DOG:
Forgot that you could even upgrade thruster parts. This looks cool.

FalloutFalcon:
we could prob just remove or comment out the t1 box. 100 credits is annoyingly cheap.

PR #4207, opened 2025-02-16
CMM Atlas Tweaks

DIB-DOG:
## About The Pull Request
Slightly re-maps the Atlas-class Light Armored Cruiser's armory and re-organizes it's contents to be more organized and friendly to the eyes. Additionally increases starting funds on the Atlas to 4000 as opposed to the previous 1000.

The armory change in question:
Screenshot 2025-03-06 225224

The dorm adjustments:
Screenshot 2025-03-06 225234

## Why It's Good For The Game
Compared to similiar ships of it's class, the Atlas class is simultaneously underfunded, and under-equipped. This change brings their funding at least up to par with a similar ship, but still leaves them behind materially. This lends itself well to the scrappier militia look of the CLIP Minutemen while also allowing them to fulfill their function without having to grind missions round start just to be functional. Additionally, the armory was just terrible to work out of, this should alleviate the issue, making the space more comfortable and better organized.

## Changelog

:cl:
add: Atlas Armory has been redesigned.
balance: Atlas starting funds increased to 4000
/:cl:

PR #4302, opened 2025-03-01
Adds CLIP gas masks to CLIP SSUs.

Sadhorizon:
## Why It's Good For The Game
Having access on both Atlas and Typhon (coming soon:tm:) is good I think. I know that #4207 adds it to the armory, but I think this is a better solution than putting gas masks in a uniform crate. Vela did the same but with an inherit SSU too, so points for bonus consistency. Also it's like, consistent with other factions (for example, IRMG).

## Changelog

:cl:
Minuteman SSUs start with the CLIP gas mask now.
/:cl:


meemofcourse (approved):
yea ok

PR #4318, opened 2025-03-05
Health scanner show structure damage and qol

ritorizo:
## About The Pull Request

- Health scanners can now detect structure damage on mechanical limbs.
- Health scanners now simply indicate the position of broken bone/structure instead of giving the full name of the damaged limb (looking at you "Bioshop Cyberkinetics 2.0 right arm")

## Why It's Good For The 

You can now heal IPC and not have the full backstory of people's limb clogging up your screen while operating.

## Changelog

:cl:
add: You can now see mechanical limb structure damage on scanner.
fix: Make the limb description in health scanner more concise.
:cl:

### DIFF
--- a/code/game/objects/items/devices/scanners.dm
+++ b/code/game/objects/items/devices/scanners.dm
@@ -311,12 +311,17 @@ GENE SCANNER
 		if(advanced && H.has_dna())
 			render_list += "<span class='info ml-1'>Genetic Stability: [H.dna.stability]%.</span>\n"
 
-		var/list/broken_stuff = list()		//WS Edit Begin - Adds bone breakage
+		var/list/broken_stuff = list()
+		var/list/damaged_structure = list()
 		for(var/obj/item/bodypart/B in H.bodyparts)
 			if(B.bone_status >= BONE_FLAG_BROKEN)		// Checks if bone is broken or splinted
-				broken_stuff += B.name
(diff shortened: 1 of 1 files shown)


Apogee-dev:
isn’t there, like, a robotics diagnostic tool? there’s a HUD, at least, and a pda cartridge. maybe we should just add a robotics scanner that can be used for ipcs and cybernetics (and borgs and exosuits) instead, feels kinda weird for the health scanner to also be good at scanning robots.

Imaginos16:
> isn’t there, like, a robotics diagnostic tool? there’s a HUD, at least, and a pda cartridge. maybe we should just add a robotics scanner that can be used for ipcs and cybernetics (and borgs and exosuits) instead, feels kinda weird for the health scanner to also be good at scanning robots.

There's the Diagnostic HUD and that's it.

ritorizo:
I don't remember seeing a diagnostic scanner on shiptest just the hud. And I'm not sure about making a new item.
We could probably use either :
- A modcomp program would access the software of the IPC/borg for diagnostic information or other stuff in the future like a config pannel ?
- using a t-ray scanner would be nice since they already exists everywere but really only have a niche use. They would also make sense since their use is to look trough metal floors which would be kind of the same as looking trough an IPC/borg's shell

thgvr:
One day we can have a separate tool for scanning synthetics but today is not that day

PR #4319, opened 2025-03-05
Make gun locker buildable and tweak it a bit.

ritorizo:
## About The Pull Request

- Gun lockers can now be built with 10 metal sheets.
- They can be deconstructed with a welder/angle grinder.
- When broken/destroyed the closet will drop it's content instead of deleting it.
- They can be anchored/unanchored with a wrench.
- You previously had to click it with a random item in hand in order to close the closet when it's full, this behavior was removed.
- You can now open/close gun lockers with Alt+click.

## Why It's Good For The Game

Gun lockers are nice.

## Changelog

:cl:
add: Gun lockers can be constructed with 10 metal, anchored/unanchored with a wrench and deconstructed with a welder.
fix: Gun lockers no longer delete their content when broken.
:cl:


generalthrax:
Think you could maybe increase the stored guns to 8 or however many the radial can hold? Always thought it was a little silly it could only hold 4 for such a big structure

FalloutFalcon (approved):
draging around my gun locker into battle to never have to reload a gun.

PR #4322, opened 2025-03-06
Firearm Condition and Jamming

SomeguyManperson:
## About The Pull Request

Ballistics now have a condition stat, which reduces when fired. If condition is low enough, the weapon can suffer a failure to feed malfunction, requiring it to be manually racked before it can continue to fire. Current numbers are. Well. Present. Some stuff might need changes to be actually good.
Surplus rounds now deal full damage! Instead of being ass because they lose up to 25% of their stopping power, they now suck because they double the rate at which a gun suffers condition loss.

To remedy this new form of suffering, you can buy a maintenance kit at the outpost. Currently 100 Cr for 5 uses, a maintenance kit can be used to improve a gun's condition. Several uses may be necessary if the gun is particularly battered.
Doing maintenance on a loaded firearm may result in a negligent discharge. Due to the amount of manipulation required in a cleaning cycle, fully unloading the firearm is the only way to guarantee it will not go off.

Current wear speed numbers
Shotguns: 30 shots to malfunction, 90 to critical
Pistols: 60 shots to malfunction, 180 to critical
Marksman rifles: 60 shots to malfunction, 180 to critical
Assault rifles: 100 shots to malfunction, 300 to critical
SMGs: 120 shots to malfunction, 360 to critical
HMGs: 300 to malfunction, 900 to critical. Good luck using one though.

Hunter's pride guns get +30 shots, while frontier arms are generally worse than others in their category in exchange for their firerate and/or cheapness

Gauss, pump actions, bolt-actions, lever-actions and revolvers can't jam

## Why It's Good For The Game

I can't believe John Receiver said "it's receiving time" and negligent discharged all over the place

## Changelog

:cl:
add: guns can jam now! Keep an eye on your arm's condition, if it says it can malfunction, it might fail to load a round while firing.
add: there's a new item in the security section of cargo, which can be used to repair gun condition.
balance: surplus rounds no longer deal reduced damage, instead doubling the rate at which condition is lost.
balance: some guns gain wear faster or slower than others in their category. If it's made out of scraps and duct tape, it's probably  faster, if it's hunter's pride, slower.
/:cl:


DIB-DOG:
Could I recommend changing the repair kit to be a cleaning kit? The idea of having to actually _repair_ your weapon after the amount of use we’re probably going to see in game seems a bit odd. But having to deal with fouling which could cause feed/extraction problems and stuff like that is plausible.

(ignore this, seems like the name isn’t actually “repair kit”. Only had time to read the about at the time)

firebudgy:
How many rounds does it take for a magazine-fed shotgun, SMG or an Assault Rifle to start jamming?

SomeguyManperson:
Current wear speed numbers have been added to the PR. This is currently done by making the gun collect wear faster or slower rather than moving thresholds, might see about changing that

Ratvarr:
for those who do not know, and cannot read the depths of code: Your gun can discharge, or unsafety, or rack the bolt and chamber a new round during cleaning.
Please. Please please please document this for the sake of players

FalloutFalcon (approved):
Might need some number tweaks and a wensday test merge but otherwise the code is good.

MarkSuckerberg (approved):
also absolutely sick as hell

MassiveMen:
Just a few things. Having gun cleaning cause the gun to misfire, while also having it cause the gun to un-safety is nothing but spiteful and somewhat frustrating. Also, the wear speed on shotguns is almost comically fast.

PR #4351, opened 2025-03-12
Brings .22lr back to 20

generalthrax:
## About The Pull Request

Says on the thing. Brings .22lr HP to 24 and .22lr AP to 18.

## Why It's Good For The Game

It was pretty bad before. Like worse than .38. A survival knife in the modern day does more damage on hit than .22. The AP remains bad so hopefully it can retain its angry swarm of bees / faunakiller niche it had before those ammo changes a While ago.

## Changelog

:cl:
balance: Boosts .22lr's damage very slightly. AP is still abysmal dogshit.
/:cl:


FalloutFalcon:
Honestly. I think a knife DOES more damage irl then 22lr. depending on the range of it.

Apogee-dev:
Wasn’t .22 _supposed_ to be worse than .38? Like i should hope so, it’s a far smaller and slower bullet. Ain’t no way it should be matching .38 and 9mm for damage when it’s half the size.

generalthrax:
The AP is comically low (which is fine). The damage was decreased in the Specialist Ammo PR a while ago (without mention, if I recall) and has been used very little (even less than before) ever since. Every surplus round in the game is just Straight Better than them in that they lose out on one damage and keep the same AP. Hardly even economical to use.

Apogee-dev:
> Every surplus round in the game is just Straight Better than them in that they lose out on one damage and keep the same AP. Hardly even economical to use.

Last time I crunched the numbers it was significantly more economical than .38, but if it’s not economical to use the solution should be reducing its price, not buffing the DPS of already ferocious guns like the pounder.

PR #4368, opened 2025-03-15
The atoll now has floor on that one exposed spot

Ratvarr:
## About The Pull Request
title says it all tbh

## Why It's Good For The Game
spite pr (I don't like seeing it)

## Changelog

:cl:
fix: The atoll has a floor tile instead of plating in the dining hall.
/:cl:


MemeSnorfer:
It's there completely on purpose to convey the idea of ummmm mildly used ship. I dunno. It was an intentional design choice. If you wanna make it look better just place a floor tile

thgvr:
Seems this is intended

PR #4381, opened 2025-03-16
Hermits use a spawner instead of evil subtype

FalloutFalcon:
## About The Pull Request
hermit mobs now use a proper spawner
same story with soda cans
makes gunslingers a little more common by proxy of being a weighted list instead of two probs.

## Why It's Good For The Game
You should not call all of the init of a hermit just to have a 45% to delete him and spawn a new one

## Changelog

:cl:
refactor: hermit mobs and soda cans now use a proper spawner
/:cl:


Erikafox:
soul removal

PR #4384, opened 2025-03-17
Improve Sandbase (The Cave One)

Erikafox:
## About The Pull Request

image
This ruin used to suck now it's just mid

## Why It's Good For The Game

ruin that isn't just a metagame bait bomb

## Changelog

:cl:
add: An unknown faction's stupid secret cave base has been improved. You can find it on sand worlds. You may explode.
/:cl:


Gristlebee (changes requested):
- Put some rock here so the diagonal corners of the concrete doesnt show.
image
- Same here
image
- The way the C4 is currently placed means it can disrupt the wires and rapidly cause the base to depower, which causes the turrets to brick and turn invisible. Redundant wiring would be good, or perhaps adjust the placements.
image
- Engineering SMESes would be good here.
image
- The atmos pumps should start on so the AACs can function
image
- The rooms feel a bit plain, some extra siding and decalling could look nice. Rust, or damaged tiles could be a nice touch as well to signal age.
- I have concerns about the bomb, especially with it having a large core. You can still unanchor it via the wires without risk of detonation. I'd replace it with a miniature core. Or if you can make it so you actually have to disarm it first before taking it, that could be interesting as well.
- The laser turrets felt rather sluggish to lock, and non-threatening. You may want to set their reaction_time var to something faster.

Gristlebee (dismissed):
Review after testing on live:
---
- The airfloor tank in the atmos room starts empty.

- If you're having the bomb be a mission, it shouldn't be sealed away, or it should have a hint on where it is so it can be reasonably found. I'd also still like the payload to be reduced.

- The dead goliaths need their density set to zero so things can pass through.

TheUSSWatercooler:
Are the wire tangles in the airlocks supposed to be indestructible? Went through the ruin recently and we tried breaking them, had to just cut through the wall.

Erikafox:
no
that's an issue with the hazard imo

PR #4400, opened 2025-03-18
Quirk Rebalance +  Bonus Language Quirks

thgvr:
## About The Pull Request
This also sorts quirks into individual files. also removes the infinite pizza box. sorry I eated it all

junkie is still here along with smoker but is not in the screenshot. these quirks are awful code though

Also adds four new language quirks, all exclusive with each other. This means your character can freely pick one extra language.

Removes a lot of quirks that just sucked, didn't work, unused, etc. The final list of quirks in the game is now this: 
image

## Why It's Good For The Game
its not

## Changelog

:cl:
add: New language quirks for the four main languages
del: Removed and rebalanced a lot of quirks
/:cl:


thgvr:
Also I have no idea how to solve this or if there's code inplace thats broken right now, but I don't think this will actually clear out removed quirks from characters.

shiptest.dme
```diff
@@ -409,7 +409,6 @@
 #include "code\controllers\subsystem\vis_overlays.dm"
 #include "code\controllers\subsystem\vote.dm"
 #include "code\controllers\subsystem\weather.dm"
-#include "code\controllers\subsystem\processing\ai_basic_avoidance.dm"
```
FalloutFalcon: ?

FalloutFalcon:
> Also I have no idea how to solve this or if there's code inplace thats broken right now, but I don't think this will actually clear out removed quirks from characters.

tg has like a preference migration system, we dont i think.

shiptest.dme
thgvr: fucked up merging master, missed that one oops

thgvr:
dunno why it would compile without that but whatever, its back

thgvr:
mergies 🥺

PR #4403, opened 2025-03-18
cirrus/nimbus changes

Erikafox:
## About The Pull Request

mildly changes the cirrus and nimbus's starting food
they get more meat and less sake

also the cirrus can breed trout for eating

## Changelog

:cl:
balance: cirrus and nimbus got a little more variety in their food. Not much mind you, but some.
/:cl:


Gristlebee:
Seems fine. But I would prefer if the sake bottle was a proper subtype if you're putting it on both ships.

PR #4411, opened 2025-03-19
Removes access restriction for toggling maintenance mode/switching parts on exosuits

SomeguyManperson:
## About The Pull Request

it's gone

## Why It's Good For The Game

it's bad (most people won't have the required access)
only thing it really stops is combat toggling maint mode to disable the suit in which case you should've simply had maintenance mode forbidden??

## Changelog

:cl:
del: exosuits no longer require access to swap their parts
/:cl:

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -70,7 +70,7 @@
 	var/internal_damage = 0 //contains bitflags
 
 	var/list/operation_req_access = list()//required access level for mecha operation
-	var/list/internals_req_access = list(ACCESS_MECH_ENGINE, ACCESS_MECH_SCIENCE)//REQUIRED ACCESS LEVEL TO OPEN CELL COMPARTMENT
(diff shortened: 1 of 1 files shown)


thgvr (approved):

Apogee-dev:
Shouldve just switched it to engineering access. Damn near every ship’s got one after all. That or add robotics access to engineer and pilot jobs. Im not quite convinced random deckhand number 42 should necessarily have full technician access to an exosuit.

SomeguyManperson:
> Shouldve just switched it to engineering access. Damn near every ship’s got one after all. That or add robotics access to engineer and pilot jobs. Im not quite convinced random deckhand number 42 should necessarily have full technician access to an exosuit.

access restrictions for suits can be added via maintenance mode and maintenance mode can only be activated if someone allows it through the status panel
If a ship has designated mechanic/pilot roles it can have access added either ic or via map edits/subtypes but most of the time it's either a purchase or scavenged item and I don't think access locks would be built-in by default

FalloutFalcon (approved):
I agree with theos' last comment

PR #4413, opened 2025-03-20
Cable healing on mechanical limbs no longer does structure damage

SomeguyManperson:
## About The Pull Request

IPCs take burn damage while in crit (and a lot of it) to the point cables are the only reliable way to get them out, meaning options are either
1: hope they have enough structure left to heal out of crit and not have the healing stall then die anyway
2: wait for the IPC to die, then heal them either with repair machinery or stacks & structure repair that won't be immediately undone by damage from crit

Structure is still dealt by healing via welding tool (most damage you take will generally be brute damage anyways)

## Why It's Good For The Game

"just let them die lol" generally shouldnt be the strat for medical stuff I think

## Changelog

:cl:
balance: healing mechanical limbs with cables no longer deals structure damage
/:cl:

### DIFF
--- a/code/modules/power/cable.dm
+++ b/code/modules/power/cable.dm
@@ -544,7 +544,7 @@ GLOBAL_LIST_INIT(cable_coil_recipes, list(new/datum/stack_recipe("cable restrain
 			user.visible_message("<span class='notice'>[user] starts to fix some of the wires in [H]'s [parse_zone(affecting.body_zone)].</span>", "<span class='notice'>You start fixing some of the wires in [H == user ? "your" : "[H]'s"] [parse_zone(affecting.body_zone)].</span>")
 			if(!do_after(user, 0.5 SECONDS, H))
 				return
-		if(item_heal_robotic(H, user, 0, 15, integrity_loss = 5))
+		if(item_heal_robotic(H, user, 0, 15))
 			use(1)
 		return
 	else


Martinpachu:
isnt the burn damage so insane no one could reasonably outheal it anyway

Burning02:
> isnt the burn damage so insane no one could reasonably outheal it anyway

It _can_ be out-healed, but you need atleast 2 people healing you at the same time, otherwise its just prolonging how long before you die

Martinpachu:
> > isnt the burn damage so insane no one could reasonably outheal it anyway
> 
> It _can_ be out-healed, but you need atleast 2 people healing you at the same time, otherwise its just prolonging how long before you die

yeaahhh.. i think this is a reasonable change? the only issue would be in pvp where youre fighting an ipc with laser guns and they never have to retreat but i dont think it matters too much

Burning02:
> > > isnt the burn damage so insane no one could reasonably outheal it anyway
> > 
> > 
> > It _can_ be out-healed, but you need atleast 2 people healing you at the same time, otherwise its just prolonging how long before you die
> 
> yeaahhh.. i think this is a reasonable change? the only issue would be in pvp where youre fighting an ipc with laser guns and they never have to retreat but i dont think it matters too much

In my own personal experience, I tend to chew through a stack of cables when having to go against lasers. Not sure about others, but that's what I went through

FalloutFalcon (approved):

PR #4414, opened 2025-03-20
IPCs with the MMI brain setting now get an MMI brain

SomeguyManperson:
## About The Pull Request

moves where the mutant brain is checked to before it's actually generated so it does something

## Why It's Good For The Game

pref thing that didnt work now does work

## Changelog

:cl:
fix: IPCs with an MMI brain now get an MMI brain instead of it giving them a positronic anyways. You are NOT a government drone.
/:cl:


thgvr:
wonder when this broke

PR #4418, opened 2025-03-21
Removes armor from boots, gloves, and jumpsuits (and a hat)

Apogee-dev:
## About The Pull Request

Strips armor (except acid and fire) off of all non-adminspawn gloves, boots, and jumpsuits that had any. And the PGF hat.

## Why It's Good For The Game

Gloves, jumpsuits, and boots combine their armor with anything worn over the top that has the same coverage (such as a hardsuit). This probably isn't good, especially since more powerful armor is more common here than on a station server and we'll likely be adding more full-body suits in the future.

## Changelog

:cl:
balance: Removed armor from gloves, boots, and jumpsuits. Armor is only a feature of actual armor items and space suits.
/:cl:


FalloutFalcon:
Idk, i would prefer some very light amour be kept on things like combat boots and gloves, miner gloves, but most of the uniforms and all the chameleon stuff i 100% agree with. The majority of players don't get access to the type of full body amour your describing as most ships are stuck with pretty much bog standard tac vests.

generalthrax:
I think it is also worth mentioning that currently boot and glove armour does not work and has not worked for years (it would be a simple change yet probably a poor one). Most of this isn't playerfacing and uniform armour for stacking is Pretty Bad.

thgvr:
yep

PR #4421, opened 2025-03-21
IPC balance change

thgvr:
## About The Pull Request
Removes IPCs dying in crit.

Removes IPCs auto-reviving on reaching low damage. This will probably mean if you get stuck in crit somewhere, you'll never die. But you can always manually succumb so this is fine. It also causes issues with some other stuff, and nullifies the surgery that exists for it.
Wiki purposes, the surgery is:
1. Unscrew Shell (Screwdriver 100%, Scalpel 75%, Knife Subtype 50%, any item 10%)
2. Open Hatch (Hand)
3. Prepare Electronics (Multitool 100%)
4. Revive (Inducer 100%, Shockpaddles 80%, Baton subtype 50%, Energy gun subtype 30%)
5. Close Hatch (Hand)
6. Screw Shell (Same as step 1)

Slight tweaks to the surgery in the diffs.

## Changelog

:cl:
balance: IPCs no longer die rapidly in critical condition
balance: IPCs no longer auto-revive
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/IPC.dm
+++ b/code/modules/mob/living/carbon/human/species_types/IPC.dm
@@ -4,7 +4,7 @@
 	sexes = FALSE
 	species_age_min = 0
 	species_age_max = 300
-	species_traits = list(NOTRANSSTING,NOEYESPRITES,NO_DNA_COPY,TRAIT_EASYDISMEMBER,NOZOMBIE,MUTCOLORS,REVIVESBYHEALING,NOHUSK,NOMOUTH,NO_BONES) //all of these + whatever we inherit from the real species
+	species_traits = list(NOTRANSSTING,NOEYESPRITES,NO_DNA_COPY,TRAIT_EASYDISMEMBER,NOZOMBIE,MUTCOLORS,NOHUSK,NOMOUTH,NO_BONES) //all of these + whatever we inherit from the real species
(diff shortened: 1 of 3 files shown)


DIB-DOG:
The day has finally come, rejoice.

Do you think it would be valid to make wire cutters work at 50% on the prepare electronics step? Be effectively free-handing the Hotwire/rewire without testing it.

thgvr:
> The day has finally come, rejoice.
> 
> Do you think it would be valid to make wire cutters work at 50% on the prepare electronics step? Be effectively free-handing the Hotwire/rewire without testing it.

Yeah sure. I just don't think there is ever gonna be a scenario right now where you don't have a multitool, but have wirecutters.

Martinpachu:
i thought inducers werent being used anymore since theyre so rare

Apogee-dev:
I thought inducers werent being used anymore because theyre jank as hell and enable troll physics. couldn’t we just, like, make paddles work at 100% and nuke inducers like we should have years ago?

Slim-Pickems:
I do know you can use hemostats to prep electronics on other surgeries. It would make sense to allow it here as well.

generalthrax:
I'm pretty sure inducers just move charge from one cell to charge from another cell. Any "troll physics" that happen from them you can do normally without an inducer. If you wanna talk inducer stuff you can probably mention it in the cargo pr.

Ziiro:
I made REVIVESBYHEALING specifically for IPCs so if you're ditching it you could probably purge it from the code base and all its interactions

Gristlebee:
Revival surgery causes brain damage even if successful. That should be removed I think if this is going to be the intended way to revive IPCs.

thgvr:
yeah thats fucked up

PR #4424, opened 2025-03-21
Additional Engineering Tools on the Cargo Market

firebudgy:
## About The Pull Request

Adds Inducers (for the upcoming #4421 PR) to cargo

Also removes a reference to greytiding.

## Why It's Good For The Game

We're going to need inducers to be available if IPCs require rebooting, so this makes them purchasable in the event of an IPC's deactivation.

## Changelog

:cl:
add: Additional tool in cargo: Inducers.
/:cl:

### DIFF
--- a/code/modules/cargo/packs/tools.dm
+++ b/code/modules/cargo/packs/tools.dm
@@ -151,11 +151,10 @@
 
 /datum/supply_pack/tools/jaws_of_life
 	name = "Jaws of Life"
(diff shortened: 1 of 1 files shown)


thgvr:
I actually don't exactly like the jaws and drill in the state they're in - jaws can be an interesting case for prying open airlocks however it should probably be its only function. Drill needs cell component added to it before I'd want to make it readily available. Inducer is fine though

firebudgy:
Gotcha, I'll remove both from this PR

thgvr (approved):

PR #4426, opened 2025-03-21
short scope gives minimum recoil

SomeguyManperson:
## About The Pull Request

adds 0.1 minimum aimed recoil to the short scope (same as the long scope unfortunately) since they are generally applicable to assault weapons

## Why It's Good For The Game

THE 0 recoil skm in question:

## Changelog

:cl:
balance: short scopes now have minimum recoil, similar to long scopes
/:cl:

### DIFF
--- a/code/game/objects/items/attachments/short_scope.dm
+++ b/code/game/objects/items/attachments/short_scope.dm
@@ -9,6 +9,7 @@
 	size_mod = 0
 	var/zoom_mod = 6
 	var/zoom_out_mod = 2
+	var/min_recoil_mod = 0.1
 	var/aim_slowdown_mod = 0.2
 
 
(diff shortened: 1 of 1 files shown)


FalloutFalcon:
why does this even exist in the first place? @Erikafox ?

Erikafox:
why does what exist

FalloutFalcon:
this var/min_recoil_aimed, every ballistic gun uses 0.1 and lasers should prob not GAIN recoil if they have a scope attached.

Erikafox:
because recoil is generally ass with zooming, so zooming disables it unless you're using a long scope (to discourage using automatics at a long distance)

FalloutFalcon:
huh. i see what you mean. went and tested it, it is quite bad.

FalloutFalcon:
> because recoil is generally ass with zooming, so zooming disables it unless you're using a long scope (to discourage using automatics at a long distance)

i definitly agree. Tho its a bit of a hacky solution. we really just need to fix the zooming amplifying shake like it does.

Issue #4427, opened 2025-03-21
[BUG]Sledgehammers don't have wielded sprites anymore


Ratvarr:
## Round ID:
N/A

## Testmerges:
N/A

## Reproduction:
Caused by #3964 
spawn gorlex hammer/basetype sledgehammer
wield
your hammer has disappeared from your body sprite.

PR #4429, opened 2025-03-21
Fixes Melee Refactor bugs and reverses undocumented knife change

### LINKED ISSUE #4428: [BUG]SolCon Energy Halberds can't go on your back anymore
## Round ID:
N/A

## Testmerges:
N/A

## Reproduction:
caused by #3964 
you cant fit it on your back anymore because its a subtype of the Desword which has no back slot ability

generalthrax:
## About The Pull Request

- Fixes sledgehammer sprites not working
- Fixes sledgehammer window demolition not working
- Fixes halberds not fitting on back
- Makes the blunt inhand files real and moves sledgehammer icons into those
- Reverses the undocumented knife force change

Fixes https://github.com/shiptest-ss13/Shiptest/issues/4428 and https://github.com/shiptest-ss13/Shiptest/issues/4427

## Why It's Good For The Game

Fixes good. I realize knife force was probably intended because of the new attack cooldown but in reality you aren't going to get that many hits in anyways at that close range since you're going to be moving backwards to avoid the stronger simplemob melee. Better to keep the old force than the new.

There are probably other issues but I don't have time to look through the other 100 files.

## Changelog

:cl:
fix: Sledgehammer sprites, sledgehammer window demolition, halberd backslot
balance: Knives now do their normal force again.
/:cl:


FalloutFalcon:
I gave knives faster swing speeds, that's documented as part of the pr, i guess i didnt mention the fact that the damage also changed but it does kinda need to change for that pr to not be a pretty major buff to them.

code/game/objects/items/melee/blunt.dm
```diff
@@ -57,6 +60,15 @@
...
 
+/obj/item/melee/sledgehammer/gorlex/afterattack(atom/A, mob/user, proximity)
+	. = ..()
+	if(!proximity)
+		return
+	if(HAS_TRAIT(src, TRAIT_WIELDED)) //destroys windows and grilles in one hit
+		if(istype(A, /obj/structure/window) || istype(A, /obj/structure/grille))
+			var/obj/structure/W = A
+			W.obj_destruction("axe")
+
```
FalloutFalcon, line 71: also its fine for this pr but with demolition mod, i dont think either of the melee need this behavoir anymore, its terribly snowflaked and ignores the type of window.

generalthrax:
> in reality you aren't going to get that many hits in anyways at that close range since you're going to be moving backwards to avoid the stronger simplemob melee. Better to keep the old force than the new.

Aforementioned reasoning for bringing it. back to the original. With fighting most mobs you are going to be (naturally) looking to avoid getting hit, which the quicker cooldown doesn't really help out much.

It's about the same time-to-kill on ordinary mining fauna (Goliaths 10 hits and Legions 5 hits compared to original Goliaths 7 hits Legion 4 hits). 

Knives also have 0 armour penetration.

FalloutFalcon:
I think we should prob settle for some sorta middle ground because while your probally not hitting EVERY cooldown you can still presume people are going to be able to hit more often. the coolddown is 3/4 the norm, i balanced most of the knives to have 3/4 the force, i think it makes sense to buff that up to like 7/8 the force or similar tho.

FalloutFalcon (approved):
I can put in another pr tho if you dont want it here.

generalthrax:
I think it's safe to keep the normal force for now. We can put it through some livetesting and see if it's like Comically Broken or something. If its not then we'll see

PR #4430, opened 2025-03-21
small grammar fix

Erikafox:
adds a cr to the payment announcer for missions


FalloutFalcon:
is our standard "100cr" or "100 cr"

PR #4431, opened 2025-03-21
update airbase turrets

Erikafox:
puts heavy ramzi turrets on the syndicate airbase because the turrets they were previously subtyped off don't exist any more.

## Changelog

:cl:
fix: the syndicate coalition has retroactively installed the correct turrets on their old abandoned airbase.
/:cl:


Apogee-dev:
A brief description of the changes in the pr body would be useful, the cl is fairly vague. stuff like which turrets you put in etc

PR #4449, opened 2025-03-22
Adding Smoke Grenades and Re-pricing Stingers in Cargo

DIB-DOG:
## About The Pull Request
This adds smoke grenades to be cargo purchaseable in a pack of 7, much like the flashbangs and teargas grenades, pricing the box at 500 credits.

This also adjusts the price of the stingbang grenades box from 1500 to 750 creating a more reasonable price per unit. 

## Why It's Good For The Game
Smoke grenades already had a box but were unable to be purchased in cargo. This simply updates their box description to something more flavorful and makes the box accessible in cargo. The 500 credit price tag puts the smoke grenades at about 71 credits per grenade, which I believe is fair considering that smoke grenades can commonly spawn as random loot on sand planets. They also lack any of the agressive effects, such as those held by teargas grenades or flashbangs.

The price of stingbangs was lowered to 750 credits for the box of 5 as to make them viable when compared with frag grenades. With the new price, a single stingbang's cost is 150 credits, instead of the previous 300 credits per grenade. When compared to the 250 credit fragmentation grenade, I think that this is a fair price change which may encourage the use of stingers, or at least make them viable if nothing else.

## Changelog

:cl:
add: Added smoke grenades to cargo
balance: Rebalanced stinger grenade price
/:cl:


code/modules/cargo/packs/sec_supply.dm
```diff
@@ -189,7 +196,7 @@
 /datum/supply_pack/sec_supply/stingpack
 	name = "Stingbang Grenade Pack"
 	desc = "Contains five \"stingbang\" grenades, perfect for stopping riots and playing morally unthinkable pranks."
-	cost = 1500
+	cost = 750
```
FalloutFalcon, line 199: Hrm. The idea behind most stun based pricing is that stuns are evil and should be avoided. This, seems fine?

DIB-DOG: I don’t think the stingbangs apply any hard stuns beyond a knockdown and some stamina damage. In my tests with them I was rarely struck with any more than 2 fragments of ballistic gel. They really feel like less lethal frags in practice. Ofc, if it’s a precedent for stuns/nonlethals to be more expensive, I could dial it back. Set the price to 1000?

PR #4454, opened 2025-03-22
add ramzi faction

Erikafox:
## About The Pull Request

adds a ramzi faction and applies it to ramzi turrets and mobs

## Changelog

:cl:
fix: ramzi installations should infight less
/:cl:

### DIFF
--- a/code/modules/faction/faction_datum.dm
+++ b/code/modules/faction/faction_datum.dm
@@ -128,6 +128,6 @@
 	name = FACTION_RAMZI
 	short_name = "RAM"
 	parent_faction =  /datum/faction/ramzi
-	prefixes = PREFIX_INDEPENDENT
+	prefixes = PREFIX_RAMZI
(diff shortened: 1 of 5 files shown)


FalloutFalcon (approved):
We should really convert turrets to work off datums instead of just the names at some point

code/modules/faction/faction_datum.dm
```diff
@@ -123,3 +123,11 @@
 	prefixes = PREFIX_INDEPENDENT
 	color = "#A0A0A0"
 	order = FACTION_SORT_INDEPENDENT
+
+/datum/faction/ramzi
+	name = FACTION_RAMZI
+	short_name = "RAM"
```
FalloutFalcon, line 129: I would prob consider the short_name to just be Ramzi tho. I dont think i added the "IND" but its meant to be for like, the shortened version that players use alot.

PR #4483, opened 2025-03-28
fix the booze mission

Erikafox:
:cl:
fix: i no longer have to answer adminhelps to fix the booze mission
/:cl:


FalloutFalcon:
sorry forgot to fix this.

PR #4488, opened 2025-03-29
Rhodes-class Colonial Ship (Colony Round Ship)

Ical92:
### This ship is specifically designed for event use and is only spawnable by admins!

Adds a new Miski-flavored ship for colony rounds, replacing the outdated Bubble-class. The Rhodes-class is not enabled for player spawning, as its made for spawning by admins during colony rounds. It also has most things that should be needed for starting out a colony, while also having space for any additions per the event runner. The ship has slots for one leader (event runner or one lucky individual) and 10 slots initially for colonists, though this can be raised/lowered by the event runner. This includes; an airlock, 2 EVA suits, 2 Candors, and basic mining gear.

SDMM Screenshot
image

## Why It's Good For The Game
Gives event admins a more up-to-date colony ship for their colony events, and less work for the event runners to make a starting point for the colony if they had previously not used the Bubble-class.

## Changelog

:cl:
add: Rhodes-class Colonial Ship
del: Bubble-class Colonial Ship
/:cl:


thearbiber:
Apologies, but I'm not sure we need something like this in modern day. Colony ships simply aren't a niche we are pursuing anymore, preferring mapped starting points. Thank you for your contribution, but I'll be closing this for now.

PR #4507, opened 2025-04-02
changes the frontie distillery turret to a spitter turret

generalthrax:
## Why It's Good For The Game

also makes the turret not get killed by the guy at the front

## Changelog

:cl:
fix: Frontie distillery turret is now a spitter and is no longer destroyed by the fronties
/:cl:


thearbiber:
I forgot the turret even existed

PR #4533, opened 2025-04-05
Poverty 2

Erikafox:
## About The Pull Request
Reduces the amount of money that should be floating around by beating the economy to death and then choking its corpse before firing a shot into its  body.

## Why It's Good For The Game

I was getting feedback that things have been going too good for too long
## Changelog

:cl:
balance: pad bounties can now have a set minimum value
balance: materials now deprecate value faster when sold
balance: reduced payment from all missions (ruin and outpost)
balance: storm missions now pay a different dividend per storm.
balance: slightly reduced value of ice to credit transactions (2:5 to 3:10)
/:cl:


thearbiber:
you should nerf asteroid mission prices also

Erikafox:
you're so right

Zevotech:
Ice mining has had it too good. Recently got about 15k credits from 1500 ice chunks (as a single person mining ice) at 1 merit to 4 credits. Might want to cap it at 1:2.5 or something.

code/datums/ruins/icemoon.dm
```diff
@@ -34,7 +34,7 @@
 /datum/mission/ruin/fallen_montagne
 	name = "dark signal investigation"
 	desc = "A hunting lodge located on an ice-world in system has recently ceased communication. We suspect they may have been assulted by pirates. If this is the case, and they have fallen, bring the Montague's corpse, so they may be buried properly."
-	value = 3000
+	value = 2000
```
FalloutFalcon: This mission is getting harder cause that ruin got a big buff in difficulty, 3k seems fair.

generalthrax:
image

The changes are good but I think these three contracts (dust, carp, asteroid) are pretty much worthless over the others. I'd probably suggest 800 base for carp / dust and 150 per storm, and probably 1800 average for asteroid with maybe 200 per.

joelTheLostOne:
this is sad but needed, I was able to run a ship while forfeiting 50% of all income as payment, which I presume wasn't the intended experience.

Problem is now I won't be able to pay 50% of all income as payment. Woe upon the Athena.

thgvr:
poverty

Madtankdriver:
If you are cutting down on the payment for cargo missions (dust, carp, asteroid)
Maybe increase the number of missions you can have at once? Maybe to 3? 
So, it doesn't feel like the space missions are completely pointless.

generalthrax:
After a round with this testmerged I'm pretty content. Probably want to give it a few more goes, probably see something like mineral selling in particular

thgvr:
Also if you aren't interested in having this drafted, please put actual information/changelog in the pr

Erikafox:
you right

PR #4538, opened 2025-04-05
Buffs the debug outfit

Gristlebee:
## About The Pull Request

The debug hardsuit no longer has slowdown, the gloves protect from shocks again and the debug bag of holding can be accessed directly from the back.

## Why It's Good For The Game

QOL. Makes the debug outfit better for well. Debugging. 

## Changelog

:cl:
add: Debug outfit is better.
/:cl:


thgvr:
funny I was actually just thinking of this. We also have the debug modsuit

PR #4539, opened 2025-04-06
Revamps ship placement test

MarkSuckerberg:
## About The Pull Request

Provides more information on the ship placement test, as well as only docking ships to one outpost instead of every single one. We don't gain much with that extra time and at the moment it's a little unbearable.

## Why It's Good For The Game

Allows us to break things faster

## Changelog

:cl:
/:cl:


FalloutFalcon:
I might make a separate test that tests ALL outposts but only with a few ships that fit into each size catagory.

PR #4550, opened 2025-04-07
only allow auto ship locking if there are MORE then 2 ships.

FalloutFalcon:
## About The Pull Request
My check was backwards! Oops!

## Why It's Good For The Game
that was a mistake

## Changelog

:cl:
fix: only allow auto ship locking if there are MORE (not less) then 2 ships.
/:cl:


meemofcourse (approved):
i'm not a coder what does this pr do

PR #4551, opened 2025-04-07
Fixes the exosuit salvage saw, for real this time

SapphicOverload:
## About The Pull Request

Exosuit salvage saw's timed action is now correctly canceled when it's supposed to, and no longer needs to lock the exosuit in place. Also fixes an issue where certain things couldn't be deconstructed with it, such as crates and shutters, and gives it the same sound as the angle grinder because the complete silence as you cut things apart with a giant saw is a bit off.

## Why It's Good For The Game

Locking the whole exosuit in place was a jank solution created out of necessity.

## Changelog

:cl:
add: exosuit salvage saw now has the angle grinder sound effect when used, instead of nothing
balance: salvage saw no longer locks your exosuit in place
fix: fixed salvage saw's action not being canceled properly when moving/turning/swapping 
equipment/etc
fix: fixed salvage saw not being able to deconstruct certain things it's supposed to
/:cl:


FalloutFalcon:
My eyes are glazing over the code to much rn will give a proper review later but will def need a test merge to ensure it wont break do_afters

PR #4556, opened 2025-04-08
[TM ONLY] Disables bone breaking

Rug-Dealer:
## About The Pull Request
Comments out the bone_break() proc and the related calls to it.

## Why It's Good For The Game
Bone breaks are not an interesting or fun mechanic.

## Changelog

:cl:
balance: You can no longer break your bones from any source.
/:cl:


Erikafox:
if you're going to remove it do it right please.
We always have git history.

thgvr:
comment out the whole proc instead and any use of `bone_break()`

thgvr:
Also update from master

Rug-Dealer:
the master update was really scary
image

FlintBarrel:
As annoying as 'random' broken bones or ribs are, will this not take a decent amount of interaction from medships? They'll really only have deaths, distress calls, and implants (depending on the ship) after this.

thgvr:
Med ships aren't Ships that Just Do Med. they also do the normal gameplay loop and going to other players calls is icing on the cake

Madtankdriver:
Ok, hold on let's think about this before removing broken bones. First thing med ships and your shuttle's doctor need to do something medical wise and removing broken bones would be one less thing for them to do. You'd also be removing chems related to fixing broken bones, and would it really make sense for you to be bashed in the chest by someone with a GOD DAMN Sledgehammer or shot in your bulletproof vest and not expect any broken bones? Like really? How realistic is that? 
Yeah, get shot 8 times in the chest and immediately suture yourself up as if nothing happened.

There is another thing about playing as an organic whereas compared to an ipc. One of the benefits of playing as an ipc is that you cannot get broken bones but have to do that whole structural repair thing with rods instead. Removing broken bones could conflict with people's choices for playing as a race.

Removing broken bones sounds like a good idea and yes, it is a pain, but would it really make sense?

thgvr:
[TM ONLY]

Madtankdriver:
OH

Rug-Dealer:
bamboozled again
i lowkey agree with you though

thgvr:
Mostly what I want to see here is how much this actually will affect gameplay. I don't see a lot of positive or interesting interactions, and there's some other stuff with more recent chest zone attack bias from rand_zone stuff that pushes it being annoying a little over the edge. An alternate solution as discussed a bit on discord was to either allow splinting of chests or just removing rib breaks.

PR #4561, opened 2025-04-08
Shotgun Ammo Adjustments

Gristlebee:
## About The Pull Request

Buckshot costs 350 from 520, a bit more expensive than a rifle round.
Rubbershot now costs 350 from 520
Shotgun blanks cost 110 from 220.
Shotgun slugs are now armor neutral.
Improvised shotgun ammo can now be printed in the autolathe directly for 5000 cm^3 of metal per casing. (2 sheets and a half)

## Why It's Good For The Game

Buckshot is currently overpriced and does not feel great to use for the cost. It's based around the ideal situation of pointblank against an unarmored target. Against fauna, that's fine. Against opponents with guns, that's becomes much less feasible as they tend to be armored (simple mob or player) and the increasing lethality of other rounds like rifles makes moving into close range much more dangerous. I feel it's much more reasonable for them to be priced on par with rifle rounds like 8x50 and other specialist ammo.

Reduces the cost of blanks as well, because I don't think it's reasonable for them to cost almost as much as actual ammo.

Makes slugs AP neutral to reflect the fact the desc says they're supposed to have better penetration.

Makes improvised shotgun ammo autolathe printable for QoL. You need an autolathe anyways to make cable coil for improvised buckshot, so it makes things a bit more efficient.
## Changelog

:cl:
add: Improvised buckshot can be autolathe printed
balance: buckshot costs 350.
balance: Rubbershot costs 350
balance:  Blanks costs 110.
balance: Shotgun slugs have 0 AP
/:cl:

### DIFF
--- a/code/modules/cargo/packs/ammo.dm
+++ b/code/modules/cargo/packs/ammo.dm
@@ -218,13 +218,13 @@
 /datum/supply_pack/ammo/blank_shells
 	name = "Blank Shell Crate"
 	desc = "Contains a box of blank shells."
-	cost = 220
+	cost = 110
 	contains = list(/obj/item/storage/box/ammo/a12g_blank)
 
 /datum/supply_pack/ammo/rubbershot
 	name = "Rubbershot Crate"
 	desc = "Contains a box of 32 12 gauge rubbershot shells. Perfect for crowd control and training."
-	cost = 520
+	cost = 300
 	contains = list(/obj/item/storage/box/ammo/a12g_rubbershot)
 
 /datum/supply_pack/ammo/techshells
--- a/code/modules/cargo/packs/ammo.dm
+++ b/code/modules/cargo/packs/ammo.dm
@@ -206,7 +206,7 @@
 /datum/supply_pack/ammo/buckshot
 	name = "Buckshot Crate"
 	desc = "Contains a box of 32 buckshot shells for use in lethal persuasion."
-	cost = 520 //6.4 ammo efficiency at 104 damage. Yes we are counting point blank.
+	cost = 350
 	contains = list(/obj/item/storage/box/ammo/a12g_buckshot)
 
 /datum/supply_pack/ammo/slugs
--- a/code/modules/cargo/packs/ammo.dm
+++ b/code/modules/cargo/packs/ammo.dm
@@ -224,7 +224,7 @@
 /datum/supply_pack/ammo/rubbershot
 	name = "Rubbershot Crate"
 	desc = "Contains a box of 32 12 gauge rubbershot shells. Perfect for crowd control and training."
-	cost = 300
+	cost = 350
 	contains = list(/obj/item/storage/box/ammo/a12g_rubbershot)
 
 /datum/supply_pack/ammo/techshells
--- a/code/modules/projectiles/ammunition/ballistic/shotgun.dm
+++ b/code/modules/projectiles/ammunition/ballistic/shotgun.dm
@@ -60,8 +60,8 @@
 	custom_materials = list(/datum/material/iron=250)
 
 /obj/item/ammo_casing/shotgun/improvised
-	name = "improvised shell"
-	desc = "An extremely weak shotgun shell with multiple small pellets made out of metal shards."
+	name = "surplus buckshot shell"
+	desc = "A makeshift shotgun shell with multiple small pellets made out of metal shards."
 	icon_state = "improvised"
 	projectile_type = /obj/projectile/bullet/pellet/improvised
 	custom_materials = list(/datum/material/iron=250)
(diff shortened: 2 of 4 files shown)


Apogee-dev:
No on the buckshot AP buff, price change is fine. Shotguns are defined by the specialist cartridges they use, and none of those should be too useful as a generalist choice; if buckshot feels bad to use against humanoid mobs and especially players, then _good,_ that's exactly as it should be; it's primarily for hunting fauna. It's very effective at that, especially after goliaths got their armor stripped. If you're fighting fronties, ramzi, or players, you should be using slugs or a different weapon entirely. Buffing buckshot just gives players even less reason to use the other ammo types even accounting for a buff to slug AP by once again making it the most effective pick in most circumstances. The other cartridge buffs are fine.

Gristlebee:
> that's exactly as it should be; it's primarily for hunting fauna

Fair enough. I think if buckshot is for primarily for hunting fauna and keeping the -30 AP the price could stand to be a little bit cheaper then. Does 300 sound reasonable?

Apogee-dev:
> reduce buckshot price further

i think 350 is fine; it still does so much damage that it'll kill almost all fauna in only one or two shots, after all. that's already substantially faster than all of the rifles, which have the same price point, especially when you consider shotguns tend to have higher capacity and fire faster than most DMRs. heck, it's already the most cost-efficient ammo choice in the entire game if that's what you're using it for, and by a wide margin. if you're using buckshot for e.g. running drill missions, you're going to be spending even less than you would with .22 or .38, which also exist more or less for that purpose.

Gristlebee:
AP's dropped as requested, but I still do think buckshot could still use a buff in some way. The anti-fauna niche doesn't really feel great gameplay wise, and people have started taking out pistols against armored mobs due to how impotent buckshot feels. 

It's an another fun angle to combat with a strong risk/reward going in close with buckshot, especially with current shiptest combat really starting to lean toward long range fighting with the rifle buffs, and having slugs be the intended choice against human mobs reduces the uniqueness and renders its close to another rifle-like.

Erikafox:
Buckshot feels depressing outside of literal pointblank against unarmored things and honestly slug feels anemic for it's intended goal. 
Imo the original proposed stats were fine.

Erikafox:
When I say depressing I mean they got shotguns on mental health watch.
It was fun to literally sprint up to targets and on them but the moment you have a particularly stiff jacket on (or god forbid, you stand 3-4 paces away) it stops being effective.
Slugs having negative AP kneecaps their long ranged usage goal where literally any other AP option can outperform them. If you told me slugs were actually broken I would believe you

generalthrax:
With incoming movement speed changes I really don't think buckshot being upped to -10 would really put anything into peril, really, besides make buckshot decent as a close range option. After captaining at least once for most uptimes for a good long while, I've had a majority of crews prefer other close-range options like the Spitter, oftentimes shoving their shotguns back into their crates for less dangerous and more damaging options.

It's dangerous to get up to that distance with shotguns and will get even more dangerous in short time. We've reached nearly tg riot shotgun levels of neutered buckshot that the prevailing opinion is that a Shredder-armed Frontiersmen could have a lower time-to-kill than the Pounder.

Apogee-dev:
i mean, yeah, people _should_ be using pistols and SMGs against armored mobs instead of buckshot, actually. That's reasonable and intended. I said neutral AP on slugs is fine, and honestly, they could use more range, so I agree slugs should be better. But I firmly believe buckshot should not be a viable option for combat against equipped humanoid targets. Shotguns _should,_ in fact, be worse than SMGs for CQB when humanoids are involved, and I don't believe any setup in which medium armor like sec vests (35 armor, yes that's medium when we consider 60 armor to be heavy) fails to increase the TTK of buckshot when it adds 1-2 shots for almost everything else will work out. Shotguns should not invalidate armor through raw damage even at point blank, and that inevitably means theyre going to feel bad against humans. Such are the breaks. Buckshot has a niche and it's running drills for pennies on the dollar.

generalthrax:
Maybe this is your vision for it but it does not play well at all. It just makes Shotguns the last choice in almost every engagement over anything else but the smallest of calibre handguns. 

The Bulldog price is not worth to buy besides trying to get it from ruins or on chance getting it from a simplemob.

The Buckmaster price is not worth it at all, ever for what it gives you with shotgun efficiency (none)

Not once in almost 200 hours of playing since buckshot has been neutered have I ever been asked to purchase any shotgun outside of ones gotten out of ruins or ones spawned with.

Apogee-dev:
I suggest playing with the buffed slugs instead and seeing how that works out before casting shotguns completely aside. Buckshot for hunting, slugs for humanoids. They won't be quite as good as DMRs, but with those values shotguns firing slugs should be competitive with other long arms while being very cheap (and, if necessary, the price for slugs could be tweaked further to drive that home).

generalthrax:
After some discussion it's probably safe to bring buckshot up to base -20 and -35 on dropoff. We can make additional changes in another pr if needed.

Gristlebee:
Ap is changed again.

PR #4563, opened 2025-04-08
Frontiersman heavies use slugs

Apogee-dev:
## About The Pull Request

Makes frontiersman heavies quite a bit more threatening- and more true to their weapon's supposed role as a "heavy machine gun"- by firing slugs instead of buckshot.

## Why It's Good For The Game

shredder was always intended to do this; makes for a much more intimidating encounter that isn't countered as easily by the intended counter to buckshot (medium armor)

## Changelog

:cl:
balance: Frontiersman heavies now use slugs in their Shredders
/:cl:


FalloutFalcon (approved):
crazy slugs dont have there own type.

PR #4564, opened 2025-04-08
Talos Shortening

firebudgy:
## About The Pull Request

With much assistance working with our mapping lead, this PR shortens the Talos by around 5 tiles, shuffles around the location of the armory/gear room and adds a small infirmary. Additional minor changes include a small adjustment to the MAA's bedroom, a reshuffling of the cafeteria tables and replacing all window spawners with actual windows.

SDMM:
image
SDMM Areas:
image
In-Game
image

## Why It's Good For The Game

The talos itself simply had too much construction space for it's own good. A frequent complaint by many was that it effectively took the wind out of them and they would be building for the entire round, and even then it wouldn't be near done. An additional complaint was that a lack of space for the Corpsman meant they had to sequester away an awkward corner in cargo, and that the Armory was an awkward one-tile navigational space. 

With much help consulting from the mapping lead, this has resulted in the current rework here to address all of those issues without actually adding any additional gear to the Talos, sans a single first aid kit and two small turrets on the front. (I couldn't remember if they were there or not so 

This also removes classified documents that I accidentally mapped in previously when trying to put down folders. Oops.

## Changelog

:cl: Cloudbreak, Latency
balance: Talos now has a small basic medical space, and the wings have been shortened to compensate. Armory completely re-arranged as well.
/:cl:

### DIFF
--- a/_maps/shuttles/inteq/inteq_talos.dmm
+++ b/_maps/shuttles/inteq/inteq_talos.dmm
@@ -29,18 +29,6 @@
 	},
 /turf/open/floor/plasteel/dark,
 /area/ship/storage)
-"aj" = (
-/obj/structure/cable{
-	icon_state = "6-8"
-	},
-/obj/machinery/atmospherics/pipe/simple/scrubbers/hidden/layer4{
-	dir = 10
-	},
-/obj/machinery/atmospherics/pipe/simple/supply/hidden/layer2{
-	dir = 10
-	},
-/turf/open/floor/plating,
-/area/ship/storage/starboard)
 "ak" = (
 /obj/structure/cable{
 	icon_state = "2-8"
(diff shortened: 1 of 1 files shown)


Apogee-dev:
I’m not really a fan of the infirmary (the corpsman cargo corner is an important part of inteq shipfeel imo) or of the fact that you’ve gone from an excess of construction space to almost none at all; the addition of the exosuit bay already cut the build space in half as it is and the space that’s left here is half taken up by incomplete MAA gubbins. Plus the MAA lost his private hidey hole.

The construction space and MAA desk have also been placed behind a room that should be restricted to enforcers too (locker room), so you’ve cut the rest of the crew off from it.

(Frankly i don’t like how the exosuit bay is such a hike away from the rest of the ship, either; gearing up and then heading to the exosuit basically requires circumnavigating the ship and it should have been next to the armory.)

my suggestion: eliminate infirmary, move exosuit bay to nose of starboard wing, put construction space in port wing. The overall shortening’s fine- i agree that there was a bit too much space in the wings overall.

thgvr:
Corpsman needs somewhere to work, where do you propose that is before this change or with this change without the infirmary

firebudgy:
> Corpsman needs somewhere to work, where do you propose that is before this change or with this change without the infirmary

We've got a plan worked out to keep the Infirmary, I'm just in the middle of a course so I can't make the adjustments right now

ionic-engine:
The building space was kind of the best part of the Talos. Engineering projects are fun, and if someone doesn't like engineering, they shouldn't choose to join or spawn a Talos in the first place. Half of it is already gone because of the mechbay anyway.

Apogee-dev:
For the record I had a chat about this with latency to iron out some of my gripes with the PR, so most of what i said higher up’s not relevant anymore, and most of it’s out of scope for this PR anyway (i’ll save it for when talos gets a more comprehensive remap).

I do think the infirmary, when you do tweak the pr a bit, should have a somewhat neglected/underequipped look to it though to reflect it not really being a focus on this ship (and to suggest players work on upgrading it a bit during the round). A little more Vaquero than Valor in character, if you get what I mean (just uh, not literally a 4 tile closet please).

firebudgy:
Changes have been made in accordance with @Apogee-dev and @Latentish's requests. See attached pictures for more details.

Apogee-dev:
Looks good! I’d tweak the enforcer gear room a little though. As cool as the central table is it’s probably going to lead to a lot of shoving when four guys try to pile past it into the airlock. Maybe move it against the left wall or into the corner where the trash can currently is?

There’s actually a second issue with that airlock: The gear room is restricted access, so any shuttle that docks there will just kind of get stuck and have to re-dock at a different port, so it isnt actually very useful to have a port there. 

On a related subject I’d like the airlock on the port wing brought back if you can manage it (that exobay office area is honestly quite enormous). I wish I had a good solution for the cargobay one. Ideally docking ports should not be in front of cargo bays at all, but AACs are still obnoxious for this kind of docking, and moving the docking port to be flush with the door would crush the lights if anything docks. Might have to take the L on the lights and camera so people dont have to spacewalk to get inside. The canera next to the exosuit catapult should be enough anyway.

What I wouldn’t give for a docking priority variable on docking ports, or some way to limit who can dock there…

Gristlebee:
The door to the MAA's office does not have an access lock.

firebudgy:
> The door to the MAA's office does not have an access lock.

Skill issue. (Fixed)

PR #4568, opened 2025-04-08
Bone breaking rework

DrCrawler:
## About The Pull Request

Fracturing a body part and the likelihood of it now depends on the actual force of the attack, becoming easier and more likely to fracture the more your body part is damaged.
This does mean that you can actually break a bone from full health, but right now that requires receiving 50+ brute damage in a single attack, which is exceedingly rare.
This also means that you can no longer break bones from something like decompression, hail, sand storms etc., since it deals less than 5 damage per individual proc call.

Force check for being able to fracture:
- Damage taken >= (Damage Threshold - clamp((Existing damage * 0.5), 0, Damage Threshold - 5)
- Where Damage Threshold = 40 (45 for head, 25 for everything else)

Formula for percentage chance of fracture:
- Raw Force of Weapon/Projectile + Existing Damage * 0.5

## Why It's Good For The Game

Fractures have felt a little unintuitive, mechanically shallow and nonsensical. This aims to solve that through tying the ability to fracture and the probability of inflicting a fracture to the received damage, existing damage and force of the attack.
This also gives armor more importance/usefulness, since it decreases damage taken, which increases the raw attack force required to create a fracture.

## Changelog

:cl:
balance: Fractures now rely more on received damage and the force of the attack.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -91,8 +91,6 @@
 
 	//WS Begin - Splints
 	if(splint_fracture) //Check if it's a splint and the bone is broken
-		if(affecting.body_part in list(CHEST, HEAD)) // Check if it isn't the head or chest
-			to_chat(user, span_warning("You can't splint that bodypart!"))
 		else if(affecting.bone_status == BONE_FLAG_SPLINTED) // Check if it isn't already splinted
 			to_chat(user, span_warning("[C]'s [affecting.name] is already splinted!"))
 		else if(!(affecting.bone_status == BONE_FLAG_BROKEN)) // Check if it's actually broken
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -91,7 +91,7 @@
 
 	//WS Begin - Splints
 	if(splint_fracture) //Check if it's a splint and the bone is broken
-		else if(affecting.bone_status == BONE_FLAG_SPLINTED) // Check if it isn't already splinted
+		if(affecting.bone_status == BONE_FLAG_SPLINTED) // Check if it isn't already splinted
 			to_chat(user, span_warning("[C]'s [affecting.name] is already splinted!"))
 		else if(!(affecting.bone_status == BONE_FLAG_BROKEN)) // Check if it's actually broken
 			to_chat(user, span_warning("[C]'s [affecting.name] isn't broken!"))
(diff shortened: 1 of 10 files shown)


DrCrawler:
~~Hold on, forgot skulls.~~
edit: solved.

Apogee-dev:
Thank god.

rye-rice:
🙏

Rug-Dealer:
no more pranking shipmates by breaking their ribs with thrown items

thgvr:
I'd like to see splints usable on chest with this as well

DrCrawler:
> I'd like to see splints usable on chest with this as well

Will try.

code/game/objects/items/stacks/medical.dm
```diff
@@ -91,8 +91,6 @@
 
 	//WS Begin - Splints
 	if(splint_fracture) //Check if it's a splint and the bone is broken
-		if(affecting.body_part in list(CHEST, HEAD)) // Check if it isn't the head or chest
-			to_chat(user, span_warning("You can't splint that bodypart!"))
 		else if(affecting.bone_status == BONE_FLAG_SPLINTED) // Check if it isn't already splinted
```
SomeguyManperson:
```suggestion
		if(affecting.bone_status == BONE_FLAG_SPLINTED) // Check if it isn't already splinted
```

DrCrawler: woops

DrCrawler:
I should make it easier to break limbs.

Overteh:
Something is definitely wrong, elzuose was beaten with a sledgehammer almost to death, but she didn't get a single fracture

DrCrawler:
Will probably need another TM while I tune the math.

generalthrax:
When simple mobs die, it seems the body that drops has all bones on it break instantly. Unsure if intended or not

DrCrawler:
> When simple mobs die, it seems the body that drops has all bones on it break instantly. Unsure if intended or not

I've only seen that happen with legions, since they do adjustbruteloss(1000) on infection. So far I've been told to leave it, so I'm gonna wait and see if this is a glaring issue or not.

DrCrawler:
I **think** the PR is nearing a merge-ready state? So far feedback, as little as there is, has been mostly positive and there's no bugs associated with this, to my knowledge. Unless something earth-shattering gets uncovered during saturday - my work is finished.
Keep it open until the end of april 13th (sunday), though, just in case something comes up.

FalloutFalcon (approved):

PR #4569, opened 2025-04-09
Sweet Potatos are no longer pathed from Potatos

generalthrax:
## Why It's Good For The Game

It prevents recipe fuckery with crafting, letting you make fries and yaki imo properly

## Changelog

:cl:
fix: Sweet Potato repathed letting you make yaki imo properly
/:cl:

### DIFF
--- a/code/modules/hydroponics/grown/potato.dm
+++ b/code/modules/hydroponics/grown/potato.dm
@@ -15,7 +15,7 @@
 	icon_grow = "potato-grow"
 	icon_dead = "potato-dead"
 	genes = list(/datum/plant_gene/trait/battery)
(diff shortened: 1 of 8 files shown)


code/modules/hydroponics/grown/potato.dm
```diff
@@ -30,19 +30,19 @@
 	distill_reagent = /datum/reagent/consumable/ethanol/vodka
 
 // Sweet Potato
-/obj/item/seeds/potato/sweet
+/obj/item/seeds/potatosweet
```
Erikafox: humble suggest standardizing as 
```suggestion
/obj/item/seeds/sweet_potato
```
instead

generalthrax: true

PR #4581, opened 2025-04-11
Removes/locks behind SRM remaining tribal weapon crafting recipes.

Sadhorizon:
## About The Pull Request
I kept the Sword, the Axe and the Spear in #3296, because there were no other melee weapons available. I think the machettes and combat knives that are now in cargo satisfy it, so I am:
- Removing the bone sword.
- Locking the bone spear and the bone axe behind being an SRM role

## Changelog

:cl:
del: Bone axe and bone sword are no longer craftable.
/:cl:

### DIFF
--- a/code/datums/components/crafting/recipes/weapon.dm
+++ b/code/datums/components/crafting/recipes/weapon.dm
@@ -261,9 +261,20 @@
 
 /datum/crafting_recipe/bonespear
 	name = "Bone Spear"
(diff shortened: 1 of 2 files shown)


FalloutFalcon:
hrm. I honestly still think SRM should be able to craft these. at-least till i finish trickweapons. Can you add them as innate recipes to srm instead.

PR #4587, opened 2025-04-12
add tablet to falm,outh

Erikafox:
## Changelog

:cl:
add: singular tablet to falmouth and also the kilo the kilo has one too now
/:cl:


generalthrax:
can you add it to the kilo too i think it also doesnt have o

FalloutFalcon:
update changelog

PR #4593, opened 2025-04-13
Changes Spitter wear rate to be more durable

generalthrax:
## About The Pull Request

2 to 0.8

## Why It's Good For The Game

Says in the description that its a durable, rugged weapon. I figure a weapon using pretty simple innerworkings wouldn't fuck up super quick.

## Changelog

:cl:
balance: Spitter changed to be 0.8 wear rate instead of 2
/:cl:


Apogee-dev:
Durable doesn’t necessarily mean _reliable,_ and it doesnt have anything to do with cleanliness. Wear rate isn’t modeling the actual weardown and breakage of parts, but of how quickly they get dirty and lose reliability- and simpler designs like this often do run comparatively dirty- so, hm. Arguments could go either way here. Maybe I should go dig up some reliability numbers on e.g. uzi vs mp5 to get an idea of how the closest irl equivalents perform.

PR #4598, opened 2025-04-13
Indie deckhand outfit tweaks

Sadhorizon:
## About The Pull Request
It always irked me that every deckhand spawns with `A somewhat uncomfortable suit designed to be as cheap as possible to manufacture.`

Now, instead, most indies start with a much more neutral black jumpsuit. The utility jumpsuit is, however, preserved for a `/cheap` variant that, similarly to how the Captain outfit does it, is to be used on Miskilamo ships.
Also, makes the Corona use the `/fancy` outfit added with the Ivory.

## Changelog

:cl:
add: Deckhands on non-Miskilamo ships now start with a black jumpsuit instead of a utility one. Corona deckhands also got fancier.
/:cl:

### DIFF
--- a/_maps/configs/independent_corona.json
+++ b/_maps/configs/independent_corona.json
@@ -28,7 +28,7 @@
 			"slots": 1
 		},
 		"Deckhand": {
-			"outfit": "/datum/outfit/job/independent/assistant",
+			"outfit": "/datum/outfit/job/independent/assistant/fancy",
 			"slots": 2
 		}
 	},
--- a/_maps/configs/independent_falmouth.json
+++ b/_maps/configs/independent_falmouth.json
@@ -35,7 +35,7 @@
 			"slots": 1
 		},
 		"Deckhand": {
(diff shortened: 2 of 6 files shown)


Apogee-dev:
The problem here is unfortunately that utility jumpsuits have a better sprite than black jumpsuits (in fact, ALL of the plain color jumpsuits look deeply boring), which is one of the reasons they were made standard deckhand kit to begin with. Frankly i’d rather we just leave it be or tweak the flavor text a bit to play up the cheapness less.

Sadhorizon:
Note that they've just been resprited too. I wouldn't say they look bad anymore!

thgvr:
the utility jumpsuit sprite sucks what are u talkin about

Apogee-dev:
> the utility jumpsuit sprite sucks what are u talkin about

i mean. yeah. utility jumpsuit is fuckin’ crusty. the design’s still more interesting than plain black imo since it has some texture to it, even if it 100% needs a resprite

that’s fair tho, no objections to the pr

PR #4601, opened 2025-04-14
Gets handrails out of the way & adds handrails to Riggs

Apogee-dev:
## About The Pull Request

Moves handrails out from the middle of cargo bay to more appropriate spots against the walls or railings on the Raleigh, Scarab, and Elated Bolide. Adds handrails to the Riggs, which has somehow not had any all this time.

## Changelog

:cl:
add: Added handrails to Riggs
fix: Moved handrails out of the way on Raleigh, Scarab, Elated Bolide
/:cl:


Sadhorizon:
Why did you... remove all the railings from Scarab's cargo bay? That seems really random?

They are there to separate the space and I certainly prefer it that way. 

In fact, removing them created the issue of freestanding handrails which seems... against your PR's goal? The Scarab did not have any of those either.

I would kindly ask you to revert the Scarab changes, as there were no handrail issues there to begin with.

Apogee-dev:
I moved the handrails on Scarab just like the other ships, so there aren’t any freestanding ones there. The reason i culled those railings was the same logic behind moving the handrails; in addition to some handrails being freestanding on the other examples it’s the issue of railings and handrails obstructing cargo space. The railings pose the same IC issue of limiting mobility in an area where you want as much room as possible to move around your cargo and place/retrieve what you need (both IC, where you may be dealing with objects larger than a crate or locker, and OOC, where wrestling with crate tetris to get at something in the back sucks).

The back-to-back railings in the middle also looked pretty weird.

Sadhorizon:
> there aren’t any freestanding ones there

There were none to begin with.

> where you want as much room as possible to move around your cargo

I think they have enough room there. I, as the map's author, like two small cages for organizing different stuff - in fact, that's what's happening right now. Work gear on the right, rest on the left.

> The back-to-back railings in the middle also looked pretty weird.

I'd disagree with that.

> where you may be dealing with objects larger than a crate or locker

The examples that you posted on discord showed 1x2 cargo areas. This is a 3x2. I do not see the issue.

I am still asking you to revert the changes.

PR #4612, opened 2025-04-14
alloys cant be sold

FalloutFalcon:
## About The Pull Request
if a sheet has more then one material it cannot be sold

https://github.com/user-attachments/assets/8bc274ee-d1c6-4f10-8638-a9f95208b2e8

## Why It's Good For The Game
prevents a very annoying issue where you sell an alloy and only recive the value of one of its materials. Autolathes can pull these apart very easily and the dang outpost dont want no dang premixed materials they can do that themselves

## Changelog

:cl:
fix: prevents you from selling material alloy sheets
/:cl:

### DIFF
--- a/code/modules/cargo/exports/materials.dm
+++ b/code/modules/cargo/exports/materials.dm
@@ -30,6 +30,14 @@
 
 	return round(amount/MINERAL_MATERIAL_AMOUNT)
 
+/datum/export/material/applies_to(obj/O, apply_elastic = TRUE)
+	. = ..()
+	if(!isitem(O))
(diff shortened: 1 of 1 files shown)


Erikafox:
I would prefer alloys being worth a minor increase in value and being sellable. I think it makes more sense.

FalloutFalcon:
> I would prefer alloys being worth a minor increase in value and being sellable. I think it makes more sense.

I kinda dislike the idea of people pre alloying there materials before they sell them. Feels weirdly janky to me.

PR #4626, opened 2025-04-16
Adds inspectable posters + species diet guide poster

iimurpyh:
## About The Pull Request

Adds readable posters which show an HTML menu when inspected, similar to paper. Posters can set `inspect_src` to gain this functionality.
Adds one poster to the Contraband pool which uses this feature: a diet guide that shows what food categories can be eaten by what species. Big thanks to RawBread (rawbread7222 on the discord) for making the sprites and icons for this poster.

If I can figure out how to map I'll put the diet guide in kitchen areas on ships like the Sunskipper in a future PR.

dietguidess

## Why It's Good For The Game

Readable posters are useful as a way to add fluff/lore details to random gen posters.
There's no good way to figure out what foods are safe for yourself or others to eat in game besides code diving. The poster is a cute in-universe way of presenting that information.

## Changelog

:cl:
add: Adds a readable species diet guide poster
/:cl:


FalloutFalcon (commented):
this pr is cool, though i think it would be nice if this actually checked the species code so we don't have to update it.
unsure if i like the concept of proliferating static HTML like this for more then just a one off poster.

iimurpyh:
Changed this to a draft because it needs additional work if other posters are to be given inspect text. Main thing I need to add is a way to use images from the CDN which I'd probably accomplish by putting placeholders in the text which get replaced with the loaded image URL.

thgvr (changes requested):
The way this relies on an external filehost to display an image is not good and surely some kind of vulnerability. This would be better suited for something on the wiki.

PR #4630, opened 2025-04-16
Makes the indie viper-23 double action.

Martinpachu:
## About The Pull Request

Before it was single-action, different from the military version which is double action, now they're both the same, with the indie version having a bit more spread because it's shorter. 
## Why It's Good For The Game

Apogee's request. Having the same gun work completely differently between the military and civilian versions is weird
## Changelog

:cl:
balance: The Viper-23 is now double action same as the military variant, with a bit more spread to compensate.

/:cl:


FalloutFalcon (approved):

rivercadaver:
THANK GOD

PR #4662, opened 2025-04-19
Replaces the HBG-7 on the crying sun with the unscoped version.

thearbiber:
## About The Pull Request

Says in the title.
image

## Why It's Good For The Game

This guns already insanely strong and basically an ERT weapon, they can go without doming people across the map with it

## Changelog

:cl:
balance: Replaces the HBG7 with the unscoped version on the crying sun
/:cl:

### DIFF
--- a/_maps/shuttles/pgf/pgf_crying_sun.dmm
+++ b/_maps/shuttles/pgf/pgf_crying_sun.dmm
@@ -240,6 +240,10 @@
 /obj/effect/turf_decal/corner/opaque/purple/border{
 	dir = 1
 	},
+/obj/item/binoculars{
+	pixel_x = 6;
+	pixel_y = -8
+	},
 /turf/open/floor/vault,
 /area/ship/security/armory)
 "bU" = (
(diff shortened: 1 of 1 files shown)


Apogee-dev (approved):
I think it’s good that he died.

DIB-DOG:
I play PGF a lot, so maybe I'm biased but I really don't see this being a massive balance issue. How often is this really being abused? Most fights take place in pretty close quarters where the scope isn't being overly utilized. Even in field fights, it has 20 tiles of total range before it completely stops, which doesn't even reach the end of the scoped view distance. By comparison, a bullet from a ballistic sniper can keep going an unlimited distance as long as it doesn't strike anything. 

If nothing else, at least map in a pair of binoculars to the armory. The scope of the rifle plays a big role in scouting and maneuvering for the marines and it will be missed without something to fill it's place.

thgvr:
Use binoculars or something, this thing was never meant to be mapped

DIB-DOG:
> Use binoculars or something, this thing was never meant to be mapped

Included in my suggestion, yes.

thearbiber:
Binoculars I can see being added, sure.

PR #4664, opened 2025-04-19
reduce value of om5 storm missions

Erikafox:
## Changelog

:cl:

balance: the value of storm scanning has been minorly rebalanced again. Overmap 5 are now worth a fair bit less.


thgvr:
ided

PR #4672, opened 2025-04-20
reverts ex_act nerf for obj

FalloutFalcon:
## About The Pull Request
Reverts the num tweaks of ex_act of om5. Until i am convinced that it was good for the game.
also some defines

## Why It's Good For The Game
tbd. I think shrinking the range of the rand of the light and heavy create more "perdictable" rings around explosions

## Changelog

:cl:
balance: devastating explosions should kill everything again.
balance: light and heavy explosions now have a wider range again
/:cl:

### DIFF
--- a/code/game/objects/obj_defense.dm
+++ b/code/game/objects/obj_defense.dm
@@ -55,15 +55,15 @@
 	if(QDELETED(src))
 		return
 	if(target == src)
-		take_damage(INFINITY, BRUTE, "bomb", 0)
+		take_damage(INFINITY, BRUTE, BOMB, 0)
 		return
 	switch(severity)
-		if(1)
-			take_damage(rand(220, 330), BRUTE, "bomb", 0)
-		if(2)
-			take_damage(rand(120, 180), BRUTE, "bomb", 0)
-		if(3)
-			take_damage(rand(20, 80), BRUTE, "bomb", 0)
+		if(EXPLODE_DEVASTATE)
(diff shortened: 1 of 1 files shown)


FalloutFalcon:
Also i think the ex_act change for /turf/closed/ex_act was good. The rands on that was weird.

thgvr:
perhaps it was done so holes don't get blown in ships from smaller dev explosions

FalloutFalcon:
> perhaps it was done so holes don't get blown in ships from smaller dev explosions

closed walls have 400, so it would do that, but then i dont think it should be a devestating explosion if we dont want it breaking walls. (Acctually thats handled by closed turf proc nvm)

FalloutFalcon:
> perhaps it was done so holes don't get blown in ships from smaller dev explosions

since this is obj it mostly affects mechs, machines, (Even items) stuff like that, a ton of which have over the nerfed devestaion numbers

thearbiber:
If i'm getting what you're saying here, explosions deleting your entire inventory was never a fun thing. Explosives in general were tuned enough to where one bomb could wipe out an entire small ship, space a cargo bay with one frag grenade, unsure if they were effected but even welding tanks could just delete large parts of a ships hull. Believe it's better like that.

FalloutFalcon:
This is part of obj so its specificly NOT the hull.

thgvr:
probably wont actually affect anything noticeably

PR #4678, opened 2025-04-21
Sarathi no longer take toxin damage from welding fuel

SomeguyManperson:
## About The Pull Request

Adds a species chem check for welding fuel on lizards so they don't Die while doing radical pyrotechnics
This also prevents it from metabolizing out currently since it's spent to use the ignite ability but that can be returned if necessary

## Why It's Good For The Game

ough. im gonna vomit. i did the funny thing.

## Changelog

:cl:
balance: sarathi no longer take toxin damage from welding fuel.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
+++ b/code/modules/mob/living/carbon/human/species_types/lizardpeople.dm
@@ -66,6 +66,11 @@
 		internal_lighter = new
 		internal_lighter.Grant(C)
 
+/datum/species/lizard/handle_chemicals(datum/reagent/chem, mob/living/carbon/human/H)
+	if(chem.type == /datum/reagent/fuel)
+		return TRUE
+	return ..()
+
 /datum/action/innate/liz_lighter
 	name = "Ignite"
 	desc = "(Requires you to drink welding fuel beforehand)"


Erikafox:
drinks an entire welding fuel tank and becomes immune to chems

Rug-Dealer:
p1k30wqz5gf51

ritorizo:
> drinks an entire welding fuel tank and becomes immune to chems

A funny fix would be to add an OD threshold that just makes you puke fire.

FalloutFalcon (changes requested):
I feel like this is gonna need to metabolism as there is a bit of jank to being able to store infinite welding fuel in your local Sarathi.

meemofcourse:
yum! so full of ``gasoline``

FalloutFalcon:
If you want them to be able to "store" a charge of welding fuel you could rework the trait so that it like coverts welding fueling in to "charges" of some sort

DIB-DOG:
Why can Sarathi even do this, anyway?

Martinpachu:
> Why can Sarathi even do this, anyway?

dragon

DIB-DOG:
> > Why can Sarathi even do this, anyway?
> 
> dragon

Shouldn’t it be something they can do with more natural “fuel” then? Like, what’s the idea, ancient Sarathi were just going around drinking lamp oil? What about before then? Is this just a power they discovered when someone decided they wanted to try drinking gasoline? Feels really strange is all I’m saying and I don’t think I’ve seen hardly anyone use it. 

Would it make sense to have it charge with the eating of raw meat? Using the oils/fats to create the flame?

thgvr:
pr's fine until new mechanic comes around unless im misunderstanding what it will actually do

what a new mechanic should be is an ability tied to an organ, and that organ "regenerates" charges like xeno plasma or whatever and refills over time & by drinking accelerant and gives access to the flame in hand thing and scales on amount of 'charge' used

FalloutFalcon (approved):

PR #4692, opened 2025-04-23
removes lava lake biome from lava worlds

generalthrax:
## About The Pull Request

removes the lava lake biome from lava worlds. lava still generates but not in extremely laggy oceans that you can possibly spawn in the middle of

## Why It's Good For The Game

image

## Changelog

:cl:
del: Removed lava lake biome from lavaworld generation
/:cl:

### DIFF
--- a/code/datums/mapgen/planetary/LavaGenerator.dm
+++ b/code/datums/mapgen/planetary/LavaGenerator.dm
@@ -285,7 +285,3 @@
 		/obj/structure/flora/ash/tall_shroom = 2,
 	)
 
-/datum/biome/cave/lavaland/lava
-	open_turf_types = list(/turf/open/lava/smooth/lava_land_surface = 1)
-	feature_spawn_chance = 1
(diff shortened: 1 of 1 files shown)


FalloutFalcon:
The only reason its even laggy is the amount of particles. I think we should really just reduce those. (mabye mabye it a prob chance lava even creates a particle generator)

generalthrax:
Lava will still generate. Just removes the biome that lets you land your ship in the middle of like 50 tiles of lava in all directions preventing you from interacting with the planet

FalloutFalcon:
I think its kind an awesome "setpiece" provided we make it easier to build over. I would honestly be fine with making it so you can print those lava proof rods or something.

generalthrax:
I'll let this sit since removing particles or changing lava rods is a bit out of scope, and I recall there being interest in culling this biome anyways.

generalthrax:
After about a week of thinking theres probably a better way to approach this now that the particles were tuned down. I'll probably open another pr letting you tile over them with normal rods instead of needing lavaproof.

PR #4698, opened 2025-04-23
Balance pass over gibonite

FalloutFalcon:
## About The Pull Request
gibonite detonated prematurely does very little (means if someone drills to fast if wont murder them) 
nerfs all the explosions of gibonite but buffs the item light range so its better for mining while not nuking a ship if someone accidentally hits one

## Why It's Good For The Game
alternative to fully removing them, though i dont think this is the only pr required, i think they should be reflavored into some sort of unstable plasma or sulfer that can be refined for a good chunk of it.

## Changelog

:cl:
add: New round tip to let people know you can defuse gibonite with a multi-tool.
balance: Gibonite explosions are weaker, premature explosions do virtually nothing.
balance: The light explosion range of the Gibonite item is a little stronger.
/:cl:


generalthrax:
This is good, but as you said it should probably require the latter reflavouring you mentioned

PR #4701, opened 2025-04-23
ignore spawners in create and destroy

FalloutFalcon:
## About The Pull Request
they circumvent previous ignores (see greeble spawners still causing flaky fails) and pretty much only spawn things we are already testing and is very little chance for the spawner itself to start hardeling


FalloutFalcon:
gonna merge this to stop the frequent flakey fails of greebles.

PR #4716, opened 2025-04-25
Scoopable Goliath Pups and Derecho touch-ups

zimon9:
## About The Pull Request

This PR adds a 10mm ammo box, and three weapon refurbishment kits to the Derecho. It also removes the hand drill and jaws of life from the wrecker's equipment room and replaces it with an angle grinder pack, and a mineral vein scanner. The Captain's locker is also given an assault belt.

This PR also makes goliath pups, and by extension, David the goliath pup, scoopable.

image

## Why It's Good For The Game

Adding a 10mm ammo box allows the Foreman to be able to load their weapon without having to buy a box of ammo roundstart. Weapon refurbishment kits make sense for a ship that may be expected to use their weapons frequently, I think. The hand drill and jaws of life from the wrecker equipment room was replaced with an angle grinder because I felt that they don't serve as much help for salvaging as an angle grinder would. I'm a little unsure about this change, though, so if it may be better to keep the tools, I would be happy to bring them back. I do still think that an angle grinder would be helpful for this ship, regardless. A mineral scanner was added because I can see this ship primarily conducting drill surveys when it's not scrapping, and this would help facilitate that, or to at least encourage it. An assault belt was added to the Captain's locker because it can be helpful to have one to hold equipment to help manage your crew in, rather than having to rely on one's bag.

Goliath pups were made scoopable because I thought it would be nice to be able to do. While goliaths are cow sized, I can see a pup being small enough to be carried around.

## Changelog

:cl:
add: Added a 10mm ammo box, three weapon refurbishment kits, an angle grinder, a mineral vein scanner, and an assault belt to the Derecho
add: Added goliath pup scooping
del: Removed wrecker power tools from the Derecho
/:cl:

### DIFF
--- a/_maps/shuttles/syndicate/syndicate_ngr_derecho.dmm
+++ b/_maps/shuttles/syndicate/syndicate_ngr_derecho.dmm
@@ -848,16 +848,11 @@
 /obj/item/gun/energy/plasmacutter{
 	pixel_y = 15
 	},
-/obj/item/screwdriver/power{
-	pixel_x = 8;
-	pixel_y = 5
-	},
-/obj/item/crowbar/power{
-	pixel_x = -9;
-	pixel_y = -1
-	},
 /obj/machinery/airalarm/directional/south,
 /obj/effect/decal/cleanable/dirt/dust,
+/obj/item/gear_pack/anglegrinder{
+	pixel_y = 3
+	},
 /turf/open/floor/pod,
 /area/ship/storage/equip)
 "hf" = (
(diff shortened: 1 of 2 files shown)


rivercadaver:
hold David gently

Apogee-dev:
In my mind’s eye I see goliath pups being the size of a big stupid Labrador retriever so that works. I think the other changes are all solid QOL improvements too, but I can’t double check the map file for any unexpected issues rn. Derecho pre-dates the angle grinder, and I would have given it a grinder instead of the power tools when i mapped it if they were available at the time.

FalloutFalcon (approved):

zimon9:
> In my mind’s eye I see goliath pups being the size of a big stupid Labrador retriever so that works.

This may be the worst thing I've made yet
The visual is very entertaining, regardless
image

PR #4717, opened 2025-04-26
Remaps The Ivory

Erikafox:
## About The Pull Request
image

Brings the Ivory's style more in line with what's been defined as ISF style while maintaining it's shape and style

overall

lost the absolution for a shadow
pistole c for ringneck
removed mega air tanks
redid decalling

## Changelog

:cl:
add: remaps the ivory. a little less gear, a bit more shape, a smidge more size.
/:cl:


thearbiber (dismissed):
Need to replace the 5.56 here with 10mm. Can't find anything else wrong, should be good for a test after that.
image

PR #4739, opened 2025-04-28
Rolls Back almost every overmap 5 mineral change

generalthrax:
## About The Pull Request

Makes Carbon Steel back into Plasteel and reverses a lot of the ore yield changes from overmap 5, which reduced ore yields ranging from 2-4x lower than it used to be.

Also cuts every single new material from overmap 5. I kept the icons and names of the remaining ores because I'd rather kill myself

May have mucked up with some ore-gen in some places but its mostly the same. Probably

## Why It's Good For The Game

Mining is abysmal dogshit, and carbon was difficult to get even for the extremely low amounts needed for carbon steel, on top of carbon additionally being abysmal dogshit to sell. Different material is probably a cool idea but it needs a more complete implementation than whatever This was.

## Changelog

:cl:
del: Removed carbon steel and all Overmap 5 minerals like copper, silicon, quartzite, lead
add: Re-added plasteel
balance: Ore yield returned to pre-Overmap 5 levels, all lathe recipes (more or less) returned to their old costs
/:cl:


FalloutFalcon (approved):

code/game/objects/items/stacks/sheets/sheet_types.dm
```diff
@@ -78,19 +78,19 @@ GLOBAL_LIST_INIT(plasteel_recipes, list ( \
...
-	singular_name = "carbon steel"
-	desc = "This sheet is an alloy of iron and carbon."
+	name = "plasteel"
+	singular_name = "plasteel sheet"
+	desc = "This sheet is an alloy of iron and plasma."
 	icon = 'icons/obj/materials/sheets.dmi'
 	icon_state = "sheet-plasteel"
 	item_state = "sheet-plasteel"
-	custom_materials = list(/datum/material/iron=3000, /datum/material/carbon=1000)
+	custom_materials = list(/datum/material/iron=2000, /datum/material/plasma=2000)
```
FalloutFalcon:
```suggestion
	custom_materials = list(/datum/material/iron=MINERAL_MATERIAL_AMOUNT, /datum/material/plasma=MINERAL_MATERIAL_AMOUNT)
```

generalthrax:
gonna make this just a broad reversal pr instead of opening another one

generalthrax:
I think I got everything and it doesn't explode when I test so this is probably ready (for review)

FalloutFalcon (changes requested):
the type /obj/item/stack/sheet/mineral/coal was something added 6 years ago when the grill was added. I would revert the removal of that tbh. Not sure the silly grill works without it.

Also i would recommend reverting the type path and name changes to ores.
They used to just be like this but if your removing the book there is no decent way to tell what ore is what unless you go off the color or your a nerd. 
```
/obj/item/stack/ore/iron
	name = "iron ore"
	singular_name = "iron ore chunk"
```
It also was not even fully done as stuff like gold is still just gold.

generalthrax:
I opted to cut the coal item because you don't actually need coal for fueling the grill, you can just use wood which is pretty easy to get. I'll get to fixing the paths and names of the remaining ore, along with probably cutting some of the higher yield ore like magnetite or whatever it was called.

generalthrax:
Got everything for real this time. Pretty sure, at least

FalloutFalcon (approved):

PR #4748, opened 2025-04-29
Renames Inteq Recruits to Auxiliaries

firebudgy:
## About The Pull Request

Renames the Inteq Recruits to Auxiliaries. No gear changes or otherwise.

## Why It's Good For The Game

Recruits tend to be in a strange position on Inteq ships, being distinctly entry-level roles but... compared to other roles like Reservists (CMM), Navyman (PGF) and Deckhand (Indie), they are far more shoehorned into the specific role. This allows them to still fill that role while having a more open set of responsibilities.

Auxiliaries themselves, as a real world unit, are rooted in history as support staff for police or military forces. This ranges from culinary staff to nurses to others. This also includes non-combat roles like radio and civil defense. 

https://en.wikipedia.org/wiki/Auxiliaries

Hopefully this also allows people looking to get introduced to the faction also have the option of simply playing a basic support role.

## Changelog

:cl:
balance: Recruits are now Auxiliaries.
/:cl:

### DIFF
--- a/_maps/configs/inteq_colossus.json
+++ b/_maps/configs/inteq_colossus.json
@@ -41,7 +41,7 @@
 			"outfit": "/datum/outfit/job/inteq/security",
 			"slots": 3
 		},
-		"Recruit": {
+		"Auxiliary": {
 			"outfit": "/datum/outfit/job/inteq/assistant",
 			"slots": 2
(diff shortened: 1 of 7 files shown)


thgvr:
sure

PR #4750, opened 2025-04-29
Dismantle Light Switches + Mount Fix

FslashN:
## About The Pull Request

Ports https://github.com/PentestSS13/Pentest/pull/278

Light switches could not be dismantled from what I've tried, and could only be destroyed. This corrects that. Screwdriver and crowbar can both be used to create a frame after a short delay. Additionally, I have fixed the pixel offset and direction when mounting light switches on the wall.

## Why It's Good For The Game

It hurts my soul to smack light switches until they die.

## Changelog

:cl:
fix: Light switches can now be properly dismantled into frames with crowbar or screwdriver.
fix: Light switches are mounted with the right directional set and with the correct offset.
/:cl:

### DIFF
--- a/code/game/machinery/lightswitch.dm
+++ b/code/game/machinery/lightswitch.dm
@@ -18,7 +18,7 @@
 		dir_amount = 8\
 	)
 
-/obj/machinery/light_switch/Initialize()
+/obj/machinery/light_switch/Initialize(mapload, ndir, building = FALSE)
 	. = ..()
 	if(istext(area))
 		area = text2path(area)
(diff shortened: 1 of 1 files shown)


code/game/machinery/lightswitch.dm
```diff
@@ -64,6 +75,24 @@
...
+		var/action_description = I.tool_behaviour == TOOL_SCREWDRIVER ? "unscrew" : "pry"
+		to_chat(user, span_notice("You start [action_description]ing \the [src]..."))
+
+		if(I.use_tool(src, user, 15, volume=60))
+			user.visible_message(
+				span_notice("[user] [action_description == "pry" ? "pries" : "unscrews"] [src] from its socket."),
+				span_notice("You [action_description] [src] from its socket.")
+			)
+			var/obj/item/wallframe/light_switch/frame = new /obj/item/wallframe/light_switch()
+			frame.forceMove(user.drop_location())
```
FalloutFalcon:
```suggestion
			try_put_in_hand(frame, user)
```

FalloutFalcon (changes requested):
Everything else looks good.

FalloutFalcon (approved):

PR #4751, opened 2025-04-29
Fixes and Improves Turret Controls

FslashN:
## About The Pull Request

Ports: https://github.com/PentestSS13/Pentest/pull/279

Turret controls could not be removed (or attacked with items) once placed, this corrects that. You are now able to attack them with stuff or dismantle them with a wrench if the controls are unlocked. Also fixes issues with mounting them, as they were mounted behind you.

Changed the way turret controls are linked. They are now linked from the main turret control to the turret, so you don't have to run around as much linking them together. Previously they were linked from the turret to the controls, so you needed to get each turret's buffer.

## Why It's Good For The Game

Fix and improvement. Hopefully this will make an easier time for those who like to build their own turret arrays.

## Changelog

:cl:
fix: Turret controls can be taken down with a crowbar and face the right direction when mounted on a wall.
refactor: Turret controls are now linked to individual turrets through their buffer instead of the other way around (minimizes running back and forth).
/:cl:


code/game/machinery/porta_turret/portable_turret_control.dm
```diff
@@ -71,13 +71,31 @@
...
+	switch(I.tool_behaviour)
+		if(TOOL_MULTITOOL)
+			if(locked)
+				to_chat(user, span_warning("The controls are locked."))
+				return
+			if(!multitool_check_buffer(user, I))
+				return
+
+			var/obj/item/multitool/M = I
+			M.buffer = src
```
FalloutFalcon: you should really never use 1 letter vars but saying its only used once and only right below it, its fine i think.

code/game/machinery/porta_turret/portable_turret_control.dm
```diff
@@ -71,13 +71,31 @@
...
+		if(TOOL_WRENCH)
+			if(can_dismantle(user))
+				to_chat(user, span_notice("You start dismantling \the [src]..."))
+				if(I.use_tool(src, user, 40, volume=75, extra_checks=CALLBACK(src, PROC_REF(can_dismantle), user, TRUE)))
+					user.visible_message(
+						span_notice("[user] dismantles \the [src]."), //Generally speaking, \the is unncessary, but I am not sure if most codebases use \improper in their item names.
+						span_notice("You dismantle \the [src].")
+					)
+					var/obj/item/wallframe/turret_control/frame = new /obj/item/wallframe/turret_control()
+					frame.forceMove(user.drop_location())
```
FalloutFalcon:
```suggestion
					try_put_in_hand(frame, user)
```

FalloutFalcon (changes requested):

code/game/machinery/porta_turret/portable_turret_control.dm
FslashN: Probably a victim of copypasta. I will fix it.

FalloutFalcon (approved):

PR #4752, opened 2025-04-29
Allows Building Industrial Suit Storage

FslashN:
## About The Pull Request

Port of: https://github.com/PentestSS13/Pentest/pull/280

Adds the ability to select which version of suit storage you want to build by using a multitool on the circuitboard. Since the change is only cosmetic, I did not want to add another circuitboard.

## Why It's Good For The Game

Disassembling an industrial unit does not allow one to build it again, so this aims to correct that.

## Changelog

:cl:
add: Industrial suit storage can now be built by using a multitool on the circuitboard.
/:cl:


code/game/machinery/suit_storage_unit.dm
```diff
@@ -215,6 +217,15 @@
 		storage = new storage_type(src)
 	update_appearance()
 
+/obj/machinery/suit_storage_unit/industrial/Initialize()
+	. = ..()
+	circuit.build_path = /obj/machinery/suit_storage_unit/industrial
```
FalloutFalcon: i think it would be better to make a dummy sub type of the circuit that has the build path instead to move behavior onto build/compile instead of init

code/game/objects/items/circuitboards/machine_circuitboards.dm
```diff
@@ -1523,6 +1523,24 @@
...
+/obj/item/circuitboard/machine/suit_storage_unit/attackby(obj/item/W, mob/user, params)
+	if(W.tool_behaviour == TOOL_MULTITOOL)
+		var/list/possible_results = list(/obj/machinery/suit_storage_unit, /obj/machinery/suit_storage_unit/industrial)
+		var/list_length = length(possible_results)
+		var/list_position = possible_results.Find(build_path)
+		if(list_position) //FALSE position should never happen, but could with var editing.
+			//There is a modulus way to do this, but no real need to get fancy.
+			build_path = list_position == list_length ? possible_results[1] : possible_results[++list_position]
+			var/obj/machinery/suit_storage_unit/information_holder = build_path
+			to_chat(user, span_notice("You modify the resulting construction. It is now set to [initial(information_holder.name)]."))
```
FalloutFalcon, line 1545: Not really needed if your assuming the list is only two long but good for expandability. No changes needed here.

FalloutFalcon (changes requested):
I like it.

code/game/machinery/suit_storage_unit.dm
FslashN: Yeah, that can be done.

FalloutFalcon (approved):
def a bit cleaner.

PR #4754, opened 2025-04-29
Rename Buttons

FslashN:
## About The Pull Request

Port of: https://github.com/PentestSS13/Pentest/pull/282

Allows one to rename buttons with a pen, provided the maintenance panel is open.

## Why It's Good For The Game

QoL. You can sort of already do this with a hand labeler, but this looks so much nicer and will prevent my untimely death as I forget which button is the holofield and which is the blast doors.

## Changelog

:cl:
add: You can now rename buttons with a pen after opening the maintenance panel with a screwdriver.
/:cl:

### DIFF
--- a/code/game/machinery/buttons.dm
+++ b/code/game/machinery/buttons.dm
@@ -106,6 +106,14 @@
 				playsound(loc, 'sound/items/deconstruct.ogg', 50, TRUE)
 				qdel(src)
 
+		if(istype(W, /obj/item/pen))
+			var/new_name = stripped_input(user, "Enter a new name for the button.", "Rename Button", name, MAX_NAME_LEN)
+			if(!new_name || !panel_open || !in_range(src, user) && loc != user) //Mostly copy pasted from airlocks.
+				return
+			else
+				name = new_name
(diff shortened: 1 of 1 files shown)


code/game/machinery/buttons.dm
```diff
@@ -106,6 +106,14 @@
 				playsound(loc, 'sound/items/deconstruct.ogg', 50, TRUE)
 				qdel(src)
 
+		if(istype(W, /obj/item/pen))
+			var/new_name = stripped_input(user, "Enter a new name for the button.", "Rename Button", name, MAX_NAME_LEN)
+			if(!new_name || !panel_open || !in_range(src, user) && loc != user) //Mostly copy pasted from airlocks.
```
FalloutFalcon: correct me if im wrong but this should do nothing.
```suggestion
			if(!new_name || !panel_open || !in_range(src, user)) //Mostly copy pasted from airlocks.
```

FslashN: Yeah, it probably does nothing. I was looking at it too. I've tested it just in case, doesn't make a difference.

PR #4757, opened 2025-04-29
Ships Only Tile Within Own Area

FslashN:
## About The Pull Request

Port of: https://github.com/PentestSS13/Pentest/pull/285

This fixes an issue where ships would tile with the environment/other ships. There was sort of an exception for this in the code but it was nonfunctional. I refactored the code just a tad and implemented a proper method for checking this.

There is a caveat here that anything built outside of the ship will not immediately tile with it, like when you are trying to expand it. However, I've altered the ship blueprints to smooth the area. Simply redefine your expansion/new area, and it will tile with the rest of the ship.

Example without PR:
trash planet

Example with PR:
Shiptiling Fix Fix

## Why It's Good For The Game

Are you tired of your ship dissolving into a formless blob of metal when visiting any trash planet? So am I.

## Changelog

:cl:
fix: Ships will only tile within themselves and not the environment/other ships. Using the shuttle expansion permit will resmooth new/updated areas.
/:cl:


FalloutFalcon:
alot of people have wanted this as a feature for a while. Very cool pr. Need to properly think about the code itself but a quick skim, it looks fine enough for test merge.

PR #4769, opened 2025-05-02
Fixes Turfs on ICW Base and fixes one wire

generalthrax:
## About The Pull Request

The acid tiles had default atmos before which fucked a bunch of stuff up. There was also one missing tile that rendered the entire north part of the map unpowered.

Also makes the ship turrets into ruin turrets

## Why It's Good For The Game

Fixes good

## Changelog

:cl:
fix: Fixed improper turfs on the ICW Base Waste Ruin that caused wind and fixed one wire
fix: ICW Base turrets are no longer ship turrets
/:cl:


Erikafox:
the turrets on the map should be the /ruin types instead of the /ship types too

PR #4773, opened 2025-05-03
Adds hunting bounties.

Sadhorizon:
## About The Pull Request
Numbers can be adjusted - they were based purely on vibes.
obraz

## Why It's Good For The Game
A new use for fauna loot!
Also, more content for SRM, Sola-class:tm: and the Ivory-class!

## Changelog

:cl:
add: Added hunting bounties - you can now sell all that useless bone and hide for (some) profit!
fix: Added missing trophies to their respective bounty.
/:cl:


FalloutFalcon (approved):
All the code looks good but i am pretty worried about it encouraging farming mobs in a kinda boring way (tho we have ice farming so.). Will have to see how it plays out in test merges. It also technically lets you farm chickens for money which is interesting.....

FalloutFalcon:
A suggestion from thrax.
image

Sadhorizon:
Issue with that is that half of mobs don't have a hide.
Plus, I think a place to loose those items is neat to have too.

As for prices, idk, I'm not an economy person.

generalthrax:
Normal hides could be like 100 and rare, probably upwards of 300. We can probably handle mobs not having hides in another pr

Sadhorizon:
Increased the prices- 50 per normal hide, 250 per rare

PR #4781, opened 2025-05-04
Gauss Gun Fixes

Gristlebee:
## About The Pull Request

Gauss guns now properly consume charge when fired.

Fixes the Claris not removing the deleted caseless rounds from it's magazine so it can actually be reloaded normally without being jammed with phantom bullets taking up space in the magazine.

## Why It's Good For The Game

Fixes are good.
## Changelog

:cl:
fix: Gauss guns properly consume charge.
fix: Claris no longer has ghost bullets haunting it.
/:cl:


thgvr:
outside of this still think the cells should be removed and gauss de-janked in general.

FalloutFalcon (approved):

PR #4812, opened 2025-05-10
Sunskipper grill fix

Sadhorizon:
## About The Pull Request
#4739 removed grill coal from the sunskipper, as well as the ability to use coal in grills at all. This _**TEMPORARILY**_ replaces it with wood. I'll revert this when grills get un-nerfed.

## Why It's Good For The Game
Did you know that coal was a material before OM5, used exclusively for grills? That was dumb, but wood sucks ass and I'm very sad that I **have** to use it now.

## Changelog

:cl:
fix: You can use the Sunskipper grill again.
/:cl:


generalthrax:
Chop down one tree. Coal is not needed for grills when wood is so plentiful

Sadhorizon:
Check what you remove from ships.

generalthrax:
It's a simple one line change to make wood be more efficient as fuel. Feel free to open the pr. I'm not re-adding a fluff material used for one machine that also got the greenlight from maintainers.

Sadhorizon:
Am I asking you to? It will get re-added when carbon is back, which I assume will be at some point.
This is a fix PR for Sunskipper specifically, because its gear was accidently removed.

PR #4814, opened 2025-05-10
Fixes being able to understand sign language without knowing sign language + refactors sign language to be an actual language

SapphicOverload:
## About The Pull Request

Understanding sign language now actually requires knowing sign language.

Sign language has also been refactored into an actual language, and adding other non-spoken languages is now supported.

## Why It's Good For The Game

fix

## Changelog

:cl:
fix: Fixed being able to understand sign language without knowing sign language
code: Refactors sign language into an actual language
/:cl:


thgvr:
This doesn't exactly solve the problem with signing... you still learn it for every language. It's fundamentally flawed.

FalloutFalcon (dismissed):
I worry that this will make it so things like borgs or ghosts wont be able to understand it. 
Could you move this to a trait for understanding sign language?

If you add the trait add it to `/datum/language_holder/proc/grant_all_language and remove_all_languages`
That should fix observers possibly not hearing it as it calls that

SapphicOverload:
Took a while but I've refactored it to be its own proper language instead of the band-aid fix I had before.

FalloutFalcon (commented):
I like it, Fits much better as its own distinct lang rather then somehow knowing a diffrent SL for every language you can speak/understand.
How does this work with partial understanding code however? what does it look like for non-signer?

FalloutFalcon:
image

PR #4825, opened 2025-05-13
Atom integrity and armor

FalloutFalcon:
## About The Pull Request
https://github.com/tgstation/tgstation/pull/61183 and https://github.com/tgstation/tgstation/pull/59474 (only the setting)

Instead of obj integrity and other obj_ intergity procs, most behavior is now handled on the atom layer.
Integrity is now also handled by a setting meaning we can reliably hook a signal onto integrity changes. (could be useful for damage overlays or similar)
renames some limb integ and SM integ stuff so they dont overlap with updated proc names.

## Why It's Good For The Game
I want to move the damage_deflection var up to atoms so i can make walls reflect like windows do. I didnt want to snowflake more code so i did the sensible thing and moved the required behavior up to where i can use it.. Im not going to do it in this pr tho because walls have a fair bit of snow flake code related to integrity i want to tackle in its own pr.

## Changelog
Nothing player facing


FalloutFalcon:
`Rounds Seen    7`

PR #4831, opened 2025-05-16
Prefs Rework

FalloutFalcon:
Extremely indev. Mostly so people can nudge me and maybe get help from others. Going for a commit every day even if its a nothing-burger commit. Will write a real desc when its in a more finished state.


thgvr:
commit every day challenge failed, i may revisit this myself or if you decide to work on this again reopen.

PR #4850, opened 2025-05-20
makes the vickland 10 rounds again + 308 HP sprite instead of Inteq

generalthrax:
## About The Pull Request

makes the vickland require 2 stripper clips again and brings the capacity up to 10. the stripper is now a 5 instead of an 8

pulls out the unused Hunter's Pride 308 sprites since apparently they exist for specialty ammo too and it makes more sense for Them to make it than Fucking Inteq

replaces the ice lodge 308 with 8x50 too

## Why It's Good For The Game

the 8 round stripper looked silly and it was cooler to reload with two strippers. 2 extra rounds dont really mean much

sprite looks better

## Changelog

:cl:
balance: Vickland is 10 rounds again. Strippers are 5 instead of 8
add: Pulls out unused 308 box sprites for 308 to replace Inteq ammo sprite
fix: Ice Lodge now has 8x50 instead of 308
/:cl:


Apogee-dev:
Eehhhhhhh I still think 10 rounds of 8x50 on a semi auto is _a lot_ of damage, especially on a gun with better handling and rate-of-fire than the Woodsman. If you want to keep the double strippers for the cool factor and not looking goofy, you could always make them 4 round clips?

generalthrax:
It didn't really matter before when it was .308 which is pretty much the same round ever since the stat change. It's got some worse zoom than the Woodsman and a clunkier, longer reload too so it's probably fine

Apogee-dev:
8x50mmR _does_ still do more damage than .308 does, but you’re right, the discrepancy’s pretty small.

On a different note, not sure about Hunter’s Pride manufacturing .308 considering they don’t actually make anything that uses it. Iirc there was some talk about renaming it to 8x51mm League and making the box CLIP surplus back when I was redesigning the SKM (sprites coming soon(tm). It’d make sense, considering CLIP were the first to start using that cartridge (in the F4). At present…. Well, inteq probably shouldn’t be making ammo at all, but at least they actually use the stuff. if you don’t wanna whip up a clip box i’d just leave it theirs for now (but like. totally change it later)

generalthrax:
If anything it's a placeholder since its probably better for the Known Gunsmith to have its logo on the box than Inteq For Whatever Reason

Apogee-dev:
yeah that’s fair enough. I’ll see about changing the round and getting it a new box sprite soonish. Ammo repaths are always hell. I remember something about giving 8x50mmR a caliber-notation name to match the other HP cartridges, but ooouuuuughhhhhhh

Imaginos16 (approved):
Ya let's get this merged after having this so long in tm hell!

PR #4851, opened 2025-05-20
Makes the beacon have two shots

generalthrax:
## About The Pull Request

makes the beacon able to shoot two rounds. removes the alof tube from cargo

## Why It's Good For The Game

beacon is kind of a sad weapon and i remember there being talk of redesigning this thing to be two shots. no sprite update in this pr (perspective makes it believable, for now at least)

alofs would make this too jank so its out of cargo. likely be better for the improv shotgun if that shitty thing ever gets a touchup

wanted to throw some other stuff in like .45-70 tweaks for better goliath crunching or adding the burst firemode but it was either out of scope or too broken (burst made it so there was no reason to ever use the single shot)

## Changelog

:cl:
del: Removed alofs tube from cargo
balance: Beacon now can fire two rounds. Sprite unchanged
/:cl:

### DIFF
--- a/code/modules/cargo/packs/weapon_attachments.dm
+++ b/code/modules/cargo/packs/weapon_attachments.dm
@@ -38,15 +38,6 @@
 	faction_discount = 0
 	faction_locked = TRUE
 
-/datum/supply_pack/attachment/alof
-	name = "Alof Tube Crate"
-	desc = "Contains an antiquated spring operated magazine attachment for the HP Beacon. Has a capacity of three rounds."
-	cost = 1000
-	contains = list(/obj/item/attachment/alof)
-	crate_name = "alof crate"
-	faction = /datum/faction/srm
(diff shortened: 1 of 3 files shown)


Apogee-dev:
Ah crap, that’s a callout for me to edit the sprite isn’t it. I’ll put it on the to-do list; over-unders are so much cooler than side-by-sides.

AquillaF:
Oh interesting, I had been planning to ask about just increasing the magazine to 3-4 bullets, as there is at least one example of a trapdoor rifle having a magazine (it was a prototype, but hey it is a fun idea). Also the fact that.. well sprite wise it looks like a trapdoor rifle, and an alof tube would kind of not work with anything other then break-action. Also also, have some shitty testing numbers 
image

FalloutFalcon (approved):

PR #4861, opened 2025-05-20
Improvised Shotgun Sprite Fix

Zevotech:
## About The Pull Request
Fixes the improvised shotgun sprite floating outside of the player's reach by properly setting its inhand dimensions to x64. I did this for the TRABUCO but I just forgot to fix the improv shotgun as well with it.

This is also a PSA for more people to use the improvised shotgun because it makes for a surprisingly effective weapon when you're dead broke and need fauna dead. If you've got double barrel muscle memory then using this thing is just that with one less shot,

image

## Why It's Good For The Game
Broken floating gun sprites are bad.

## Changelog

:cl:
fix: improvised Shotgun sprites are now fixed and are actually in your hand again.
/:cl:


FalloutFalcon:
> use the improvised shotgun

Its so UGGGGGLLLLY tho..... (also in practice its rare to be THIS broke, and that is really just tiny indie ships, most ships have SOMETHING for every to shoot that is going outside)
If I made this as an SRM i think i should be executed lol.

FalloutFalcon (approved):

PR #4869, opened 2025-05-23
removes sand and mud slowdown

generalthrax:
## About The Pull Request

Cuts sand and mud slowdown, setting it to 0.

## Why It's Good For The Game

Most planets that have slowdown tiles usually have significant amounts of ranged enemies (notably hermits, both the rifleman and SKM) on top of the primarily sand generation. It's pretty miserable considering the simplemobs are immune to it.

Snow and water keep it since the risk of getting obliterated by 300 homeless men with assault rifles is significantly lower

## Changelog

:cl:
balance: Sand and mud tiles no longer slow you down
/:cl:


thgvr:
was meaning to do this, we can revisit this after a larger slowdown source audit

FalloutFalcon (approved):

PR #4871, opened 2025-05-23
Shotgun Door Breaching

Erikafox:
## About The Pull Request
You can use a shotgun on a door to force it open. Unless its bolted. Then it's just fried

## Changelog

:cl:
add: You can now use a shotgun on a door to breach through the door
/:cl:


Apogee-dev:
It’d be nice if we could require slug ammo for this. Give slugs another application to give people more reason to keep some handy, you know? They’re often used for this kind of thing irl too so it makes sense.

That and, as far as I can tell, this currently doesnt care which ammo you’re actually using, just that you have a shell chambered? Which means you could breach using beanbags or blanks (which is incredibly funny but probably not intended)

FalloutFalcon:
This should really account for security_level. This should prob not instantly open it if its AIRLOCK_SECURITY_PLASTEEL

generalthrax:
Turns out certain guns pathed from shotguns like the Beacon and Flaming Arrow can also do this

Erikafox:
Great Pathing At Work There

generalthrax:
image

Erikafox:
I guess I'm really gonna have to look at making it a component or element.

code/modules/projectiles/guns/ballistic/toy.dm
```diff
@@ -77,6 +77,7 @@
 	recoil = -10 //its a toy...
 	recoil_unwielded = -10
 	wear_rate = 0
+	door_breaching_weapon = FALSE
```
FalloutFalcon, line 80: what about letting you try cause its kinda funny but does nothing if the projectile doesnt have high enough damage, this would only apply to the toy, the other exceptions should be able to do the interaction at all. this also means you wouldn't need that manual list of exempt projectiles. not required but would be a nice touch

PR #4881, opened 2025-05-25
Dogtags are faster to strip

Gristlebee:
## About The Pull Request

Makes dogtags take one second, down from four seconds to strip off corpses.

## Why It's Good For The Game

QoL. Stripping dogtags off corpses is flavourful and thematic, but does get a bit tedious, especially when there are a large number of corpses I find.

## Changelog

:cl:
balance: Dogtags are faster to strip off corpses.
/:cl:


FalloutFalcon (approved):
Half of the point of irl dogtags is your ability to remove them.

PR #4903, opened 2025-05-31
Stops base overmap from getting random colors

thgvr:
## About The Pull Request
Consistency is better than having this show up sometimes 
image

## Changelog
:cl:
del: Overmap will no longer randomly recolor 1/10 times
/:cl:


FalloutFalcon (approved):
On top of the feature being questionable. This is a bad way to handle the code. The subsystem should just create a random datum, not the datum copying data from a random other datum.

PR #4832, opened 2025-05-16
More newfood updates

thgvr:
## About The Pull Request
Continues #4101
Pretty much had to remake it from how fucked with conflicts it was

## Changelog

:cl:
add: A lot of food changes, notably steaks don't come with plates
/:cl:

PR #4910, opened 2025-06-05
readds easter and bunnies

FalloutFalcon:
## About The Pull Request
both improperly removed in https://github.com/shiptest-ss13/Shiptest/pull/4832
bunnies are fine but the way they were removed meant bunnies used in map gen were likely invisible
i went ahead and removed the egg laying variant but they are still pathed under chickens for now.
i left out stuff like adding easter eggs to maint loot. 

## Changelog

:cl:
fix: bunnies should function properly again
/:cl:


Ms-Mee:
As a bunny I can confirm that bunnies laying eggs during easter is simply a myth and does not happen

Don't fact check this

Aquidu:
yep. we're so back.

thgvr:
simple mob freeze :clueless:

FalloutFalcon:
> simple mob freeze :clueless:

fairly certin i dont need to update the checks because you never removed subtypes meaning the bunny type techinicly exists

PR #4914, opened 2025-06-06
Replaces the Talos' TEG with two turbines

SapphicOverload:
## About The Pull Request

Removes the TEG and a few thrusters from the Talos and replaces them with two methane-burning turbines. This also fixes methane combustion resulting in the wrong byproducts and adds support for the gas turbine computer to control multiple turbines in the same interface.

<details>
<summary>Images</summary>

image

image

image

image

</details>

## Why It's Good For The Game

The TEG is an inherently bad choice of a power source for most ships because the lack of space for a proper cooling loop forces ships to use their own fuel-oxidizer mixture to cool the cold side, causing it to ignite inside the pipe network upon heating up which vastly reduces throughput and by extension power generation. This can be overcome by separating fuel and oxidizer and sending only one through the cold side and injecting them into the combustion chamber separately, but that demands additional space for separate fuel and oxidizer reserve tanks in this ship's already crowded engine room. The only reason it can even meet power demands is thanks to plasma combustion being so broken that reaching into the tens of millions kelvin is comically easy, something I haven't yet been able to fix.

To solve this, I decided to replace the TEG on the Talos with two turbines. I decided on two because the ship is an even number of tiles wide and would look very strange with a turbine not quite lined up properly, and because a single stock turbine would not provide enough power for such a large ship mid-flight. Methane seemed a better fuel choice with its higher energy density compared to hydrogen in order to sustain running two turbines at once.

## Changelog

:cl:
add: Replaced the Talos' TEG with two turbines
add: Adds support for controlling multiple turbines from one turbine control computer
fix: Fixed methane combustion producing the wrong byproducts
/:cl:


thearbiber:
Not really a fan of this, the TEG on the talos currently works fine enough, not to mention how quickly this would run out of fuel.

SapphicOverload:
> Not really a fan of this, the TEG on the talos currently works fine enough, not to mention how quickly this would run out of fuel.

As I stated in the PR body itself, the only reason it works fine is because plasma combustion is broken and easily reaches absurd temperatures that shouldn't be possible just by burning plasma.

And, when testing the dual turbine setup, I left it running for an hour and came back to see a quarter of the fuel tank used up, giving it an estimated 4 hours of non-stop use. By no means does it even need to run constantly for that long, but it can.

thgvr:
turbine > teg tbh. TEG is just really janky

thearbiber:
Apologies, but I'm going to be closing this for now. The TEG may be janky but I do not see the vision of adding two turbines, and the layout of it/how it looks in general just does not please me. I'm not confident of the power generation after loading it up myself, and overall I'd prefer a working TEG over this. Still, thank you for your contribution.

SapphicOverload:
> Apologies, but I'm going to be closing this for now. The TEG may be janky but I do not see the vision of adding two turbines, and the layout of it/how it looks in general just does not please me. I'm not confident of the power generation after loading it up myself, and overall I'd prefer a working TEG over this. Still, thank you for your contribution.

I don't understand the problem with the layout or the idea of having more than one turbine, can you elaborate on what's wrong with it that I can't address?

Also, when I tested this myself I had no trouble generating enough power even at the default settings, and it can easily be turned up or upgraded for more power if needed. I did notice power consumption is abnormally high until the telecomms power storage unit is full, but that can be easily fixed by turning down its target input and output settings to be more reasonable.

thearbiber:
The interior changes here are pretty strange to me flow wise, regarding the turbines, computers, and tank locations. It is slightly less cramped than the current talos in some places, but again it more so feels unnatural and slapped on. Specifically the SMES and top turbine do not feel like natural placements. Beyond that, I do not believe we should start phasing out the TEG from large ships. It does behave slightly strange currently, however this is something that can be fixed in the future, and I'd rather a better working TEG than a double turbine setup. It's been working for quite a while just fine, keep in mind. It is also possible to get a proper cooling loop on the talos with minimal changes, however if the engine bay is being reworked I'd rather a remapping of the backend interior keeping the TEG than a double turbine replacement. 

Turbines are fun, and are a lot easier to understand/work with, but I feel their place is mostly on smaller/medium sized ships as a more unique form for power gen that isnt just a pacman. Not as a full TEG replacement.

PR #4928, opened 2025-06-08
[Please God Testmerge This] Ports Jitter/Drunkenness/Dizziness as status effects

Erikafox:
## About The Pull Request

Ports Jitter/Drunkenness/Dizziness as status effects, which should make them less weird to deal with. Generally speaking.

## Why It's Good For The Game

I don't like the 12000 jitter stack making you look like a crackhead behind 7/11 going through withdrawls (drank an energy drink)

## Changelog

:cl: melbert,  Erika (porting fox)
refactor: Drunkenness is now a status effect. The drunk alert should no longer persist on aheals and in some rare cases, and should overrall act a bit more consistently.
refactor: Dizziness is now a status effect, similar to above. You no longer should be stuck with an off center client view if you get very dizzy.
refactor: Refactored status effect examine text. Should overall have slightly better grammar and more consistent span uses.
refactor: Jitterriness is now a status effect
del: some contractor related cruft that i didn't want to update
/:cl:


thgvr:
i need crack

PR #4937, opened 2025-06-12
Fixes hotspots deleting themselves after being created

SapphicOverload:
## About The Pull Request

Fixes `perform_exposure` not returning `TRUE` when successful, causing hotspots to always delete themselves instantly.

## Changelog

:cl:
fix: Fixed hotspots immediately deleting themselves
/:cl:


thgvr:
This may have unforeseen consequences

PR #4957, opened 2025-06-19
ships can no longer hail themselves

Erikafox:
## Changelog

:cl:
balance: ships can no longer hail themselves
/:cl:


Aquidu:
i actually don't like this because this is very useful for getting everyone's attention on a big ship when they might not all have the frequency yet. i would like to see command announcements added as a substitute for this before you remove it. wrangling a large crew is already hard enough

PR #4966, opened 2025-06-20
diseases respect mob stasis

Erikafox:
## Changelog

:cl:
balance: diseases respect mobs being in stasis now
/:cl:


Erikafox:
>look up the proc chain
>it should already be doing this

oughhh i hate life.dm

PR #4992, opened 2025-06-25
Removes combat exosuits from cargo + mech rocket removal

generalthrax:
## About The Pull Request

Removes the gygax, NT gygax, and the durand from cargo. Equipment and upgrade kits remain (just in case)

Also removes the BRM-6 from cargo and the SRM-8 from cavecrew

## Why It's Good For The Game

-They trivialize PVE content (including all ruins and drills of all difficulties), almost nothing is able to challenge a crew with a fully-operational combat exosuit
-You'd have to constantly W-key into the hardest ruin never repairing to probably even have a chance to be killed
-You can just ignore like half the drill mobs because they can't deal damage to you
-Repair is slow but free and you can just do it and continue on your way

-They heavily skew PVP into a stomp for the side that has them, so much hangs on the Suit Alone for victory
-Their counters include: ions (do almost nothing), high-calibre automatic fire (easily dealt with by just trading fire with the person shooting at you and killing them), rockets (oneshot and lame), taipan (need to hit all 3 shots on a moving target with high recoil without getting killed)
-Most combat I've personally seen and participated in has devolved into one side constantly withdrawing due to being unable to do anything to an exosuit and then the slaughter when they can run no more. Not very interesting

But why don't we make changes to fit them into the loop better?

-Decrease armour to make them more balanced? Brings back a lot of the issues they had before where they weren't worth their price because they were so flimsy and most guns could permanently put them out of commission, and put a huge sinker in your budget
-Increase the price? Only would delay the problem to later in the round, which Shiptest rounds often last long anyways
-Decrease stored ammo and increase maintenance prices even more? Was already done and from observation didn't help much, and adjusting these further would either make them not worth using or holding comically small amounts of ammunition for their high upkeep
-More system faults that happen upon taking more damage? Theres already a fair few of them and almost all of them suck, and I'm not convinced more would do much
-More expensive repair? Only solves one problem of having to invest to keep your suit maintained, but doesn't prevent you from stomping out the enemy / a ruin anyways
-Increase ruin / enemy difficulty to compensate for mechs? Mappers shouldn't have to increase the difficulty of their ruins to compensate for a no-slowdown, high-armour exosuit, and would just make things miserable to go through for a normal crew

## Changelog

:cl:
del: Removed combat exosuits from cargo. Other methods of getting player-accessible exosuits unchanged.
del: Removed the BRM-6 breaching rockets from cargo and removes SRM-8 missile rack from Jungle Cavecrew
/:cl:

### DIFF
--- a/_maps/RandomRuins/JungleRuins/jungle_cavecrew.dmm
+++ b/_maps/RandomRuins/JungleRuins/jungle_cavecrew.dmm
@@ -2482,7 +2482,6 @@
 "Ec" = (
 /obj/effect/turf_decal/industrial/outline/yellow,
 /obj/effect/decal/cleanable/dirt,
-/obj/item/mecha_parts/mecha_equipment/weapon/ballistic/missile_rack,
 /obj/item/mecha_parts/mecha_equipment/weapon/ballistic/carbine,
 /obj/structure/closet/crate/secure/weapon,
 /turf/open/floor/plasteel/patterned/brushed,
--- a/code/modules/cargo/packs/mechs.dm
+++ b/code/modules/cargo/packs/mechs.dm
@@ -391,13 +391,13 @@ weapons
 	faction = /datum/faction/nt
 	faction_discount = 20
 
-/datum/supply_pack/mech/weapon/missile_rack
+/*/datum/supply_pack/mech/weapon/missile_rack
 	name = "BRM-6 kit"
 	desc = "Contains a low-explosive missile launcher, excellent for breaching through obstacles."
 	cost = 3000
 	contains = list(
 		/obj/item/mecha_parts/mecha_equipment/weapon/ballistic/missile_rack/breaching
-	)
+	)*/
 
 /*
 ammo
@@ -434,10 +434,10 @@ ammo
 	faction = /datum/faction/nt
 	faction_discount = 20
 
-/datum/supply_pack/mech/ammo/missile_rack_ammo
+/*/datum/supply_pack/mech/ammo/missile_rack_ammo
 	name = "BRM-6 missile box"
 	desc = "Contains a box of six breaching missiles designed to explode upon striking hard surfaces."
 	cost = 600 //these are remarkably inefficient against anything except exosuits
 	contains = list(
 		/obj/item/mecha_ammo/missiles_br
-	)
+	)*/
--- a/code/modules/cargo/packs/mechs.dm
+++ b/code/modules/cargo/packs/mechs.dm
@@ -56,69 +56,6 @@ Build Your Own Suit
 	faction = /datum/faction/syndicate/cybersun
 	faction_discount = 40
 
-/*/datum/supply_pack/mech/gygax_parts
-	name = "501p construction kit"
-	desc = "An agile combat exosuit sold across the galaxy by Cybersun. Or at least the parts to it."
-	cost = 12000
-	contains = list(
-		/obj/item/mecha_parts/chassis/gygax,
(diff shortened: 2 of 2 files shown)


thgvr:
yeahg

thgvr:
just remove the commented out stuff also instead of commenting it

githubuser4141:
just make the exosuit feel more believable, I am not a maths genius (I suck at it even) but try imagining what the mech is actually made out of, where the ammo is stored / how much ammo there is, make it use REAL AMMO rather than abstracted mech ammo (the same ammo everything else uses), how thick / tough the armor actually is, is it sacrificial armor that wears away when damaged, how fast / agile / situationally aware it could actually be? I think it's unbelievable that mechs could be so unstoppable for their weight/size and ease of use, it goes against the rest of the setting's tech level

thgvr:
no need for ideasguying here

thearbiber:
probably for the best considering mech rework is estimated to come out in 9 business years

Madtankdriver:
Just my piece of mind but I don't really this is the right call. Untill you have an alternative I don't think you should just remove stuff like this just because it seems "unfair". It's a combat mech for crying outloud. I know no matter what I say it just won't matter but this is good content that can't just be removed without consequences. 

Some peoples character's backgrounds are build around piloting combat mechs and some shuttles are build around using combat mechs for clearing tough ruins. In my mind PvP will just end up with whoever has more players on their shuttle will almost always win. People may also get much scummier with how they fight which may not be enjoyable. (Abusing explosives, spam using snipers) 

Though, I do agree the SRM-8 and the BRM-6 are a bit much and should be cut.

Combat mechs in my mind are a way for smaller crews like crews of 5, be able to counter much larger crews
then their own or deal with threats like very dangerous drill missions and really big ruins. It isn't fun to be brutally beat into a pulp because the other crew has 2 times the crew you have.

(An example could be 2 PGF shuttles, a total of 24 people alongside a subshuttle to breach vs a independent small crew of 5. Seem like a fair fight? Think that Gygax could even out the odds just a bit for the indie crew?)

In the end and in my opinion untill you have an alterative to combat mechs, PLEASE don't remove the combat mechs. Just much like R&D.

thgvr:
There will never be an alternative to mechs as they are now. Something like this that doesn't interact with nearly any system in the game and allows you to effortlessly clear any PVE challenge our mappers create doesn't belong. There's been a lot of ruins scaled up due to general player power levels rising, and even the hardest of those are made trivial by any mech.

FalloutFalcon:
> PvP will just end up with whoever has more players on their shuttle

Everything you say here about PvP assumes that for some reason the larger ship WONT have a mech and that the 5 crew indie ship is the only one with one. I also dont think "The crew with more money wins" is all that different then "The crew with more crew wins" neither are particularly interesting.
The inability to access mechs prob has a negligible effect on people using "scummier" methods as most of them were already the best way to deal with a mech. (Explosives)

FalloutFalcon (approved):
I do however agree that downsides of mechs could/should be addressed but right now they are negatively affecting the game in the ways thgvr and thrax already described so its pretty much a WYCI situation.

joelTheLostOne:
As Resident Mech Fan Numero Tres (Second to MADTANK and probably some other freak) I do kind of agree with this change as a temporary measure, as long as a mech rework is still cooking, I'm fine with it. It's just a bit of a shame because factional mechs were so fucking cool god DAMN IT I LOVE FACTIONAL MECHS MAN

generalthrax:
i know the most permanent change is a temporary one but it was gonna happen sooner or later

PR #4993, opened 2025-06-26
Disables EMP storms and reduces volume of overmap event sounds

thgvr:
## About The Pull Request
title

## Why It's Good For The Game
I disabled emp overmap hazards a while ago due to the exact same issues that are once again popping up, being that it's very easy to softlock a ship and doubly so when these can spawn directly on top of outposts.

## Changelog

:cl:

del: Disabled EMP storm overmap hazards, as well as the associated mission
soundadd: Reduced volume of almost every overmap hazard alert sound

/:cl:

PR #2345, opened 2023-09-08
disable ion storms

thgvr:
## About The Pull Request
title

## Why It's Good For The Game
Sucks for players and admins alike

## Changelog

:cl:
del: ion storms are removed
/:cl:

PR #4996, opened 2025-06-26
Prevents ion storms from softlocking ships

SapphicOverload:
## About The Pull Request

Significantly reduces the time ion storms disable APCs for, preventing them from softlocking ships. Considering ion storms are the only source of EMPs players are likely to ever encounter, I felt this was the right solution to the problem without removing them from the game entirely.

Alternative to #4993

## Why It's Good For The Game

https://github.com/user-attachments/assets/8927fe30-3567-4849-a284-6a299cda96a5

## Changelog

:cl:
balance: Significantly reduced the time ion storms disable APCs for, preventing softlocks
/:cl:


Anticept:
It would also be a good idea to take out the power draining effects as well, or at least make it something that drains *slowly* over time instead of boom, 90% of your stored power is gone.

thgvr:
This is definitely not the only issue and disabling power at all means you will hit another hazard or be slowed down permanently in the storm.

SapphicOverload:
> This is definitely not the only issue

What else is wrong with them that needs to be fixed?

> and disabling power at all means you will hit another hazard or be slowed down permanently in the storm.

I don't think the possibility of running into another hazard is necessarily too harsh, considering how fast you would need to be moving to be disabled long enough to end up on another tile by the time it's over anyway. I think it's more interesting when there are *some* consequences for not being careful.

Anticept:
> What else is wrong with them that needs to be fixed?

They can hit the outpost too, which should probably be a safe spot from overmap events tbh.

thgvr:
Well these storms were brought back one way or another after being disabled some years ago in [2345](https://github.com/shiptest-ss13/Shiptest/pull/2345), and this PR doesn't actually push forward any changes that would make it remotely reasonable to have ion storms spawn this much. It's pretty much just an issue of the way EMPs function, as well as overmap hazards this dangerous that move around and can potentially spawn on top of you just not being a good idea or actually that interesting for anyone maybe other than the person piloting

FalloutFalcon (approved):
Made semi redundant by #4993 but an improvement to the behavior regardless

PR #5001, opened 2025-06-27
Exosuit overheating mechanics + readds combat exosuits to cargo

SapphicOverload:
## About The Pull Request

Exosuits now have an overheat level which is built up when firing weapons, moving with the actuator overload active, absorbing damage with defense mode, or being hit by EMPs and ion weaponry. Exceeding 100 overheat will cause the exosuit to start sparking and take damage over time, and when reaching the cap of 150 the exosuit will shut down completely for a few seconds, unable to move or use any equipment or abilities. Overheat dissipates over time by 5 per second, and ramps up to 10 per second over time when standing still.

This also adds a structural damage system for exosuits, requiring titanium sheets (or steel for non-combat exosuits) in order to fully repair them.

New equipment has been added to help mitigate overheating:
- External Radiator - Increases cooling rate by 1.5 per second. Requires an atmosphere.
- Auxiliary Cooling System - Increases cooling rate by 3 per second. Disabled by EMPs/ions.
- Heat Sink - Slows down both heating and cooling by 20%.

A number of other changes are included as well:
- Gygax maximum integrity reduced by 20%. (300 -> 240)
- Ion rifle price reduced by 70%. (10000 -> 3000)
- Melee weapons' demolition modifier now affects exosuits.
- Readded Downpour and Solaris to cargo.
- Combat exosuits are readded to cargo for their respective factions, as well as the black market.
- NT-MP/Basenji charge now has an after-image effect when charging.
- Fixed Durand defense mode shield appearing detached from the exosuit when moving.

<details>
<summary>Demonstration</summary>

https://github.com/user-attachments/assets/6ecb3a20-0c2a-4071-af38-1b0614e48e73

https://github.com/user-attachments/assets/abcf04c9-8ad2-4353-b7b0-1ba9c8aae4e3

image

image

</details>

Original PR: yogstation13/Yogstation#22538
After-image effect ported from here: yogstation13/Yogstation#21557
Partially reverts #4992

## Why It's Good For The Game

This addresses many of the main issues with combat exosuit balance:
- Sustained damage being limited only by ammo capacity, making energy weapons too strong to be available by normal means
- Lack of a weapon balancing mechanic other than power consumption
- Effectiveness and cost of intended counters
- Gygax having no weakness to compensate for having the best mobility
- Damage to exosuits being only a minor setback unless completely destroyed

## Changelog

:cl: sapphicoverload, moltijoe
add: Added exosuit overheating mechanics
add: Added exosuit structural damage
add: Added after-image effects to NT-MP and Basenji when charging
add: Combat exosuits readded to the black market, and to their respective faction's cargo
add: Downpour and Solaris readded to cargo
balance: Gygax maximum health reduced by 20% (300 -> 240)
balance: Reduced ion rifle price by 70% (10000 -> 3000)
balance: Exosuits are now affected by demolition modifiers
fix: Fixed Durand defense mode's shield not lining up properly when moving
/:cl:


thgvr (changes requested):
This is absolutely not enough or a good direction for mechs and this mechanic just kind of tacks more jank on top of mechs. Nothing here addresses anyone's issues, especially mine, with mechs and you're re-adding them to cargo here. Genuinely no solution will fix mechs as they are in the game now rather than just cutting them entirely and revisiting the concept.

SapphicOverload:
Can you elaborate a bit? Why do you think it's not even worth trying to fix them?

thgvr:
They ignore pretty much every gameplay, mechanical, and health system

generalthrax:
Don't believe anything here really fixes anything besides encouraging players to let their exosuits cool off when they aren't pushing (which they often spend repairing anyways). Seems to weaken exosuits in drawn out fights as well, which exosuits handily prevent, as exosuits win fights often within the opening minutes of a fight breaking out.

On the other changes:
-Melee has been and always will be suicidal against exosuits
-Cheap ions just makes instakill IPC / FBP / prosthetic guns more accessible. Ions barely even do anything to exosuits with good capacitors and good cells anyways (which almost anyone will do with money)
-Gygax armour being reduced just brings it partially back to older values, which made them problematic in an entirely different way

SapphicOverload:
> They ignore pretty much every gameplay, mechanical, and health system

This is hyperbole. Yes, they're not affected by certain things. They also have their own counters as well, which are now a lot easier to acquire and are more effective than before. This PR doesn't have to be the way exosuits must stay forever if merged, more work can be done on them in the future. I believe that this is a step in the right direction at the very least, and a lot of people like exosuits as a feature and don't want them to have to be removed.

ritorizo:
Combat exosuits behing on top of the meta makes sense since they are armored combat vehicle, which is imo a nice niche to have in a setting like shiptest. They could be balanced by either being made flimsier or slower (depending on the model probably). But they will always be stronger than infantry.

thgvr:
maybe about 5% of the overall consideration was pvp and right now you seem to be under the impression that is the entire reason, which is not true

as I've said before it's pretty much impossible to balance anything when crews can relatively easily acquire a mech from outpost supply that trivializes any mob-based challenge which a lot of our content and ruins rely on, and as we expand those features mechs will just continue to be as trivializing to any encounter as they are now

and, yes, mechs do ignore pretty much every normal gameplay mechanic as well as the entire health system, in favor of using its own barebones durability system with dirt cheap simple repairs and basic power usage.

making ion weaponry super cheap is also not really good too, we have a pretty big focus on synthetic life and honestly they should probably get kicked out of the outpost catalogue anyways.

I also don't really think it's a good idea to come with the quick knee jerk when there hasn't even been an uptime with mechs not purchaseable

SapphicOverload:
> On the other changes: -Melee has been and always will be suicidal against exosuits -Cheap ions just makes instakill IPC / FBP / prosthetic guns more accessible. Ions barely even do anything to exosuits with good capacitors and good cells anyways (which almost anyone will do with money) -Gygax armour being reduced just brings it partially back to older values, which made them problematic in an entirely different way

- Melee is not meant to be effective against exosuits, I just thought at the very least demolition modifier should still apply to them anyway. Though locking it down could open a window for someone to get in close with a sledgehammer or something.
- Five to six direct hits with an ion rifle is enough to fully lock down an exosuit by overheating it, assuming it had no overheat from any other source which is very unlikely during a fight.
- IPC/FBP balance being that way was always a terrible idea, something will need to be done about it but that's out of scope for this PR. Considering how rare PvP is and the fact ion rifles will only ever be bought to be used against exosuits, I don't think this is likely to have much of an effect.
- Gygax is already in a bad position because of it being both tough and incredibly fast, the exact numbers can be tweaked if this ends up being too far in the other direction.

SapphicOverload:
> maybe about 5% of the overall consideration was pvp and right now you seem to be under the impression that is the entire reason, which is not true
> 
> as I've said before it's pretty much impossible to balance anything when crews can relatively easily acquire a mech from outpost supply that trivializes any mob-based challenge which a lot of our content and ruins rely on, and as we expand those features mechs will just continue to be as trivializing to any encounter as they are now

Considering the cost of a combat exosuit, by the time you're able to afford one you're already swimming in credits and don't have much of a reason to continue clearing ruins and such anyway, not that it's much of a challenge without one considering what else those credits are usually spent on.

> and, yes, mechs do ignore pretty much every normal gameplay mechanic as well as the entire health system, in favor of using its own barebones durability system with dirt cheap simple repairs and basic power usage.

There are a lot of potential solutions to this, maybe requiring a resource or a kit from cargo to repair them or adding a limb-based damage system.

> making ion weaponry super cheap is also not really good too, we have a pretty big focus on synthetic life and honestly they should probably get kicked out of the outpost catalogue anyways.

3000 credits is still not exactly cheap, and I doubt anyone will be spending that on a special gun to kill one synthetic in particular you don't like when you can just shoot them with a normal gun that you already have to kill them normally.

> I also don't really think it's a good idea to come with the quick knee jerk when there hasn't even been an uptime with mechs not purchaseable

As if I haven't been working on this for weeks.

githubuser4141:
meh, that's why I thought having all mechs be able to equip every item would be a good idea. Then you could just remove or soft-remove combat mechs (armored mechs) entirely, nerf Ripley armor, and presto mechs would be balanced - mostly designed for utility, and let you put weapons on the Odysseus/Ripley/Whatever as you see fit. But they wouldn't be special too-powerful dedicated combat mechs.

SapphicOverload:
I've added a structural damage system, taking damage past the first 25% of the exosuit's integrity will add half of any additional damage taken as structural damage, which must be repaired with either titanium for combat exosuits or sheet metal for non-combat exosuits.

thgvr:
ultimately, still, I don't think either of these solve any real problems with mechs and I'm not interested in merging them

FalloutFalcon:
I would recommend reverting the revert of the cargo stuff. I think the core concern is that this pr may not be enough to readd them to cargo but I dont see a strong reason against the PR outside what thgvr has said about it ignoring other systems/mechanics but I think this system doesn't really apply here? I could see it working WITH systems tbh as you could possibly tie it in with the existing temp system (im not requring this in the scope of this pr)? Heating is more efficient in colder environments and weapons that deal temp damage (I think the only instance of this rn is acctually watcher shots?)

SapphicOverload:
Moved them from the open market to just their respective factions or the black market.

thgvr:
Nobody has interest in merging this, sorry.

PR #5005, opened 2025-06-27
Deleted an empty file that someone commited by accident.

Anticept:
Per title. Nothing special, just a little cleanup.


thgvr:
wtf

PR #5014, opened 2025-06-29
Makes Hexacrete have a sane recipe

firebudgy:
## About The Pull Request

Replaces Phenol in Hexacrete's recipe with Stable Plasma. 

## Why It's Good For The Game

This makes Hexacrete actually worth making, as Phenol is a pain to mix in bulk to make walls and flooring. Concrete already is used pretty rarely, and you never see Hexacrete because of the Phenol requirements.

## Changelog

:cl:
balance: Hexacrete now requires Stable Plasma to create.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/recipes/others.dm
+++ b/code/modules/reagents/chemistry/recipes/others.dm
@@ -602,7 +602,7 @@
 
 /datum/chemical_reaction/hexement
 	results = list(/datum/reagent/cement/hexement = 1)
-	required_reagents = list(/datum/reagent/cement = 6, /datum/reagent/phenol = 1)
+	required_reagents = list(/datum/reagent/cement = 6, /datum/reagent/stable_plasma = 1)
 	required_temp = 400
 	mix_message = "The mixture rapidly condenses and darkens in color..."
 


Anticept:
I unironically use barrels to mix mass amounts of any chem if I need it, but the chlorine is still a pain in the ass to get in bulk short of buying a big jug at the outpost. There's some stuff you can grind and alchemy for it though.

Alternative suggestion: increase the amount of hexcrete output.

`results = list(/datum/reagent/cement/hexement = 7 )` for example, which would be 1:1 with inputs (6 cement, 1 phenol)

Also, cement should probably output 8-10 units per mix too. Some water evaporates from cement mixes, but much of it actually cures and becomes part of the concrete, so mixing up a batch of concrete should also be close to 1:1 input to output. As it stands right now you get 1:0.5 total inputs to outputs.

firebudgy:
> I unironically use barrels to mix mass amounts of any chem if I need it, but the chlorine is still a pain in the ass to get in bulk short of buying a big jug at the outpost. There's some stuff you can grind and alchemy for it though.
> 
> Alternative suggestion: increase the amount of hexcrete output.
> 
> `results = list(/datum/reagent/cement/hexement = 7 )` for example, which would be 1:1 with inputs (6 cement, 1 phenol)
> 
> Also, cement should probably output 8-10 units per mix too. Some water evaporates from cement mixes, but much of it actually cures and becomes part of the concrete, so mixing up a batch of concrete should also be close to 1:1 input to output. As it stands right now you get 1:0.5 total inputs to outputs.

I'll do that with regular cement, but making Hexacrete not take phenol is the main thing I want to do here

FalloutFalcon:
> use barrels to mix mass amounts of any chem

Breaking bad mixing vats coming 2006

thgvr:
ok

PR #5019, opened 2025-07-01
Fixes manifest and join menu UIs not updating properly

SapphicOverload:
:cl:
fix: Fixed manifest and join menu UIs not updating properly
fix: Fixed certain text characters breaking memos 
/:cl:


MarkSuckerberg (approved):
very awesome actually

PR #5022, opened 2025-07-02
Makes hoodies count as warm clothing

SapphicOverload:
## About The Pull Request

Hoodies now have minor cold protection.

## Why It's Good For The Game

It's what they do.

## Changelog

:cl:
fix: Hoodies are now warm
/:cl:


FalloutFalcon (changes requested):
Wintercoats, windbreakers, and other jackets all have the same value so that looks good(i think we need to rename this define tbh.)
however, shouldn't the hood also have the same value? (you could do this by making hoods grab the suits value or just hard code it)

FalloutFalcon (approved):

PR #5024, opened 2025-07-03
A little cover for some rain: Ponchos

thearbiber:
## About The Pull Request
Adds in a variety of recolorable ponchos, a neck item now located in your loadout slot.
image

## Why It's Good For The Game

more clothing options good, and who likes rain anyways?

## Changelog

:cl: TheArbiber, GeneralThrax
add: ponchos
/:cl:


code/modules/clothing/neck/_neck.dm
```diff
@@ -310,6 +310,219 @@
...
+/obj/item/clothing/neck/poncho/Initialize()
+	. = ..()
+	if(!base_icon_state)
+		base_icon_state = icon_state
+	make_hood()
+
+/obj/item/clothing/neck/poncho/Destroy()
+	. = ..()
+	qdel(hood)
+	hood = null
```
FalloutFalcon, line 361: really hurts to see you copy paste the hood code. 
I know ponchos are definitely an outer outer layer of clothing but the neck layer feels a bit misused here and its whats forcing you to do copy paste all this (tho someone could prob refactor this hood logic into like a element/component). 
Would like others thoughts on that tho as i'm unsure myself whether it should be suit or neck. neck definitely lets it fill a "niche" for people wanting to wear armor under it. (and therefore HIDE armor under it which is a worrying concept)

FalloutFalcon (commented):
what type of poncho are these meant to be? they read as more of a fabric/cloth one then anything particularly rain resistant, if i had to guess its just the color of the one you show, i think of much brighter colors when i think of ponchos.

icons/mob/clothing/neck/Ponchos.dmi
FalloutFalcon, line 1: snake_case for file names of any dream maker file type really (and 90% of other ones).

code/modules/clothing/neck/_neck.dm
generalthrax: Doing anything else was a fair bit out of my ability and I think it'd be Worse if it was just a suit item that fit on your neck

thgvr: btw its fine and can be deconverted later if someone turns it into a component or whatever

icons/mob/clothing/neck/Ponchos.dmi
generalthrax: Should be good

code/modules/clothing/neck/_neck.dm
FalloutFalcon: Does the sprite really convey that its a neck item? It goes down to there belt line. If we want it to be a neck item i would think it should sit on the shoulders more? Im thinking of the project zomboid poncho here that has that rolled up state.
Sorry this image is not super readable but best one i could find
Untitled

thearbiber: The whole point is that its a poncho, yeah. I dont see how hiding armor is an issue, just shiftclick if you're suspicious, and any big armor would instantly be noticeable because it covers the entire body. Not really interested in making it a suit slot. Ponchos are quite large to protect your body, but theyre also just a put over your head and be done with thing.

FalloutFalcon: > I dont see how hiding armor is an issue, just shiftclick if you're suspicious

My thought is that its sprite is very familiar to hoodies and since hoodies are suit slots and the sprite itself doesn't look like it would be a neck slot, it could definitely be a bit confusing. I would prob just assume its a hoodie if I hadn't seen the pr meaning i prob wouldn't see any thing to make me question it. Its not really a major balance concern saying yea only the lighter armors are really affected, that is interesting even, but its to note.

Imaginos16 (approved):
Looks good.

code/modules/clothing/neck/_neck.dm
rivercadaver: hiding armor visually might be a good side-effect. makes you potentially have to second-guess if the random indie chilling in the bar is a mark or not. in an active firefight you're left-clicking till spaceman goes horizontal anyways, and a inspect is a keybind away to see if someone really does have level 4 plates and a sawn-off under it

also as said before any kinda hardsuit will be easily visible underneath, and medium tac has arm/legplates which are a good tell. i think the confusion with hoodies will be pretty easily cleared up once the poncho has been commonplace for long enough, and with the way im wearing them they'll be there soon

FalloutFalcon (approved):

PR #5035, opened 2025-07-05
Fix typo in parameter name for MouseDrop_T function in pipe dispenser

Ossa88:
## About The Pull Request

Corrects the only instance of mob/usr in a proc line across the entire database.

Corrects Open Dream OD2000 Error from most recent Open Dream Build

## Why It's Good For The Game

Linter Pass is Good

## Changelog

:cl:
/:cl:

### DIFF
--- a/code/game/machinery/pipe/pipe_dispenser.dm
+++ b/code/game/machinery/pipe/pipe_dispenser.dm
@@ -144,13 +144,13 @@
 
 //Allow you to drag-drop disposal pipes and transit tubes into it
 /obj/machinery/pipedispenser/MouseDrop_T(obj/structure/pipe, mob/user)
-	if(usr.incapacitated())
+	if(user.incapacitated())
 		return
 
 	if(!istype(pipe, /obj/structure/disposalconstruct) && !istype(pipe, /obj/structure/c_transit_tube) && !istype(pipe, /obj/structure/c_transit_tube_pod))
 		return
 
-	if(get_dist(usr, src) > 1 || get_dist(src,pipe) > 1)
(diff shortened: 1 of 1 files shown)


code/game/machinery/pipe/pipe_dispenser.dm
```diff
@@ -143,7 +143,7 @@
 	return TRUE
 
 //Allow you to drag-drop disposal pipes and transit tubes into it
-/obj/machinery/pipedispenser/MouseDrop_T(obj/structure/pipe, mob/usr)
+/obj/machinery/pipedispenser/MouseDrop_T(obj/structure/pipe, mob/user)
 	if(usr.incapacitated())
```
FalloutFalcon: should stuff in this proc be using usr or user? I would assume user saying its bothering to pass it in?

Ossa88: ya know i completely forgot to actually update that... had it tested in my repo went it works ill just make a quick for ship... and went to bed... forgot to actually update the usr stuff let me do that real fast

Ossa88: Updated,

I only updated the instance within the MouseDrop_T proc... There is a to_chat within this same file that still uses "usr" even after defining "mob/user"... Which should be the case as not every instance of to_chat will be contained where a "mob/user" has been defined.

More than likely a much wider clean up might be done with proper rules when the "usr" define and the "mob/user" are* both instances are being used in the same proc.

FalloutFalcon (approved):

PR #4580, opened 2025-04-11
Sifnos-class Mining Cutter

Sadhorizon:
## About The Pull Request
MELD drilling ship. They go drill and stuff.

**TODO:**
- [ ] Desc

<details><summary>Screenshots:</summary>
<p>

obraz

StrongDMM-2025-04-11 21 03 19

StrongDMM-2025-04-11 21 03 23

</p>
</details> 

Roles:
- 1x Captain (more industrial outfit tho) (CM-23)
- 2x Drill Operator (2x CM-1, to be replaced with ECM-6)
- 1x Medical Technician
- 1x Mechanic
- 1x Crewman
obraz

Cool features:
- Upgraded drill
- Refinery
- Bodycams & a room to view them in

## Why It's Good For The Game
More CLIP ships - MELD is kinda cool too.

## Changelog

:cl:
add: Added Sifnos-class Mining Cutter - a CLIP-MELD drilling ship.
/:cl:

PR #5049, opened 2025-07-07
Sifnos-class Mining Cutter

Sadhorizon:
## About The Pull Request
Re-opening of #4580.

**TODO:**
- [ ] Add a MELD stamp when those are merged

<p>

scrnshot1

obraz

obraz

</p>
</details> 

Roles:
- 1x Captain
- 2x Drill Operator (2x ECM-6)
- 1x Medical Technician
- 1x Mechanic
- 1x Crewman
obraz

Features:
- Upgraded drill (T3 parts)
- Refinery
- Bodycams and a room to view them in

## Why It's Good For The Game
CLIP kind of needs more ships I think. This adds a new niche - and a new department - to their roster.

## Changelog

:cl:
add: Added Sifnos-class Mining Cutter - a CLIP-MELD drilling ship.
/:cl:


_maps/configs/minutemen_sifnos.json
```diff
@@ -0,0 +1,44 @@
...
+	"$schema": "https://raw.githubusercontent.com/shiptest-ss13/Shiptest/master/_maps/ship_config_schema.json",
+	"map_name": "Sifnos-class Mining Cutter",
+	"prefix": "CLSV",
+	"manufacturer": "Lanchester Shipyards",
+	"namelists": [
+		"GENERAL",
+		"CLIP MINUTEMEN",
+		"NATURAL"
+	],
+	"description": "The Sifnos-class is a mining vessel utilized by the MELD division of CLIP. Although the original prototypes were designed with moving freight in mind, ever since an ore smelting unit has been added its use shifted more into material gathering, which is its main purpose today. While the extraction methods are not as advanced as those seen on the Vela-class cruisers, utilizing primarily deep drilling technology, it ends up being more reliable due to lesser operational costs, as well as a smaller crew component. This makes it quite a common sight in the frontier.",
```
Erikafox:
```suggestion
	"description": "The Sifnos-class is a mining vessel utilized by the MELD division of CLIP. Although the original design was intended as a freighter, modern production runs have turned the Sifnos's purpose to mining minerals. The inclusion of a mineral processing array where there was once cargo racks allows on-site refinement, and the small crew compliment allows it to deploy into sectors of space a more specialized vessel (such as the Vela) would be unsuitable. This has made it quite a common sight in the frontier.",
```

code/modules/clothing/outfits/factions/clip.dm
```diff
@@ -186,6 +194,24 @@
 		/obj/item/clothing/mask/gas/clip=1
 		)
 
+/datum/outfit/job/clip/fieldengi
```
Erikafox: why subtype this rather than use industrial miner? Consider changing Industrial Miner to suit your needs

Erikafox (changes requested):
image
Remove the start point landmarks
Kh5uewASiq
Shutters should be placed over the pre-chargers rather than the engine. 
image
I would give this room a more technical spin. Aside from the techfloor on the very cramped generator and the telecommfloor under the relay, it doesn't feel quite like an engineering room. 
image
Ditto for this room, although I'm not entirely a fan of how it's practically two rooms in one. I feel like you could introduce a subdivision and make the refinery more akin to something like the Scarab. Since you have a paper on relinking the refinery present already, I don't think making a set up that won't auto-configure is out of the question.
Speaking of that paper, please add a notice board and have the paper spawn on it, like how the colossus functions. It's much more soulful than a piece of paper on a table.
image
Very minor, every other CLIP ship uses a lighter wood for the Captain's Quarters.
image
Look into making the bodycameras start pre-configured. Fairly sure it's possible.
image
I would not have the welding fuel start facing the cargo door. It's kinda a meme on the falmouth and it will be here too
image
The medic should have an explicit area set aside for treatment if they're going to have supplies. Even if it's just a foldout against a wall.
image
No minuteman stamp, they aren't involved here
image
Try shrinking the window, moving the shutter control to be directly adjacent to the windows, and adding a secondary control for the cargo bay shutters in here. It'd be good flavor for a master-controller set up.

The other major worry I have is how this will feel
image
Unsure if the sightlines that the "pillars" create will end up feeling good. This may be another good case for explicit subdivision.

Good work thus far.

Sadhorizon:
> Shutters should be placed over the pre-chargers rather than the engine.

This is based on how the Atlas did it. I'm happy to change it, but I figured this is like a CLIP standard or smth (I remember commenting about it on the Atlas PR and it being ignored so I figured it must be one). Same goes for the Arke actually.

> I don't think making a set up that won't auto-configure is out of the question.

I spent a bunch of time trying to make it work, unsuccessfully. I'm guessing because they are technically different rooms (and will be kept as such). Though maybe because there is a wall in-between. I honestly don't know, those things are confusing.

> The medic should have an explicit area set aside for treatment if they're going to have supplies. Even if it's just a foldout against a wall.

The medic is basically copied from the Arke. They can make their own spot there, they can make it here too.

> No minuteman stamp, they aren't involved here

Read the TODO list! It's a placeholder rn, will be swapped when the MELD one is merged.

Erikafox:
image
This doesn't satisfy the desire for a more technical area. Take some time to experiment with decals and tiles.
The other engine room is also going to need this.

>Shutters should be placed over the pre-chargers rather than the engine.

My thoughts on the matter are: How does a ship get thrust from a sealed off engine? It doesn't.

>The medic is basically copied from the Arke. They can make their own spot there, they can make it here too.

The arke is much more compact. You can get away with finding 2-4 tiles of space for the medic to use here.

>I spent a bunch of time trying to make it work, unsuccessfully. I'm guessing because they are technically different rooms (and will be kept as such). Though maybe because there is a wall in-between. I honestly don't know, those things are confusing.

Think you misunderstood what I said here. Since the set up isn't going to autolink, you have free reign to experiment with more interesting ones (that preferably don't bump against the engine room).

Sadhorizon:
> My thoughts on the matter are: How does a ship get thrust from a sealed off engine? It doesn't.

Still a standard set by other CLIP ships. I agree with the thought process but I much prefer stuff is in-line with one another.

> Think you misunderstood what I said here. Since the set up isn't going to autolink, you have free reign to experiment with more interesting ones (that preferably don't bump against the engine room).

Yeah, no. I like this setup a lot and am not planning to change it anytime soon. It's one of the more distinct areas on this ship and I do not see any issues with it. I'm not going to count issues with ancient code as one either.

> This doesn't satisfy the desire for a more technical area. Take some time to experiment with decals and tiles.

I also don't really see any issues here. It's overall a small room, it's detailed as is with catwalks, some depth near the engines and the tiles under machinery. I think some tiles can just be empty, without much clutter. Less is more sometimes.

Erikafox:
>Still a standard set by other CLIP ships. I agree with the thought process but I much prefer stuff is in-line with one another.

It's not universally followed, and is something we're trying to avoid.

Erikafox:
>Yeah, no. I like this setup a lot and am not planning to change it anytime soon. It's one of the more distinct areas on this ship and I do not see any issues with it. I'm not going to count issues with ancient code as one either.

If you aren't interested in addressing the reviews left then the ship won't be merged. Mind the bluntness but mapping is collaborative effort and the maptainers exist to help you create the best version of your map. Not to dig in and argue with you.

Sadhorizon:
I would really appreciate a review that can't be boiled down to "I don't like how this looks, please change it."
People have different styles and, let me tell you, it feels awfully fucking discouraging to get shut down like that.

I have adressed all the real gameplay concerns (minus the medic one, still thinking about the solution). I do not see any other issues.

Sadhorizon:
I suppose I am done, then. Kind of stupid but whatever. I won't compromise over opinions, I am done with doing so. I have felt like this is a consistent issue and one that is awfully shitty to deal with.

Maybe I'm onto something, or maybe it's just not for me anymore. I don't know.

Either way, it's been an honor mapping here. Thank you for all the time spent both reviewing my maps (with merit), and especially enjoying them in game.

PR #5053, opened 2025-07-09
add slowdown to crates + make duffelbags more available

Erikafox:
## About The Pull Request
as title

## Why It's Good For The Game

Duffelbags are supposed to be the quick to move storage solution, and I think this would encourage them being used in that position.

Also you should be able to buy satchels and such.
## Changelog

:cl:
add: you can now purchase duffelbags at da outpost
balance: pulling a crate now slows you down
/:cl:


Zevotech:
probably going to piss many people off but i think this is really good considering closets slow you down but crates somehow didnt. peak pr

PR #5057, opened 2025-07-11
Breach Mining

Erikafox:
## About The Pull Request
Changes how drilling works.

Drills now spawn "seismic disturbances" that spawn the monsters. They stay in place for the duration of the "wave". After a certain amount of waves you get a short break that is EXPLICITLY NOTIFIED

dreamseeker_euQbH5FbEw
dreamseeker_k03O0zdMKd

## Changelog

:cl:
add: Drills now spawn a spawner, which in turn spawns the monsters. It's an innovative system really.
fix: ACTUALLY FIXED DRILL TURF PLACEMENT THIS TIME
/:cl:

### DIFF
--- a/code/modules/mining/ore_veins.dm
+++ b/code/modules/mining/ore_veins.dm
@@ -112,7 +112,7 @@ GLOBAL_LIST_EMPTY(ore_veins)
 	currently_spawning = TRUE
 	START_PROCESSING(SSprocessing, src)
 
-/obj/structure/vein/proc/stop_spawning(currently_spawning)
+/obj/structure/vein/proc/stop_spawning()
 	if(currently_spawning)
 		currently_spawning = FALSE
 		STOP_PROCESSING(SSprocessing, src)
@@ -121,6 +121,8 @@ GLOBAL_LIST_EMPTY(ore_veins)
 	return TRUE
 
 /obj/structure/vein/process(seconds_per_tick)
+	if(!currently_spawning)
+		return
 	try_spawning_spawner()
 
 /obj/structure/vein/proc/try_spawning_spawner()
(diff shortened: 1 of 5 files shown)


FalloutFalcon:
Code looks good (i am very sleep deprived so remind me to look at this later again), feature is definitely an improvment, tho I do think the visuals are kinda iffy (mainly that the smoke particles dont really seem to fit with the breach at all) but your working with what we already have so its fine.

code/modules/mining/ore_veins.dm
```diff
@@ -90,9 +114,67 @@ GLOBAL_LIST_EMPTY(ore_veins)
...
+		if(length(active_spawners) >= vein_class)
+			return
+
+		var/turf/open/spawning_tile = pick_tile()
+
+		var/obj/effect/drill_spawner/bug_breach = new /obj/effect/drill_spawner(spawning_tile)
+		active_spawners += bug_breach
+		bug_breach.our_vein = src
+		bug_breach.AddComponent(spawner_type, mob_types, spawn_time, faction, spawn_text, max_mobs, spawn_sound, spawner_distance_min, spawner_distance_max)
+		bug_breach.start_death_timer(wave_length - 5 SECONDS)
```
FalloutFalcon, line 150: not in the scope of this pr, however i think the behavior could be made faster by keeping the vein as the real spawner but using the breaches as the location it spawns things in. It would prevent players from being overwhelmed by breaches with independent max_mobs and reduce the plenty of failing to destroy the holes which im unsure is the desired effect.

FalloutFalcon (approved):
Good improvement to the mechanic.

code/modules/mining/ore_veins.dm
Erikafox: the breaches lingering after the drill stops is intended, so is the potential to overwhelm.

PR #5072, opened 2025-07-15
Ballistic Shield Balancing

generalthrax:
## About The Pull Request

- Reduces block chance to 60
- Reduces integrity from 900 to 600
- Dramatically increases slowdown from 0.5 to 1.25
- Repair now requires 10 plasteel sheets and a lengthy do_after

## Why It's Good For The Game

During the short time its been merged Ive seen people use the shields to go Protagonist Mode on Far Greater numbers taking hilarious amounts of fire with little drawback. They were even repairing their shields while being fired upon

This should keep ballistic shields as a slow, defensive tool to act best as mobile cover instead of a station-esque desword 

## Changelog

:cl:
balance: Shields block slightly less, have less integrity, slow you down much more, and now require time and 10 plasteel to repaiar
/:cl:

### DIFF
--- a/code/game/objects/items/shields.dm
+++ b/code/game/objects/items/shields.dm
@@ -9,8 +9,8 @@
 	slot_flags = ITEM_SLOT_BACK
 	force = 10
 	item_flags = SLOWS_WHILE_IN_HAND
-	slowdown = 0.5
-	drag_slowdown = 0.5
+	slowdown = 1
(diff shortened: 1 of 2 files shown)


Ratvarr:
both shields still say to use metal to repair instead of plasteeel

generalthrax:
true

PR #5082, opened 2025-07-17
Remove mining skill and tweak mining

thgvr:
## About The Pull Request
Tones down the integrity damage of drills. For whatever reason it was set to pretty much always do 100 damage, with that being equal to a rock wall. This also made some tools do a shitton of damage to walls when they shouldn't like the jackhammers.

also deletes mining skill because. mining skill

## Changelog

:cl:
del: Removed mining skill and cape
balance: Pickaxes now take four swings to break rock walls
balance: Drills take three
balance: Jackhammers and drill damage on walls was toned down
balance: Angle grinders do less damage to walls
balance: Plasma cutters are less effective vs walls
/:cl:


code/modules/mining/equipment/angle_grinder.dm
```diff
@@ -28,7 +28,7 @@
 	usesound = 'sound/weapons/anglegrinder.ogg'
 	tool_behaviour = null // is set to TOOL_DECONSTRUCT once wielded
 	toolspeed = 0.6
-	wall_decon_damage = 350
+	wall_decon_damage = 200
```
Erikafox, line 31: this is gonna slow down salvaging and should prolly be in the cl too

code/modules/mining/equipment/mining_tools/drills.dm
```diff
@@ -0,0 +1,41 @@
...
+	item_state = "jackhammer"
+	usesound = 'sound/weapons/sonic_jackhammer.ogg'
+	hitsound = 'sound/weapons/sonic_jackhammer.ogg'
+	force = 20
+	armour_penetration = 15
+	wall_decon_damage = 35
+	demolition_mod = 1.2
+	attack_verb = list("blasted", "smashed", "slammed", "hammered")
+
+/obj/item/pickaxe/drill/jackhammer/old
```
Erikafox, line 34: didn't keep the knockback, intentional?

thgvr: yeah

LePiromano:
i got the concussive gauntlets from that one lavaland ruin the other day and they straight up didn't work, there a fix for that? (or just remove them ): )

Madtankdriver:
I do understand that some tools do a lot of damage to walls that shouldn't be able to but why the decrease of angle grinder wall damage and removal of just the mining skill?

I don't think many people want to wait around mining even more slowly or salvaging then already is. The rounds are about 3 to four hours? From experience, it takes about an hour or more to cut down a decently sized ruin. With this change it will be even longer. 

Salvaging should be a viable option to make money instead of just spamming drills to get diamonds.

thgvr:
mining skill is pointless slop, mining definitely needs to be slowed down regardless and having free Ore Vision after mining 10 rocks is ass. never really liked having any rock formation on planets or rocks leading into ruins being just trivial due to how easy it is to instantly break them.

rest of the changes we'll see, angle grinder seemed quite fast to me as well as the mech version

Madtankdriver:
Oh! Thats what you were worried about, the ore vision. Ok that makes sense.

Though, the mech saw would be pretty fast with how expensive it is and it being a really big saw lol.
Still thinking salvaging walls with saws shouldn't take forever but still speedy to get the job done :)

thgvr:
you also must consider that having walls so easily taken down by just about anything (sonic jackhammer. jesus) it can nullify a lot of ruin design. I'd like to have airlocks generally be easier to get through than walls

Madtankdriver:
Do you want airlocks to be easier to get into so that people would breach into them instead of breaking down walls with a silly tool like a knife that way or? Because I dig that. 
But when people are done with clearing ruins some people like on the Rhode class shuttle might would like to scrap the walls for valuable metals and plasma since they start out with a Ripley to use for scrapping.

Either way, you know what is best :D

Randomguy523:
Coming off a round with the testmerge earlier, nice to see the mining skill get nuked, the auto-scan especially was always kinda silly... But making it take multiple hits to mine stuff felt off, and also makes any mining _tediously_ slow. Mining is already going to be slowed down quite a bit with mining skill being removed, not sure if we need ~3-4x longer mining on top of that, ice mining in particular also gets hit hard by this.

As for the wall damage side of the PR, things felt fine and doesn't hinder people scrapping too much, so no complaints there.

FalloutFalcon (changes requested):
Just going off player feedback here, I think we shouldn't nerf basic mining for ores and transportation too hard in this pr. 
I think either reduce the nerf of mining implements
or
nerfing the intergirty of procgen rocks/buffing the integ of mapped rocks.
I definitely agree that rocks inside ruins are pretty much pointless rn but we dont want to make it a pain in the ass to actually get to the ruin itself.

PR #5083, opened 2025-07-17
stops baby goliath subtype from exploding

thgvr:
## About The Pull Request
stops baby goliath subtype from exploding. how was this not done already

## Changelog

:cl:
del: baby goliaths and gruboids no longer explode
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/mining_mobs/elites/goliath_broodmother.dm
+++ b/code/modules/mob/living/simple_animal/hostile/mining_mobs/elites/goliath_broodmother.dm
@@ -212,7 +212,6 @@
 /mob/living/simple_animal/hostile/asteroid/elite/broodmother_child/death()
 	. = ..()
 	visible_message(span_warning("[src] explodes!"))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/hostile/mining_mobs/elites/goliath_broodmother.dm
```diff
@@ -212,7 +212,6 @@
 /mob/living/simple_animal/hostile/asteroid/elite/broodmother_child/death()
 	. = ..()
 	visible_message(span_warning("[src] explodes!"))
-	explosion(get_turf(loc),0,0,0,flame_range = 3, adminlog = FALSE)
 	gib()
```
FalloutFalcon, line 215: flavor text doesn't really add up with it no longer exploding? what about making it a flame range of 1 so you still get the pop without setting people on fire (assuming its not on top of you)
that or just update the flavor text.

FalloutFalcon (changes requested):

PR #5096, opened 2025-07-19
more randomization for simple humans

FalloutFalcon:
## About The Pull Request
Adds both species and random seed randomization of simple humans 
image

## Why It's Good For The Game
No more 1 billion generic white male frontiersmen

## Changelog

:cl:
add: The appearance of simple humans is now much more varied
/:cl:


Maksoplo:
Finally we can now shoot the generic green Sarathi and the generic purple Elzu Frontiersmen.

Gristlebee:
I think non-humanoid body types should be temporarily disabled until sprites are in, cause we ended with weird cases like this where a vox spawns in an outfit where it doesnt have sprites for. (This is supposed to be vigi sec)
image

FalloutFalcon:
> I think non-humanoid body types..

Oh yea acctually very fair.

PR #5101, opened 2025-07-20
Remove purchasing concealed weapons bays

Erikafox:
## About The Pull Request

you can no longer purchase concealed weapons bays on the black market

## Changelog

:cl:
balance: you can no longer purchase concealed weapons bays on the black market
/:cl


FalloutFalcon:
can i get a why its good for the game?
fairly certain Ive complained about this exact item before being too common. However I think it makes more sense to just massively reduce the chance for it to appear. I think its a cool item but that we just see it too often rn mabye? Its also old antag cruft so its kinda whatever.
image

PR #5107, opened 2025-07-21
Girders act like barricades

Gristlebee:
## About The Pull Request

Girders act similarly to barricades like sandbags, allowing bullets to pass through if you're firing adjacent to it. They have a 50% chance (down from 80) to block bullets otherwise.

They have only a 20% chance to block bullets while unanchored however, since they're loose and are therefore worse cover.

## Why It's Good For The Game

I think this adds more depth to how you interact with the enviroment and destructible walls. A girder is now an opportunity to get a good angle once a wall goes down, rather than effectively a second obstacle you need to destroy to engage opponents. 

## Changelog

:cl:
add: Girders act like barricades
/:cl:

### DIFF
--- a/code/game/objects/structures/girders.dm
+++ b/code/game/objects/structures/girders.dm
@@ -6,6 +6,7 @@
 	density = TRUE
 	var/state = GIRDER_NORMAL
 	var/girderpasschance = 50 // percentage chance that a projectile passes through the girder.
+	var/unanchoredpasschance = 80 // provides worse cover while unanchored since it's loose and moving about
 	var/can_displace = TRUE //If the girder can be moved around by wrenching it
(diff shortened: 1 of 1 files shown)


Zevotech:
How does this work with displaced girders? I'd imagine we'd want to avoid people using them as mobile sandbags.

Gristlebee:
> How does this work with displaced girders? I'd imagine we'd want to avoid people using them as mobile sandbags.

As currently implemented,  they'd still work the same. It wont be fantastic cover, since it's still a 50/50 being shot through it.  I do see how this could potentially be problematic. If it ends up being an issue, I could make it more likely bullets to pass through and hit you if it's unanchored since it's loose and not secure, and would logically be worse cover.

Erikafox:
i would do that

PR #5109, opened 2025-07-22
Adds the Venetia-class Frontier Yacht

thearbiber:
## About The Pull Request
Noticed we were lacking "comfy big ship" vibe for a bit, and figured that one that can do gameplay shit would be fun. Starts with 1 captain, 1 2ic, 2 Inteq enforcers, 2 scrappers, and four passengers.
image
image
image

## Why It's Good For The Game
Think a more relaxing ship would be good for the game. This one shouldnt be sitting in port all day either, believe i gave it enough reason to go out and still Do Things while also having a mix of relaxing inside. The part inteq compliment should be interesting aswell.

## Changelog

:cl:
add: Venetia-class Superyacht
/:cl:


Erikafox (dismissed):
Are you attached to superyacht? The ship just doesn't feel big enough to me. It's more plain yacht sized. A different designation (I was thinking about Trabacalo, or maybe Frontier Yacht) might flow more naturally
<hr>

StrongDMM_AYOKUcHUrh
Empty radio closet
<hr>
image
image

weird perspective

<hr>

image
This is a very flammable area, put some fire alarms in the rooms, and an extinguisher. Captain's quarters probably needs a fire alarm too

<hr>

image
I would hate to be the enforcer with no chair. If it can take 2 enforcers their workspace should have space for both of them

<hr>

image
this spot can prolly use an extinguisher too

<hr>

image
This hallway could potentially be subdivided. 
image
You might expand this room into it and make it a proper crew lounge. I think the crew quarters could benefit from a proper luxury lounge, with the couch pit as the centerpiece. 
image
I tried some very lazy organizing to see if I could make it work. It kinda does? Feel free to ping me if you want to pursue changing it up because I'd be willing to work in parallel to make it work.
<hr>

image
Keep in mind that we're getting multiple cooking machines very soon. You'll want to make sure that the space can support it. Annexing part of the hall and adjacent dorm (not all the quarters need to be created equal...) would work as a solution for the space.

<hr>
image

Engineering feels like an afterthought in the design, especially with all the machinery behind the railing, the table at a weird corner, the grav gen snug against SSUs, and the crate (on its own is okay, but it feels like it wants a shelf)

If you intend to have the ship frequently use its (very large) airlock you should include an air recycling set up. 

Also needs fire alarm and extinguisher.

I think you'll get more out of tearing up the engineering space and starting from the top.
<hr>

image
You could turn the hallway space into a proper undressing / store your stuff area for the sauna, remove the direct inteq access from the hallway, and have a nice little room (again subdividing a larger area).

<hr>
image
I have a test for larger cargo bays, which is "if it has space for a vehicle does it comfortably fit it"
image
I think moving the equipment more to the sides, kinda like this
image
Would benefit that illusion. You might be able to use the space differently if you pursue the dorm scheme.

<hr>

image
I would make the fax and stuff a proper desk, with a chair and some space to work. Other KFP ships also use a little rougher pattern under electronic equipment.
<hr>

image
I would run this from engineering instead of the captain's quarters.

image
There should be a secondary set of buttons to operate the door somewhere in the ship.

thearbiber:
Should b good for test, updating images now

PR #5134, opened 2025-07-28
Fixes the volume of epipens so all reagents are added to a body upon use, and adds pens with chitosan to specialist medkits

zimon9:
## About The Pull Request

Epipens had chitosan added to them, but I think their volume wasn't increased to account for this addition, so chitosan never entered the bloodstream of an epipen user. This should increase the volume of epipens so that all of the reagents in the list spawn.

An emergency first-aid autoinjector pen was also added to both brute and burn specialist kits. Two hypovolemic-response autoinjector pens were added to oxygen deprivation kits. 

@generalthrax noticed that the volume that the epipen could hold was too low to hold the chitosan in the original pen, which was why increasing the volume added it to the reagent list for epipens. I wanted to thank them for noticing this for me and notifying me about it.

## Why It's Good For The Game

Chitosan didn't get added to the list of reagents inside of a body upon use because the volume of epipens were less than the total volume of all the reagents that are supposed to be in an epipen -- so this PR should resolve that issue. Having more medipens in specialist kits may also prove helpful in instances of heavy wounding.

## Changelog

:cl: GeneralThrax
add: added emergency first-aid autoinjectors and hypovolemic-response autoinjectors to specialist medical first-aid kits
fix: increased volume of epipens so that all listed reagents are spawned

/:cl:


code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -113,8 +113,8 @@
 	base_icon_state = "medipen"
 	lefthand_file = 'icons/mob/inhands/equipment/medical_lefthand.dmi'
 	righthand_file = 'icons/mob/inhands/equipment/medical_righthand.dmi'
-	amount_per_transfer_from_this = 13
-	volume = 13
+	amount_per_transfer_from_this = 15
+	volume = 16
```
generalthrax: there any reason this is 16 instead of 15?

zimon9: oh, oops, I forgot to change this-- thank you!

PR #5149, opened 2025-08-01
Adds Metal Improvised Splints

DIB-DOG:
## About The Pull Request
Adds the recipe for improvised splints using metal rods along with supporting sprites. The recipe uses 4 metal rods and 2 lengths of cloth.

image

## Why It's Good For The Game
It's within reason that you could make improvised splints out of metal rods as well as wooden planks, so this allows for that.

## Changelog

:cl:
add: Adds improvised metal splints
/:cl:


FalloutFalcon:
you gotta sync your local master but also when creating a branch, create it from the UPSTREAMS (this repo) master instead of your local, its just better. 
image

FalloutFalcon (approved):
This definitely seems like something that could be made into slap crafting or something similar rather then having two different recipes but this works for now.

DIB-DOG:
Gotcha gotcha, the guide I learned from always said to work from the local vs the upstream and it's worked since, but if that's the better way I'll adjust.

FalloutFalcon:
> Gotcha gotcha, the guide I learned from always said to work from the local vs the upstream and it's worked since, but if that's the better way I'll adjust.

you WORK on a local branch, but when you create that branch, you should create it from upstream so its most up to date.

FalloutFalcon:
otherwise you have to manually sync your local master with the upstream which is what it looks like you have been doing

DIB-DOG:
correct

PR #5151, opened 2025-08-01
Improves and Re-Adds Fax Spam Event

Ical92:
## About The Pull Request

Replaces Fax Spam papers with advertisement pamphlets with lore tidbits, as well as only faxing once for each fax machine.

## Why It's Good For The Game
More lore and flavor and improving an event

## Changelog

:cl:
add: Added improved fax spam event
/:cl:


FalloutFalcon (changes requested):
Can you move it to only occur on 1 fax machine/ship and improve the sprites to look like a pamphlet.
One of the original reasons it was removed was due to it often occurring multiple times in a uptime/round and feeling really repetitive.

Ical92:
> Can you move it to only occur on 1 fax machine/ship ... One of the original reasons it was removed was due to it often occurring multiple times in a uptime/round and feeling really repetitive.

It's only one pamphlet faxed once per fax machine, at most twice per round which I feel like is still a lot better than how it was before..

In regards to spriting I tried to but my codersprites did not meet my coderstandards . But I will try again

FalloutFalcon:
> In regards to spriting I tried to but my codersprites did not meet my coderstandards . But I will try again

I think we have ancient sprites used for awaymissions if you really need.

Ical92:
We do but those also do not meet my coderstandards™, plus I already have aseprite open

Ical92:
Also sighs I didn't even really intend them to be pamphlets I just could only come up with that word because of the ancient awaymission ones .. they're more like fliers

FalloutFalcon (changes requested):
Still need to make it specific to 1 machine or ship.
I would also go check in with the loreheads that your descriptions are good but otherwise looks fine!

FalloutFalcon (approved):
It only sends 1 flyer per ship so im ok with this now.

PR #5159, opened 2025-08-02
fix gelled bone wounds not processing

Erikafox:
## About The Pull Request

gel causes bones to heal slowly but it wasn't starting the process to actually do that

## Changelog

:cl:

fix: bone gel on its own now causes bone breaks to heal
/:cl:


thgvr:
oops

PR #5164, opened 2025-08-03
Lawnmower at compile time

FalloutFalcon:
## About The Pull Request
mapswitch define so we arnet setting pixel offsets at turf init
oh also fixes rocks being offset weird while im at it 
image

## Why It's Good For The Game
faster

## Changelog

:cl:
code: turf offset at compile time instead of init
/:cl:


Zevotech (approved):
yeag. better implementation of what I did.

PR #5169, opened 2025-08-05
Removes the chance for bulbs and light tubes to start out broken in ship and outpost areas

zimon9:
## About The Pull Request

Sometimes when a ship is spawned in, some of its lights start out of the gate being broken. Ultimately only a minor nuisance, but a nuisance nevertheless. This PR should toggle that behavior from select areas.

A flag has been added to areas labeled 'NO_RANDOM_LIGHT_BREAKAGE' which determines if an area will allow a light tube or a light bulb to randomly break during late initialization. This flag is added to ship areas and outpost areas by default.

## Why It's Good For The Game

While it does give some busywork, I feel like it may be nicer to have this behavior be gone. I'm open to other opinions on this, however.

The reason why the flag was added to both ship areas and outpost areas is under the idea that both of these areas are supposed to be relatively well maintained (as opposed to being dilapidated, like how a ruin may be)

## Changelog

:cl:
add: Added a flag to areas to toggle light breakage behavior during late initialization
/:cl:


Erikafox:
making it a flag on area might be good for keeping the behavior for ruins while discarding it for things we Dont Want It On.

zimon9:
> making it a flag on area might be good for keeping the behavior for ruins while discarding it for things we Dont Want It On.

Oh yeah, good idea

FalloutFalcon (changes requested):
move it to an area flag not a bool on the turf please.

zimon9:
> move it to an area flag not a bool on the turf please.

Oh, how do I do that? (I thought a bool was a flag in this case)

FalloutFalcon:
> Oh, how do I do that? (I thought a bool was a flag in this case)

image

zimon9:
> > Oh, how do I do that? (I thought a bool was a flag in this case)
> 
> image

Oh, my bad -- I've fixed it now
I tested it and it seems to be working fine, but let me know if it looks okay?

FalloutFalcon (approved):

PR #5188, opened 2025-08-08
Increases the junkie pill to ten unit per pill

Goat-Real:
## About The Pull Request

increases the amount of reagent per pill in the junkie's starting pill bottle to ten units (up from one)

## Why It's Good For The Game

one god damn unit between seven pills is not enough to satisfy addictions for any meaningful time. 70 units of meth/happiness/morphine/crank should be enough time for the junkie('s crew) to get the chems/ruin loot/stab someone in the outpost/buy it from someone because right now one unit per basically guarantees the junkie will be suffering from unavoidable withdrawl of a random chem, some of which have incredibly lethal withdrawls. 

## Changelog

:cl: Goat
balance: The reagent amount per pill in the roundstart junkie pill bottle has been increased (1 unit -> 10 units)
/:cl:

### DIFF
--- a/code/datums/traits/negative/addictions.dm
+++ b/code/datums/traits/negative/addictions.dm
@@ -30,7 +30,7 @@
 		for(var/i in 1 to 7)
 			var/obj/item/reagent_containers/pill/P = new(drug_instance)
 			P.icon_state = pill_state
-			P.reagents.add_reagent(reagent_type, 1)
+			P.reagents.add_reagent(reagent_type, 10)
 
 	var/obj/item/accessory_instance
 	if (accessory_type)


FalloutFalcon:
I know you are forced to have an addiction but 70u of meth still seems like alot for a -2 quirk.. I would mabye reduce the units to like 5 but add a few extra pills? I thought you only needed a single unit to reset withdraw symptoms anyway.

Goat-Real:
I can change it but, adding extra pills would either require a pill bottle subtype which has extra storage or just dumping loose pills somewhere. There's also a three other possible drugs which are way less 'beneficial' to be addicted to in this case but I can also just change meth to spawn with less pills/units if that's wanted since meth is easier to come across with survival pens.

I just have it set to ten cause I hate the chat spam of addiction messages and how fast withdrawals progress. So I kept it at ten so there's some sort of downtime at least between the withdrawals.

Apogee-dev:
if bottle size is the limiting factor here, couldn’t you give them a second pill bottle?

thgvr:
dont think this is a good change, a 10x increase isn't good here. if anything you should be looking at the addiction code instead of a bandaid like this.

Junkie only exists right now because I didn't want to refactor smoker, so your experience will be pretty shit with it no matter what as it's designed for a station environment

PR #5196, opened 2025-08-09
Add parent calls for atom_break and atom_destruction

blinkdog:
## About The Pull Request
This PR adds some calls to the parent functions `atom_break` and `atom_destruction` in subclass overrides.

## Why It's Good For The Game
Just cleaning up some code to remove linting errors.

## Changelog

:cl:
fix: Added calls to parent methods where expected
/:cl:

### DIFF
--- a/code/modules/power/multiz.dm
+++ b/code/modules/power/multiz.dm
@@ -62,9 +62,6 @@
 		icon_state = "cablerelay-broken"
 		broken_status = RELAY_ADD_CABLE
 
-/obj/machinery/power/deck_relay/atom_destruction()
-	return ..() //this shouldn't break under usual means
-
 /obj/machinery/power/deck_relay/Destroy()
 	break_connections()
 	return ..()
--- a/code/modules/power/multiz.dm
+++ b/code/modules/power/multiz.dm
@@ -62,6 +62,7 @@
 		broken_status = RELAY_ADD_CABLE
 
 /obj/machinery/power/deck_relay/atom_destruction()
+	. = ..()
 	return //this shouldn't break under usual means
 
 /obj/machinery/power/deck_relay/Destroy()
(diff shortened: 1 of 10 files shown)


code/modules/power/multiz.dm
```diff
@@ -62,6 +62,7 @@
 		broken_status = RELAY_ADD_CABLE
 
 /obj/machinery/power/deck_relay/atom_destruction()
+	. = ..()
 	return //this shouldn't break under usual means
```
FalloutFalcon: This seems to explicitly NOT want it to break but also I have no clue what a deck relay is.. If it wants to be indestructible add that flag. `if(resistance_flags & INDESTRUCTIBLE) //sod off`. This line also does nothing anyway if you add a parent call.
```suggestion
```

FalloutFalcon (changes requested):
Otherwise looks good.

code/modules/power/multiz.dm
blinkdog: I added the flag to the object, but kept the parent call to pacify the linter.
Give it another look and see if it looks OK?

code/modules/power/multiz.dm
```diff
@@ -62,7 +63,7 @@
 		broken_status = RELAY_ADD_CABLE
 
 /obj/machinery/power/deck_relay/atom_destruction()
-	return //this shouldn't break under usual means
+	return ..() //this shouldn't break under usual means
```
FalloutFalcon: My point is you can remove both lines here as return ..() does actually nothing.

blinkdog: Good call! I've updated the PR to remove those two lines.
Give it another look and see if it's ready for merge?

FalloutFalcon (approved):

PR #4456, opened 2025-03-22
Phorid and Outfit System QoL (2)

meemofcourse:
## About The Pull Request

re-pr of #4407, essentially a lite version of #3310

* When given an outfit, the game now checks if you're a phorid and gives you a phorid outfit that isn't tied to a fucking string check of your job name
  * NGR, SolGov, Hardliners and NT get their corresponding suits for phorids
* Phorids now start with a plasma tank in their pocket which is automatically turned on so you don't die when getting out of cryo 
* Emergency lockers no longer use RNG to determine what supplies they should have and now contain plasma belt tanks
  * Removes an unnecessary fire locker subtype
* The skirt preference during character creation now draws from a variable instead of just assuming [path]/skirt has a skirt (this will break a lot of things but i've done my best to lessen the impact)
  * This doesn't work with phorid outfits because they're not outfit/job (and if I did make them subtypes of outfit/job they'd loop infinitely)

imagen

## Why It's Good For The Game

i think people should account for the less fortunate in their crew. also better code and makes the skirt pref more flexible (does anyone even use it?)

## Changelog

:cl:
add: Emergency closets no longer determine what contents they should have through rolling dice
add: Phorids now have tanks roundstart, and should no longer have to scramble to get their emergency box as soon as they get out of cryo
code: Moved phorid outfitting to the actual outfit system
code: Outfits now draw from a variable when looking for skirt variants of clothing. Please report any weirdness.
/:cl:

PR #3310, opened 2024-08-26
[DNM] Phorids are now a pref option for species

meemofcourse:
## About The Pull Request

TODO:
- [x] Make it actually work
- [ ] Phorid-friendly map and gear changes
- [ ] Make sprite accessories account for phorid preference
- [x] Delete plasmaman species (ish)
- [ ] Species phorid sprites (p.s. i'm not a spriter)
- [ ] Plasmaman outfit sprites
- [ ] Figure out a less shitcode way for ignition to be handled
- [x] Outfit system updates
- [ ] Let admins/the game spawn phorids outside of a player spawn

imagen
imagen

* ~~Utterly butchers~~ Adapts the FBP option into a Species Variants preference, which allows you to choose between having a Full Body Prosthetic or being a Phorid. This essentially permits merging a species' traits and existing sprite framework with being a phorid.
* Makes the outfit system account for the fact that someone is a Phorid, and permit it to be customized so there's factional Phorid outfits.
* Closes #2639 and closes #2888

## Why It's Good For The Game
Plasmamen are really, really old, in dire need of a maintenance update to the point that they may be cut otherwise, and an interesting addition to the roster. This is, hopefully, enough to bring them up to speed.

## Changelog

:cl:
add: Phorids are now a preference for certain species
add: FBPs have IPC traits
del: Plasmaman species
code: Changed how FBPs are handled
/:cl:

PR #5203, opened 2025-08-10
Phorid and Outfit System QoL (3)

meemofcourse:
## About The Pull Request

re-pr of #4456, essentially a lite version of #3310

* When given an outfit, the game now checks if you're a phorid and gives you a phorid outfit that isn't tied to a fucking string check of your job name
  * NGR, SolGov, Hardliners and NT get their corresponding suits for phorids
* ~~Phorids now start with a plasma tank in their pocket which is automatically turned on so you don't die when getting out of cryo~~ 
   * Apparently this causes checks to fail and I cannot for the life of me figure out why
* Emergency lockers no longer use RNG to determine what supplies they should have and now contain plasma belt tanks
* Removes an unnecessary fire locker subtype
* The skirt preference during character creation now draws from a variable instead of just assuming [path]/skirt has a skirt (this will break a lot of things but i've done my best to lessen the impact)
  * This doesn't work with phorid outfits because they're not outfit/job

imagen

## Why It's Good For The Game

i think people should account for the less fortunate in their crew. also better code and makes the skirt pref more flexible (does anyone even use it?)

## Changelog

:cl:
add: Emergency closets no longer determine what contents they should have through rolling dice
add: Phorids now have tanks roundstart, and should no longer have to scramble to get their emergency box as soon as they get out of cryo
code: Moved phorid outfitting to the actual outfit system
code: Outfits now draw from a variable when looking for skirt variants of clothing. Please report any weirdness.
/:cl:


code/game/objects/items/tanks/tank_types.dm
```diff
@@ -98,7 +98,7 @@
 	volume = 6
 	w_class = WEIGHT_CLASS_SMALL //thanks i forgot this
 
-/obj/item/tank/internals/plasmaman/belt/full/populate_gas()
+/obj/item/tank/internals/plasmaman/belt/populate_gas()
```
FalloutFalcon, line 101: I think we can print plasmaman tanks? This might cause printed ones to start full.

meemofcourse: oh true

meemofcourse:
BWAAAAAAH

meemofcourse:
come the fuck on

PR #5204, opened 2025-08-10
Water Phase Transitions and Default Chemical Temperatures

Ical92:
## About The Pull Request
Adds phase transitions between ice↔water→steam, and adds default temperatures to chemicals for when they are first created (shouldn't include reactions) so everything containing ice doesn't immediately melt. Also changes stomachs sending mix_messages normally to instead grumble upsetly for harmful reactions and loudly for large reactions.

Also changes the reagent_recipe_collisions unit test to not collide for a cold reaction and a warm reaction with no lower bound set since the warm reaction can be done for any temperature range.

Since this changes add_reagent() it should definitely be TM'd to make sure nothing breaks... nothing should, but more testing never hurt.

## Why It's Good For The Game
More opportunities for chemistry with heating ice to react water with other chemicals, and boiling off water to filter mixtures. Also adds the ability for future chemicals to start at a specific temperature, ie. for pre-cooled cryoxadone.

Also makes reactions in stomachs slightly less clear, not having the stomach tell you what the mix_message is but just grumbling for certain reactions (waiting until after the chem cleanup to set which reactions since it shuffles them around)

## Changelog

:cl:
add: added phase transitions for ice and water
add: your stomach no longer tells you what's going on inside it, but it might still grumble!
/:cl:


meemofcourse:
when are we getting boiling water nuclear reactors

Ical92:
when you ~~code it~~ port it from goon

Ical92:
oughhhh

Ical92:
Evil and accursed

Ical92:
oh ooprs I thought I closed this before..

PR #5205, opened 2025-08-10
[VISV] Sparrowhawk-class Gunship

Apogee-dev:
## About The Pull Request
sparrowhawk_ingame
sparrowhawk
sparrowhawk_areas

Adds the Vigilitas Sparrowhawk-class Gunship, a 5-player ship meant to give Vigilitas characters a bit more focus than the single secoff slot they usually get.

Crew slots:
- Captain, a bottom-level Vigilitas manager. Pilot the ship. File all the paperwork. Lead your intrepid security officers into battle. Blow your own face off when your shots reflect off the hull.
- Security Officer x2. Beat up pirates. Beat up anarchists. Beat up yourself for wasting your life on a job that doesn't pay anything close to what it should.
- Paramedic. Put band-aids and gold star stickers on the secoffs when they inevitably get beat up. Smoke three packs a day. Ask yourself why you aren't paid as much as a real doctor when you're basically doing his job anyway.
- Engineer. Hit pipes with a wrench until they stop sounding funny. Hit the assistant with a wrench until he stops sounding funny.
- Assistant. Stock the kitchen. Clean up the ship. Realize that in practical terms you rank lower than the iguana.

Ship's armory contains 2 Commander pistols, 2 mini eguns, an energy shotgun, and an L-204 laser rifle. The captain also has a non-recharging antique laser. These will be replaced with the new Sharplite weapons when those are added.

## Why It's Good For The Game

Vigilitas needs some love, and NT & Friends need more diversity.

## Changelog

:cl:
add: VISV Sparrowhawk-class gunship
add: Captain variant of the Vigilitas HOS job
/:cl:


meemofcourse:
welcome back ranger

meemofcourse (changes requested):
small review

image

* nitpick but maybe we could start using nanotrasen_vigi_X or nanotrasen_ns_X for distinguishing between the different nanotrasen groups

image

* rotate the shutters on the cockpit so they look like this please

image

* there is a preset nanotrasen relay, do use that preset

image

* for things like the handcuffs in this ammo locker or the pepper spray refill, you might benefit from storing them all in the same place. maybe on a wall closet on the table with the rechargers

image

* i'd suggest splitting the flight deck windows and the ship windows

Apogee-dev:
Implemented everything from the review except for the note about the pepper spray refill and spare cuffs etc. The security lobby isn't a restricted area, and I didn't want the pepper spray dispenser out in the open where any member of the crew could get to it. Same for the spare cuffs, I wanted that to be something the captain distributes rather than something the officers can grab at will.

meemofcourse (approved):
looks good to me

thearbiber (changes requested):
image
 
I dont think we need to use mindshields for anything in modern day. That and the chameleon extras could be axed. Also believe we're moving towards getting suited up with prespawned clothes, so a "naked" version of this outfit would be good to spawn with. (uniform and necessities. Armor and all that on the ship to suit up in)

Apogee-dev:
The mindshield and chameleon bits are because this outfit is effectively just a copy of the regular nanotrasen HoS outfit. If we’re dropping those, I’d much rather we make it a concrete decision and do that all at once instead of leaving it inconsistent and only on newer outfits, otherwise things will just get more and more confusing and inconsistent as we go on.
as for the naked outfit, that’s fair enough. i’ll add an ungeared version tonight.

PR #5207, opened 2025-08-11
Allows working subshuttles to be spawned in ruins and the outpost.

Gristlebee:
## About The Pull Request

You can now map and spawn working subshuttles in ruins and outposts. Ruins can be mapped with normal subshuttle dock landmarks, but for outposts you will need to use /obj/effect/landmark/outpost/subshuttle_dock for them to work properly.

Adds a dock to NT Ice and touches up the Ancon subshuttle.

image
image

## Why It's Good For The Game

I think it's nice to have the option to have working subshuttles in ruins and outposts, and they should make good set pieces when needed.

## Changelog

:cl:
add: Subshuttle spawners can now be mapped onto ruins and outpost maps to spawn functional ships.
/:cl:

### DIFF
--- a/code/controllers/subsystem/overmap.dm
+++ b/code/controllers/subsystem/overmap.dm
@@ -702,6 +702,13 @@ SUBSYSTEM_DEF(overmap)
 		quaternary_dock.adjust_dock_for_landing = TRUE
 		docking_ports += quaternary_dock
 
+	else // we've spawned a ruin and are now checking for any docks that it has
+		for(var/obj/docking_port/stationary/port as obj in SSshuttle.stationary)
+			if(port.virtual_z() == vlevel.id)
+				if(port in docking_ports)
+					continue
+				docking_ports += port
+
 	var/list/datum/weakref/spawned_mission_pois = list()
 	for(var/obj/effect/landmark/mission_poi/mission_poi in SSmissions.unallocated_pois)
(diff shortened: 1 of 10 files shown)


Madtankdriver:
PEAK

thearbiber (approved):

code/controllers/subsystem/overmap.dm
```diff
@@ -702,6 +702,13 @@ SUBSYSTEM_DEF(overmap)
 		quaternary_dock.adjust_dock_for_landing = TRUE
 		docking_ports += quaternary_dock
 
+	else // we've spawned a ruin and are now checking for any docks that it has
+		for(var/obj/docking_port/stationary/port as obj in SSshuttle.stationary)
```
FalloutFalcon, line 706: what is the usecase for `as obj` here? Type defing the var should already filter out anything that is not an `/obj/docking_port/stationary`

FalloutFalcon (approved):
Otherwise looks good.

code/controllers/subsystem/overmap.dm
Gristlebee: I was going off how other docking code handled it as a guideline. I'm happy to change it.

Gristlebee: changed it, should be all good now

PR #5211, opened 2025-08-12
You can always place rods on hyperspace

Gristlebee:
## About The Pull Request

You can always place rods on hyperspace (transit) tiles rather than just within in shipbounds.
## Why It's Good For The Game

Being only to place rods on hyperspace within certain areas is a bit unintuitive and restrictive. It makes subshuttle boarding a slog for starters, since it basically restricts you to outpost elevator-like chokepoint combat, made even worse by the fact the "walls" in this case are kill tiles.

This should allow for much more freedom of movement and interesting decisions to be made.

## Changelog

:cl:
balance: You can always rod on hyperspace tiles.
/:cl:

### DIFF
--- a/code/game/turfs/open/space/transit.dm
+++ b/code/game/turfs/open/space/transit.dm
@@ -44,10 +44,6 @@
 /mob/living/throw_atom_into_space()
 	apply_damage_type(25, BRUTE)	// This may not seem like much, but if you toss someone out and they go through like four tiles, they're goners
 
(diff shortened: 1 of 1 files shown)


thgvr:
doesn't this let you build structures in hyperspace

Gristlebee:
> doesn't this let you build structures in hyperspace

Yeah, seems like you can, which is pretty problematic. Just going to close this and see if I can come up with a better idea

PR #5220, opened 2025-08-15
Less Waste Density

generalthrax:
## About The Pull Request

-Significantly decreases the wall density of metal and concrete cave biomes
-Reduced the amount of caves that generate
-Heavily reduces natural hivebot and hivebot fabricators across the board
-Removes the hivebot metal biome
-Lowers restricted gas amount to a fraction of the amount and limits it only to Sulfur Dioxide like salty sand worlds

## Why It's Good For The Game

Wastes are seldom visited except for wanting to put the crew into the trenches for little reward. They're so dangerous that it is often not worth the time to risk visiting the planet ruin.

Hivebot numbers were decreased because upwards of a hundred or more would often spawn on just one, dwarfing numbers of even the largest and most difficult of ruins. 

Looking to just make Wastes more walkable and able to be traversed without walking into 80 concrete walls, 300 hivebots, and an atmosphere composed of 100% chlorine gas

Old
image

New
image

## Changelog

:cl:
del: Removed hivebot metal biome from wastes
balance: Wastes are now significantly less cavey and have much less hivebots. Current cave biomes have less concrete and wall density
/:cl:

### DIFF
--- a/code/datums/atmosphere/planetary.dm
+++ b/code/datums/atmosphere/planetary.dm
@@ -90,11 +90,7 @@
 		GAS_CO=2,
 	)
 	restricted_gases = list(
-		GAS_CHLORINE=1,
-		GAS_AMMONIA=1,
-		GAS_SO2=3,
-		GAS_CO=2,
-		GAS_O3=1,
+		GAS_SO2=0.1,
 	)
 	restricted_chance = 10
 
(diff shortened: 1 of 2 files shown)


firebudgy:
You should post a before and after sample of terrain generation so people see what the major differences are visually

firebudgy:
I am stupid

Issue #5225, opened 2025-08-16
[BUG]mechanic hivebots can for some reason get to MEGAFAUNA LEVELS OF HEALTH.


Ratvarr:
## Testmerges: 
[#5088: 'Update ship_datum.dm' by Erikafox at commit 123eb157be](https://github.com/Shiptest-SS13/Shiptest/pull/5088)
[#5162: 'Chem Cleanup 2025' by Erikafox at commit 3f706d0971](https://github.com/Shiptest-SS13/Shiptest/pull/5162)
[#5105: 'Tortuga Touch Ups' by Gristlebee at commit 4a16e56249](https://github.com/Shiptest-SS13/Shiptest/pull/5105)
[#5087: '[Vigilitas Interstellar] Adds the Shrike-class Fast Response Frigate' by Burning02 at commit b7dde6791d](https://github.com/Shiptest-SS13/Shiptest/pull/5087)
[#5043: 'CLIP Ocean Outpost + adds static names for outposts' by thearbiber at commit 8d113a35d9](https://github.com/Shiptest-SS13/Shiptest/pull/5043)
[#5198: 'Lawnmower 2: Finishing the Job' by Zevotech at commit 3d12c73708](https://github.com/Shiptest-SS13/Shiptest/pull/5198)
[#5184: 'New Wasteplanet Ruin: Salvage Facility' by goober3 at commit ba6cd79f4e](https://github.com/Shiptest-SS13/Shiptest/pull/5184)
[#5212: 'Space ruins spawn with asteroid fields around them.' by Gristlebee at commit cf0ce4b84a](https://github.com/Shiptest-SS13/Shiptest/pull/5212)
[#5207: 'Allows working subshuttles to be spawned in ruins and the outpost.' by Gristlebee at commit e763d29de6](https://github.com/Shiptest-SS13/Shiptest/pull/5207)
[#5176: 'Kali Andhi Remaster' by rivercadaver at commit 1d5229a514](https://github.com/Shiptest-SS13/Shiptest/pull/5176)

## Reproduction:
feed a hivebot some scrap x350385938 times
it now has more health than bubblegum and a more effective gun than a mower and cannot be killed because it has 3600 health. THIS IS KIND OF BAD TO NOT HAVE A CAP FOR.

Erikafox:
resolved by #5230

PR #5230, opened 2025-08-18
Hivebots have a cap on how much they can grow.

Gristlebee:
## About The Pull Request

Addresses #5225.

Hivebots now cap out at 100 extra max health from scrap, and visually grow larger depending how they've eaten to communicate how screwed you are.

## Why It's Good For The Game

3600 HP hivebot is bad.

## Changelog

:cl:
balance: Hivebots have a cap on how they can grow from scrap.
/:cl:

### DIFF
--- a/code/modules/mob/living/basic/space_fauna/hivebot/hivebot_ai_behavior.dm
+++ b/code/modules/mob/living/basic/space_fauna/hivebot/hivebot_ai_behavior.dm
@@ -61,8 +61,6 @@
 /datum/ai_behavior/hunt_target/repair_machines/target_caught(mob/living/basic/hivebot/mechanic/hunter, obj/machinery/repair_target)
 	hunter.repair_machine(repair_target)
 
-/datum/ai_behavior/find_hunt_target/scrap_machines
-
 /datum/ai_behavior/find_hunt_target/scrap_machines/valid_dinner(mob/living/source, obj/structure/salvageable/yummers, radius)
 	if(length(yummers.salvageable_parts))
 		return TRUE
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/basic/space_fauna/hivebot/hivebot_ai_behavior.dm
```diff
@@ -61,8 +61,6 @@
 /datum/ai_behavior/hunt_target/repair_machines/target_caught(mob/living/basic/hivebot/mechanic/hunter, obj/machinery/repair_target)
 	hunter.repair_machine(repair_target)
 
-/datum/ai_behavior/find_hunt_target/scrap_machines
-
```
FalloutFalcon: nit: I think we should keep this, it makes it clear that it doesn't have a type def elsewhere.

FalloutFalcon (approved):
I really like the idea of massive mechanical beasts that grew from salvaging for decades, I think they could be implemented as a separate sub-type however.

Ratvarr:
> I really like the idea of massive mechanical beasts that grew from salvaging for decades, I think they could be implemented as a separate sub-type however.

as funny as it is, these things reached more health than megafauna and is already fires a riflegrade cartridge.
it can and will wipe even the best crews by ammo attrition alone.

code/modules/mob/living/basic/space_fauna/hivebot/hivebot_ai_behavior.dm
Gristlebee: I put it back, should be addressed now.

FalloutFalcon:
> as funny as it is, these things reached more health than megafauna and is already fires a riflegrade cartridge. it can and will wipe even the best crews by ammo attrition alone.

To be clear I would not have them as a mob that CONTINUES to grow but rather the implied flavor that they have grown for along time. Likely used a boss monster in mapping.

PR #5236, opened 2025-08-20
updated required byond server version across several files and refactored to remove some old stuff

Anticept:
Title. A subsequent pull request relies on this. 1644 has a major bugfix for in style loops.

Tested this and #5237 on 515.1647 and 516.1667

## Changelog

:cl:
server: Cleaned up __byond_version_compat.dm, minimum version is now 515.1644
server: updates .tgs.yml, dependencies.sh, and dockerfile to use 515.1647
/:cl:

### DIFF
--- a/.tgs.yml
+++ b/.tgs.yml
@@ -3,7 +3,7 @@
 version: 1
 # The BYOND version to use (kept in sync with dependencies.sh by the "TGS Test Suite" CI job)
 # Must be interpreted as a string, keep quoted
-byond: "515.1642"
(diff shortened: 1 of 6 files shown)


FalloutFalcon (approved):
Looks good, I agree with the change, however, i think the current tgs is 1642 (cant check rn, it might be down) so we will need to update that first (I cannot last I checked)

Anticept:
Mark has already been made aware of these changes, he knew they were coming.

PR #5239, opened 2025-08-21
Solarbundswaffenkammer makes their guns useful again

Ratvarr:
## About The Pull Request
most of it is in the changelog
I increased the solarian guncells to 15000 charge, and tweaked namely the GAR as said in the changelog to have a reasonable tradeoff of cell-to-shot ratio of 96 rounds standard, and 64 HC. I didn't tweak the other guns, because their energy consumption is fine as it is, with the increased full-charge state.
This isn't like. Gamebreaking because nobody plays Sol, but makes them actually able to shoot their own guns instead of looking like 2-bit gangsters with SKM's that actually work.

## Why It's Good For The Game
I get nobody plays SolCon but when you can't even shoot a full magazine out of your GAR I think it's a little silly.
Also you can buy everything properly and not get ferromagnetic HC lances ONLY

## Changelog

:cl:
add: Ferromagnetic Lances BACK into cargo.
balance: Solarian guns now have better rates of cell decrease, namely the GAR, which can fire 3 standard magazines, or 2 HC magazines.
balance: Ferromagnetic rounds now stack better and reasonably. Slugs: 10; Lance: 16; Pellets: 22, now you can load your guns actually with a reasonable amount of boxes!
fix: Solarian HC box sprites
/:cl:


Apogee-dev:
if you’re changing ammo count per box, remember to adjust prices in cargo accordingly so we dont get any weird cost effectiveness shenanigans like we had before the ammo price audit

Ratvarr:
so, i moved the ammo prices up a twing to make lances inline with 556CLIP, and just eyeballed pellets up by 50-ish since you get a whole 88 pellets

FalloutFalcon (approved):

rivercadaver:
now make the Claris good

PR #5242, opened 2025-08-21
Frag Grenade Adjustment + Removed Riot Launcher from Cargo

generalthrax:
## About The Pull Request

- Increased frag grenade cost to 500
- Slightly increased Tear gas and Flash bang cost to 200
- Removed riot grenade launcher underbarrel from cargo
- Removed heavy explosive damage and slightly reduced flame range from frags
- Removed devastation from Concussion grenades

## Why It's Good For The Game

Frags just seemed wayyyyyy too cheap for how effective they are for PVP (and to a lesser extent PVE). See them spammed out a loooot.

Feel like whatever balance grenades have is pretty much negated by the riot. You can just spit it out beyond normal throwing distance (and ludicrously far if you have it on a DMR or Sniper) with a quickie firemode toggle. Was considering simply locking it from DMRs and snipers but I feel like you can still get a pretty similar effect with just normal ADS. Might reconsider, would like to hear thoughts

As of 9/8: Wanted to lean grenades away from the direct explosive damage and more towards the shrapnel. Removing heavy explosive damage reduces the impact of cookoff and puts more emphasis on the shrapnel (which still does a lot of damage but a fair bit less so cause its tied to shrapnel + bullet armour instead of explosives and all the jank with that)

## Changelog

:cl:
del: Removed riot grenade underbarrel from cargo
balance: Increased frag grenade and other auxiliary grenade costs slightly
balance: Removed heavy explosive damage from Frags and devastation from Concussion grenades
/:cl:

### DIFF
--- a/code/game/objects/items/grenades/syndieminibomb.dm
+++ b/code/game/objects/items/grenades/syndieminibomb.dm
@@ -18,6 +18,7 @@
 	name = "HE Grenade"
 	desc = "A compact shrapnel grenade meant to devastate nearby organisms and cause some damage in the process. Pull pin and throw opposite direction."
 	icon_state = "concussion"
+	ex_dev = 0
 	ex_heavy = 2
 	ex_light = 3
 	ex_flame = 3
@@ -28,9 +29,9 @@
 	icon_state = "frag"
 	shrapnel_type = /obj/projectile/bullet/shrapnel
 	shrapnel_radius = 4
-	ex_heavy = 1
+	ex_heavy = 0
 	ex_light = 3
-	ex_flame = 4
+	ex_flame = 1
 
 /obj/item/grenade/frag/mega
 	name = "FRAG grenade"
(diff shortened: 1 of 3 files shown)


DIB-DOG:
I think I agree that frags could cost more but I also feel like a jump from 250 to 750 is a bit steep. 

I see frags being used but I wouldn’t say they’re being abused. By raising their price, people are going to become less and less likely to use them, not simply because of the price of acquiring it, but because knowing that it’s worth 750 dollars and once you throw it, it’s gone, is going to make people extremely hesitant to even throw it at all. Think expensive potions in an rpg “what if I need it for later?”.

increasing the price to 750 I think is going to make them completely non-viable for any situation other than PvP, which detracts from some fun in using them.


Rounding back, I do think they could cost more still, maybe 500?

thearbiber:
Believe 750 is a fine price for something that is capable of near instakilling an entire group of people, while also being able to delimb easy. Even against simplemobs, you toss one in a room and by all accounts at least one or two are dying. I think people being hesistant to use something that strong is probably for the better, because the only other routes are probably full removal or nerfing them slightly somehow.

DIB-DOG:
I’d argue people can make much stronger bombs than the purchasable frags for a much lower price quite easily. The only thing holding most people from doing that is a matter of convenience.

FalloutFalcon (approved):

generalthrax:
Made some small changes, notably removing heavy damage from the Frags and bringing the price down to 500

PR #5246, opened 2025-08-22
fixes hermits punching a hole through ship hull

Erikafox:
## About The Pull Request
basic mobs are now checked for inside isanimal + human simplemobs will not drop loot if they are currently qdelling

## Changelog

:cl:
fix: animals should stop erupting into gore fountains when you land
fix: hermits will refrain from causing damage to your hull during landings
/:cl:

### DIFF
--- a/code/__DEFINES/is_helpers.dm
+++ b/code/__DEFINES/is_helpers.dm
@@ -115,7 +115,7 @@ GLOBAL_LIST_INIT(turfs_without_ground, typecacheof(list(
 #define isbasicmob(A) (istype(A, /mob/living/basic))
 
 //Simple animals
-#define isanimal(A) (istype(A, /mob/living/simple_animal))
+#define isanimal(A) (istype(A, /mob/living/simple_animal) || istype(A, /mob/living/basic))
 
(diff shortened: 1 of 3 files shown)


code/__DEFINES/is_helpers.dm
```diff
@@ -115,7 +115,7 @@ GLOBAL_LIST_INIT(turfs_without_ground, typecacheof(list(
 #define isbasicmob(A) (istype(A, /mob/living/basic))
 
 //Simple animals
-#define isanimal(A) (istype(A, /mob/living/simple_animal))
+#define isanimal(A) (istype(A, /mob/living/simple_animal) || istype(A, /mob/living/basic))
```
FalloutFalcon: this will cause some runtime from things using this as a check before type casting them as simple animals

FalloutFalcon: this would still cause runtimes and i think your other change in this pr fixes it by itself?
```suggestion
#define isanimal(A) (istype(A, /mob/living/simple_animal))
```

Erikafox: thought i saved removing that

PR #5256, opened 2025-08-24
Disables Prayers

firebudgy:
## About The Pull Request

Disables the Prayer verb for use.

Saycode scares me and there's weird other stuff interlinked so disabling it seemed like the best option.

## Why It's Good For The Game

Admins don't read Prayers, they're not how you're supposed to bring up issues that need admin intervention. Chaplains have ceased to exist entirely (usually the main reason for Prayers to even be used), and most admins can attest that with no sound notification and no easily spottable special text that they are incredibly hard to even notice and respond to.

The nature of them being "IC" requests also makes it very strange to respond to them in general.

This disables them entirely so people who have actual requests will take them through the proper channels rather than using a verb that nobody will read, and makes no IC interactions at all for any observers.

## Changelog

:cl:
balance: You keep screaming but God won't hear your Prayers.
/:cl:

### DIFF
--- a/code/modules/admin/verbs/pray.dm
+++ b/code/modules/admin/verbs/pray.dm
@@ -1,6 +1,6 @@
 /mob/verb/pray(msg as text)
-	set category = "IC"
 	set name = "Pray"
+	set hidden = TRUE
 
 	if(GLOB.say_disabled)	//This is here to try to identify lag problems
 		to_chat(usr, span_danger("Speech is currently admin-disabled."), confidential = TRUE)


Gristlebee:
My take on this issueis that this a cultural one surrounding player's hesitance around ahelping rather than a code one. I understand your perspective on this, but I do not a think code solution is ultimately the route here.  I'd personally think it'd be more effective to update the rules and associated pages with proper ahelp/prayer conduct, or potentially change the name of the prayer verb to something more concrete or grounded in our setting and see if it changes how people approach it.

FalloutFalcon:
I used it for my SRM fairly regualry tho I did rarely get a response back from what I can remember right now.

FalloutFalcon:
> or potentially change the name

`Wish` could work well here, would help justify praying if your character is not religious.

Erikafox (changes requested):
You gotta remove it properly

PR #5259, opened 2025-08-25
Adds a new beach ruin, frontiersman moat, and removes two beach ruins aswell

thearbiber:
## About The Pull Request
As stated. Adds this new beach ruin and also nerfs treasure troves assault rifle to an SMG. Notable loot would be an f3 drop, dynamic mission, cm23, some tools and a defib.
<details><summary>Images</summary>
<p>
image

image

image

</p>
</details> 

## Why It's Good For The Game

one, beaches need more modern ruins. Two, engi and pirate crash were pretty bad. They had nothing going on really. Also, the assault rifle on treasure trove was slightly absurd.

## Changelog

:cl:
add: Frontiersman Moat ruin
del: Crashed Engineer, Beach Pirate Crash
balance: assault rifle on treasure trove replaced with an SMG
/:cl:


Gristlebee (dismissed):
- I'd extend the patio here out by one tile so there's walking space to get to the couch rather than having to need to walk atop it. I don't think the curve on the couch works either.
image

- I think it'd be fun to put some personal effects in these closets to personalize them and give a little more to look at.
image

- I think adding some non-grey sidings, or mixing up the tile use could help make the interior of the bunker here pop more. Catwalks running with the wires may be nice as a suggestion.
image

- I think some minor gun wielding enemies around the table could make this feel a bit more lived in, and make crossing the moat from the north have a token resistance.
image

- I'd like to see some more fanfare or implication that this guy is special. There's nothing to telegraph or imply that this person is special for having a 100% drop.
image

- Some minor traps on the less defended bridges, or obvious landing points from the moat could be a nice touch.
image

- This corpse on the beach head has an outfit with extra gear on it. I think it would be more fitting for it to be inside the bunker in the corpse dump, and have one of the basic clip corpses be there instead.
image

- Having sentry variants that don't chase for illestren guards would be good so they don't run off from their posts on aggro.

PR #5261, opened 2025-08-26
Fixes gun maint kit having unlimited uses

### LINKED ISSUE #5260: [BUG] Firearm Maintenance Kit can be used an infinite number of times, rather than just 5
## Reproduction:
  Use the firearm maintenance kit on multiple weapons at the same time, such that the kit reaches "0 uses" while a do_after is still active on a second weapon. This will cause the kit to reach "-1 uses" after which it can be used infinitely.

meemofcourse:
## About The Pull Request

image

* Changes the if(!uses) to if(uses <= 0)
* Closes #5260 

## Why It's Good For The Game

capitalism

## Changelog

:cl:
fix: Gun maintenance kits no longer have unlimited uses
/:cl:

### DIFF
--- a/code/game/objects/items/gun_maint_kit.dm
+++ b/code/game/objects/items/gun_maint_kit.dm
@@ -18,7 +18,7 @@
 /obj/item/gun_maint_kit/afterattack(atom/target, mob/user, proximity)
 	if(!proximity)
 		return
-	if(!uses)
+	if(uses <= 0)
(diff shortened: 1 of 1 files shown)


FalloutFalcon (approved):

FalloutFalcon:
I was gonna request that it qdels itself but since it has custom materials, you can scrap it, which is cute.

PR #5294, opened 2025-09-08
Allows one to see how many hours are left in an outpost mission, and extends mission durations by 50%

zimon9:
## About The Pull Request

This PR adds functionality to see how many hours are left to complete a contract, and not just the modulo of the minutes remaining. This PR also extends outpost mission timers across the board by 50%.

image

## Why It's Good For The Game

As they are now, the current timers are perfectly sufficient for crews operating adequately. However, I've noticed that, sometimes, some crews can run into some bad luck and they can all end up dying. In the time between their distress call being answered, and them returning to the outpost, they may find that, rubbing salt into the wound, their mission timers have run out. This PR aims to address this, and allow more time in the event of accidents, events, and extended roleplay.

## Changelog

:cl:
add: Added 50% more time to outpost mission contracts
fix: Fixed mission timers not showing how many hours are left in a contract
/:cl:


code/modules/missions/mission.dm
```diff
@@ -97,7 +97,7 @@
 	if(duration)
 		var/old_dur = duration
 		var/dur_mod = duration * dur_mod_range
-		duration = round(rand(duration-dur_mod, duration+dur_mod), 30 SECONDS)
+		duration = round((rand(duration-dur_mod, duration+dur_mod) * 1.5), 30 SECONDS)
```
FalloutFalcon: could you just go through and manually adjust the numbers up? this seems pretty band-aid.

code/modules/missions/outpost/_outpost.dm
```diff
@@ -60,7 +60,7 @@
 		"value" = src.value,
 		"duration" = src.duration,
 		"remaining" = time_remaining,
-		"timeStr" = time2text(time_remaining, "mm:ss"),
+		"timeStr" = num2text(text2num(time2text(time_remaining, "hh")) - 20) + ":" + time2text(time_remaining, "mm:ss"),
```
FalloutFalcon: any reason you cant just do
```suggestion
		"timeStr" = time2text(time_remaining, "hh:mm:ss"),
```

FalloutFalcon (changes requested):

code/modules/missions/outpost/_outpost.dm
zimon9: Ah, well, when I do that, it gives the time formatted as something like "20:03:45" or "21:05:10", which was incorrect -- this was the only solution I could find for having the output actually match the correct duration.

code/modules/missions/mission.dm
zimon9: Yeah, you're right -- I've manually increased the numbers for each of the mission durations now by 50%

code/modules/missions/outpost/_outpost.dm
FalloutFalcon: But when it is showing the module of minutes its still accurate?

zimon9: Yeah -- let me paste a screenshot of what I mean
image

zimon9: I ran down a few in order to verify, and it looked like the modulo of the minutes and seconds was correct, but the hours wasn't -- It was kind of a hacky solution, but I opted to remove 20 from the hours output of time2text manually in order to correct for the discrepancy, due to that.

FalloutFalcon: ugh. insanely hacky but fine for now since i cant spot an obv reason whats causing it.

FalloutFalcon (approved):

code/modules/missions/mission.dm
```diff
@@ -97,7 +97,7 @@
 	if(duration)
 		var/old_dur = duration
 		var/dur_mod = duration * dur_mod_range
-		duration = round(rand(duration-dur_mod, duration+dur_mod), 30 SECONDS)
+		duration = round((rand(duration-dur_mod, duration+dur_mod)), 30 SECONDS)
```
FalloutFalcon:
```suggestion
		duration = round(rand(duration-dur_mod, duration+dur_mod), 30 SECONDS)
```

zimon9: My bad! Thanks for catching that

code/modules/missions/outpost/_outpost.dm
zimon9: yeaaa it's really really weird -- I was trying to figure out why it was happening for a couple of hours, but I couldn't see why

zimon9:
image

On the main server, it looks like this 20 hour bias isn't included, so I've removed the code accounting for that, and that should hopefully fix it.

There's a better solution that'd work in a local server and the main server itching at the back of my head, but I might have to dig around to find what it might be.

FalloutFalcon:
it might be jank from how we calculate time. could be related to timezone stuff.

PR #5297, opened 2025-09-09
Enhances the visibility of the helm's lock status

zimon9:
## About The Pull Request

This PR adds feedback into the chat when a keyfob is pressed, in the form of a message in the chat, and an audible sound cue. It also displays the status of the helm lock when examined -- because presumably a device that could toggle the status of a ship's helm remotely would feature an LED to indicate this status.

image

## Why It's Good For The Game

Ship keys have the ability to toggle a helm's status remotely, but it doesn't give one feedback in the chat when it's done. These changes should hopefully make it clearer.

## Changelog

:cl:
add: Added feedback upon keyfob toggle
/:cl:

### DIFF
--- a/code/modules/overmap/ships/controlled_ship_datum.dm
+++ b/code/modules/overmap/ships/controlled_ship_datum.dm
@@ -496,6 +496,7 @@
 	else
 		helm_locked = !helm_locked
 		playsound(src, helm_locked ? 'sound/machines/button4.ogg' : 'sound/machines/button3.ogg')
+		user.visible_message(span_notice("[user] presses the keyfob on [shipkey]."), helm_locked ? span_notice("You press the keyfob on [shipkey], locking the helm.") : span_notice("You press the keyfob on [shipkey], unlocking the helm."))
 
 	for(var/obj/machinery/computer/helm/helm as anything in helms)
 		SStgui.close_uis(helm)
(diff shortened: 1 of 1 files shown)


Erikafox (changes requested):
a locked computer could prolly use an icon state rather than having this be hidden in the examine

thgvr:
should be on examine of the helm not the key

zimon9:
Setting this to draft while finishing up the helm icon sprites

zimon9:
> should be on examine of the helm not the key

You're right, I should make it an examine on the helm -- though if the status were not visible on the key as well, you'd need to toggle the helm's status in order see what the status is, which might be a little bit extra interaction than may be required for this? Of course, the text indicating how the helm's status has been modified can be removed from keyfobs as well, but then they'd be no different from current keyfobs, and my intention was to give users a way to quickly and easily check the status of their helm without having to go back to the ship itself.

thgvr:
what I would do is have being locked change the screen's icon on the helm computer, examine too sure. it's a little too easy and abit odd if the keys themselves say the status

zimon9:
Oh, alright -- I'll remove the function in keyfobs for seeing that status, then. I'm currently working on sprites for the helm but the sidesprites are proving.. difficult.

This is what the frontstate looks like so far: 
Aseprite_0BYx0h3woC

Thoughts?

PR #4629, opened 2025-04-16
Overmap Expansion #5.02 Fixes Small ship docking ports

rye-rice:
## About The Pull Request
Fixes the docks of smaller ships to have proper mobile docking ports, therefore allowing ship to ship docking. In theory.
Also standardizes mobile docking ports being at the airlock except in circimstances where it physically cannot

## Why It's Good For The Game

image
the vaqueors docking port is finally not in a stupid location
docking ports working
image

## Changelog

:cl:
add: Most small ships should have more stationary docking ports
fix: Mobile docking ports on most small ships
fix: disable_on_owner_ship_dock should work more properly now
/:cl:

PR #5298, opened 2025-09-09
Slightly Increases the Length of the Transit Area Border

generalthrax:
## About The Pull Request

Increases the base Transit Area size from 12 to 18

Probably should get looked at I'm not super sure if this has Unforeseen Consequences attached

image

## Why It's Good For The Game

For some reason ships are placed in the top right corner of the transit area instead of the center. Don't really know how to put it in the center or if it Should be in the center. This pretty much just gives more space to north-facing subshuttle ports so they can be boarded by longer subshuttles like the Flea or the Kunai.

## Changelog

:cl:
add: Slightly increased Transit Borders so subshuttles can dock better to north-facing subshuttle ports
/:cl:


rye-rice:
This is redundant by #4629
Use `"tranist_x_offset": and `"tranist_y_offset":` in the json to fine tune the location of the ships in transit

PR #5302, opened 2025-09-09
Adds claymores to the outpost export market, and removes them from outpost cargo

zimon9:
## About The Pull Request

This PR adds claymores as a possible export at the outpost market, at 1250 credits each as the base price, and it removes the option to purchase claymores on the market.

## Why It's Good For The Game

No reward without a little risk, right? I remember seeing someone taking the time to carefully defuse a few in order to sell them, only for them to wind up disappointed because they weren't actually a valid export. It makes sense for CLIP to want to get these removed from the sector as much as landmines, I believe, so this PR aims to add that option.

Claymores as they currently are cost too much to be useful for a defensive emplacement, and they're also just a little too destructive when they _do_ catch someone, oftentimes instantly killing someone and shredding much of their equipment.

## Changelog

:cl:
add: Added claymore exports
del: Removed claymore purchases
/:cl:

### DIFF
--- a/code/modules/cargo/exports/scraping.dm
+++ b/code/modules/cargo/exports/scraping.dm
@@ -5,6 +5,12 @@
 	elasticity_coeff = 0.1
 	export_types = list(/obj/item/mine/pressure/explosive)
 
+/datum/export/claymore
+	unit_name = "defused claymores"
+	desc = "The Confederated League maintains an active bounty program for the disposal of UXO. Make the frontier a safer place today!"
+	cost = 750
+	export_types = list(/obj/item/mine/directional/claymore)
+
 /datum/export/anomaly
 	unit_name = "stabilized anomaly core"
 	cost = 3000
--- a/code/modules/cargo/exports/scraping.dm
+++ b/code/modules/cargo/exports/scraping.dm
@@ -8,7 +8,7 @@
 /datum/export/claymore
 	unit_name = "defused claymores"
 	desc = "The Confederated League maintains an active bounty program for the disposal of UXO. Make the frontier a safer place today!"
-	cost = 750
+	cost = 1250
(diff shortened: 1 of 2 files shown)


code/modules/cargo/packs/sec_supply.dm
```diff
@@ -216,14 +216,6 @@
...
 
-/datum/supply_pack/sec_supply/claymore
-	name = "C-10 Claymore Crate"
-	desc = "Contains one motion-activated directional mine, perfect for ambushing enemy infantry. Still debatably legal to sell!"
-	cost = 750
-	contains = list(/obj/item/paper/fluff/claymore,
-					/obj/item/mine/directional/claymore)
-	crate_name = "C-10 Claymore crate"
-
 /obj/item/paper/fluff/claymore
```
FalloutFalcon: would want to move this paper item elsewhere

FalloutFalcon (changes requested):

code/modules/cargo/packs/sec_supply.dm
zimon9: that's a good point -- I'm not sure where would've been a good spot for it, so I opted to just remove it entirely

FalloutFalcon (approved):

code/modules/cargo/exports/scraping.dm
```diff
@@ -5,6 +5,12 @@
 	elasticity_coeff = 0.1
```
FalloutFalcon, line 5: was it intentional to have the price fall off 10x slower then landmines (since the default for claymores would be 0.01)

zimon9: Oh, my bad, let me fix that

zimon9: Sorry about that, it should be fixed now

PR #5304, opened 2025-09-09
Add several new adjectives

Retrino:
## About The Pull Request

Adds a handful of new adjectives for character setup. Listing them all would be kind of a word jumble nightmare, so please refer to the file <3 Some examples:
Ambitious, Classy, Impulsive, Lackadaisical, Mischievous, Rugged, Whimsical

## Why It's Good For The Game

Having more to choose from allows for people to be more unique with their prefix, making them a bit easier to spot in a crowd outside of appearance alone (which is very hard to do with uniforms). Also just fun

## Changelog

:cl:
add: Adds some new adjectives to character setup
/:cl:


Erikafox (changes requested):
Adjectives should be something that you can tell at a glance, preferably a physical descriptor, please go through and try to match that.

My litmus is "would I describe someone I don't know with this term"

Retrino:
> Adjectives should be something that you can tell at a glance, preferably a physical descriptor, please go through and try to match that.

That's sooorta what I tried to vibe check while going through them, but some were harder to pin down than others for what could be felt out by first impressions. I tried to use the original list as a guide, though I think a lot of adjectives really just boil down to opinion. Going to remove some, let me know if there's more specific ones that don't mesh

FalloutFalcon:
Sorry we discussed it and even still most of these dont fit with what we actually want adjectives to be. Someone is gonna actually cut some of the existing ones, once they do that I would recommend looking at what they cut and seeing if you could make another pr based on that info.

PR #5305, opened 2025-09-09
Replaces Moth's Surplus Prosthetics with Humaniform

generalthrax:
## About The Pull Request

Says on the tin. Replaces the generic surplus moth prosthetic with humaniform (except for the head it looks cursed otherwise)

## Why It's Good For The Game

Before you say "humaniform is only for humans", take a look at this

image

You can't tell me _that_ looks better than this.

image

There's also the point that moths are human descendants, along with the original point that the generic looks like shit.

## Changelog

:cl:
add: Moths now use humaniform as their prosthetic limb type (except for the head)
/:cl:


FalloutFalcon:
rachnids might deserve the same treatment?
image
image

generalthrax:
Don't want to genericize it as the "fallback" prosthetic. Just thought it'd be a cute detail to be on moths too cause of being descendants of humans.

PR #5306, opened 2025-09-09
bombed starport wind fix

thearbiber:
## About The Pull Request
This should fix all wind on bombed starport

## Why It's Good For The Game

is that a tornado or are you happ-AAAAAAAAAAAAAHHHHHHHHHHH

## Changelog

:cl:
fix: wind on bombed starport
/:cl:

### DIFF
--- a/code/game/turfs/open/floor/plating/jungle.dm
+++ b/code/game/turfs/open/floor/plating/jungle.dm
@@ -53,6 +53,8 @@
 	light_range = 2
 	light_power = 1
 
+//for bombed spartport
+
 /turf/open/floor/carpet/red/jungleplanet
 	planetary_atmos = TRUE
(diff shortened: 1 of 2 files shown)


code/game/turfs/open/floor/plating/jungle.dm
```diff
@@ -53,6 +53,31 @@
 	light_range = 2
 	light_power = 1
 
+/turf/open/floor/carpet/red/jungleplanet
```
Zevotech, line 58: oughhhhhh i wish we didnt have to do this. planetary turf refactor never. as long as it works i suppose.

Erikafox (approved):
thanks cheese boy

Zevotech (approved):
looks good. only suggestion is to leave a comment above the added turfs in jungle.dm listing the maps they are used on for posterity

PR #5314, opened 2025-09-12
Gas mask resprite+sec hailer resprite/reflavoring 

thearbiber:
## About The Pull Request
Sec hailer is reflavored to a generic half mask, also adds it to loadout so its not literally nowhere.
Species variants include kepi, sarathi and human for the half mask and kepi and human for the full mask (I had 0 ideas for the sarathi mask)
image

## Why It's Good For The Game
The old sprites were pretty ass. These should be nicer, and as for the half mask in loadout, its literally just a reflavored breath mask. Gas masks as they are now are only valuable with eye protection, which this has none.

## Changelog

:cl:
add: Gas mask resprite, sec hailer resprite and reflavoring
add: half mask to loadout
/:cl:


Apogee-dev:
My only concern is with the all-red color scheme, since that’s often associated with certain factions or jobs and could potentially lead to confusion. Maybe a teal visor with orange or red filters would work here? Something similar to the colors indie miner and engineer gear use

Imaginos16:
> Something similar to the colors indie miner and engineer gear use

Turns out that was the entire point. This does match gear that engineers and miners use, particularly their hardsuits:
image

Since the case was made, I think I'm fine with letting this enter the game.

Imaginos16 (approved):

PR #5318, opened 2025-09-13
tg clickbox pr port

Erikafox:
## About The Pull Request
ports https://github.com/tgstation/tgstation/pull/61032 from tg

## Why It's Good For The Game

another pre-req + easier to make small objects (world icons) clickable
## Changelog

:cl: Ghommie, erika (port fox)
 fix: bees, hivelord, its brood and legion skulls no longer use the MOUSE_OPACITY_OPAQUE setting that makes the invisible portions of their icons count toward mouse clicks, which can get in the way of players trying to click stuff below their layer. Instead, to be easier to click on, they use a component that gives them a near-invisible underlay.
/:cl:


thgvr:
With the way this works I'm really wondering if its better or worse than just adding 1 alpha pixels to the sprites in question. seems kinda arbitrary here idk. This file could get *really* big
image

FalloutFalcon (approved):
I think I def agree with thgvr that this needs to not be abused but its still fine to have around and the rest of the code changes are good.

thgvr:
i guess

PR #5322, opened 2025-09-15
Fixes some broken clothing equip delays

Tamamitsune:
## About The Pull Request

~~Updates the equip delays for all items to be slightly faster, and~~ allows equipping armor while moving (not hardsuits).
Also, fixes a number of factional suit items using the wrong equip delay values

## Why It's Good For The Game

~~Some of the equip delays were pretty dang long and felt honestly just clunky. While you don't put on/off clothes too often, it did feel silly when just slipping on gloves took over a second, or putting on/removing a jacket took two whole seconds.~~
A decent number of factional suits were using the hard-suit equip delay and restrictions, when half the time you were just tossing on a jacket.

## Changelog

:cl:
fix: Fixed some factional suit items using the wrong equip delay
balance: You can now move while equipping armor, such as bulletproof vests.
/:cl:


thgvr:
Afraid this isn't the answer. If there is an item that feels like it's taking too long to equip, there is a really good chance it is using the wrong delay. This stuff was carefully tuned and intended so you can't just swap your gear on a whim and give some importance to the act of gearing up before an expedition from your ship.

Tamamitsune:
> Afraid this isn't the answer. If there is an item that feels like it's taking too long to equip, there is a really good chance it is using the wrong delay. This stuff was carefully tuned and intended so you can't just swap your gear on a whim and give some importance to the act of gearing up before an expedition from your ship.

Fair enough, I went a little overboard on a number of the tweaks I'll admit. I reverted those defines changes, and went through and found a number of items using the wrong delay and fixed those up.

code/modules/clothing/factions/hardliners.dm
```diff
@@ -111,6 +118,13 @@
...
 	togglename = "buttons"
 
+	equipping_sound = EQUIP_SOUND_MED_GENERIC
+	unequipping_sound = UNEQUIP_SOUND_MED_GENERIC
+	equip_delay_self = EQUIP_DELAY_SUIT
+	equip_delay_other = EQUIP_DELAY_SUIT * 1.5
+	strip_delay = EQUIP_DELAY_SUIT * 1.5
+	equip_self_flags = EQUIP_ALLOW_MOVEMENT | EQUIP_SLOWDOWN
+
 /obj/item/clothing/suit/toggle/armor/vest/hardliners/Initialize()
```
FalloutFalcon, line 128: I know this is a jacket but I would assume anything with decent armor should have the balance of armor?

code/modules/clothing/suits/armor.dm
```diff
@@ -22,7 +22,7 @@
 	equip_delay_self = EQUIP_DELAY_SUIT
 	equip_delay_other = EQUIP_DELAY_SUIT * 1.5
 	strip_delay = EQUIP_DELAY_SUIT * 1.5
-	equip_self_flags = null
+	equip_self_flags = EQUIP_ALLOW_MOVEMENT | EQUIP_SLOWDOWN
```
FalloutFalcon, line 25: Will need to get thgvr's opinion of whether this can actually be left in.

code/modules/clothing/factions/hardliners.dm
Tamamitsune: Sorry for hella late reply, had two rough work weeks in a row; but yeah i agree, those lines make it so it does have the typical armor delay on it, there wasnt a base clothing/suit/toggle/armor to inherit from, just the clothing/suit/toggle giving it the standard jacket one. Would it be better to add clothing/suit/toggle/armor to inherit from?

code/modules/clothing/suits/armor.dm
Tamamitsune: fair enough, i personally just feel its a bit odd/stiff being forced to stand still for that, when i could feasibly see most of the armor items that aren't full suits not completely inhibiting movement while trying to fit 'em, and i feel the slowdown and time it takes already to actually fully put them on stops any issues concerning balance

thgvr:
we'll see how it goes

PR #5330, opened 2025-09-16
Opaque Smoke reduces your chance of being hit

Gristlebee:
## About The Pull Request

Being in opaque smoke will now conceal you and reduces your chance of being hit by projectiles.

Being nearby to the concealed target will raise your chance to hit, while having thermals will entirely negate the concealment protection.

## Why It's Good For The Game

Charging any kind of entrenched chokepoint is an utter nightmare, especially with subshuttle boarding or outpost combat on the elevator. Smoke providing protection should make for an interesting defensive option for advances and retreat. 

## Changelog

:cl:
add: Concealment status
add: Smoke applies concealment
/:cl:


Madtankdriver:
Peak

code/datums/status_effects/buffs.dm
```diff
@@ -376,3 +376,22 @@
...
+	desc = "You're concealed and harder to hit with projectiles."
+	icon_state = "concealed"
+
+/datum/status_effect/concealed/tick(seconds_per_tick)
+	. = ..()
+	//look for smoke on tile
+	var/turf/smokey_tile = get_turf(owner)
+	for(var/i in smokey_tile)
+		if(istype(i, /obj/effect/particle_effect/smoke))
+			return TRUE
```
FalloutFalcon: pretty sure locate is way faster then setting up a for loop and type checking.
```suggestion
	if(locate(/obj/effect/particle_effect/smoke) in get_turf(owner))
		return TRUE
```

code/datums/status_effects/buffs.dm
```diff
@@ -376,3 +376,22 @@
 	owner.cut_overlay(overcharge)
 	owner.remove_movespeed_mod_immunities(type, /datum/movespeed_modifier/equipment_speedmod)
 	owner.remove_movespeed_mod_immunities(type, /datum/movespeed_modifier/gun)
+
+/datum/status_effect/concealed
+	id = "concealed"
+	var/concealment_power = 75
+	alert_type = /atom/movable/screen/alert/status_effect/concealed
```
FalloutFalcon, line 383: would give this a `tick_interval = 1` (default is `1 SECOND` so `10`), i think we want to avoid giving people up to a full second to peak out that grants the buff, this just ensures it fires on every fast_process basically.

FalloutFalcon (dismissed):
cant wait for half and full cover next /j

code/datums/status_effects/buffs.dm
Gristlebee: gotcha. It's done

PR #5338, opened 2025-09-17
Fix a warn

Erikafox:
## About The Pull Request

some signals are registered with the proper method 

## Changelog

:cl:
code: fixed the register signal warning
/:cl:


FalloutFalcon (approved):
Dont think this is the only one. I need to comb through it again and see if i can find any more at some point.

PR #5341, opened 2025-09-18
Silicon related stuff in cargo

Burning02:
## About The Pull Request

As is the title, this PR adds stuff related to cyborgs mostly into cargo without breaking any balance, I hope.

## Why It's Good For The Game

As it is currently with the pseudo-removal of RnD from almost nearly everywhere ingame, obtaining upgrades for cyborgs has become next to impossible without having a Talos in-sector and enough credits to convince them to research such. This'll fix that somewhat, by adding modules that I believe not to be too strong, but still allow cyborgs some form of customization without digging a hole into your pocket.

## Changelog

:cl:
add: Adds the "Silicon Modules" cargo tab, with a handful of useful upgrades meant for Cyborgs.
balance: The Cyborg chassis, MMI, and Boris module have been moved to cargo
/:cl:

### DIFF
--- a/code/modules/cargo/packs/silicon_modules.dm
+++ b/code/modules/cargo/packs/silicon_modules.dm
@@ -11,6 +11,7 @@
 	desc = "Contains disassembled Silicon chassis within, with nearly all parts included! Cables not included."
 	cost = 1250
 	contains = list(/obj/item/robot_suit,
+					/obj/item/bodypart/head/robot,
 					/obj/item/bodypart/chest/robot,
 					/obj/item/bodypart/l_arm/robot,
 					/obj/item/bodypart/leg/left/robot,
--- a/code/modules/cargo/packs/silicon_modules.dm
+++ b/code/modules/cargo/packs/silicon_modules.dm
@@ -1,0 +1,92 @@
+/datum/supply_pack/silicon
+	category = "Silicon Modules"
+	crate_type = /obj/structure/closet/crate/science
+
+/*
+		Cyborg Kit
+*/
+
+/datum/supply_pack/silicon/borg
+	name = "Silicon Construction Kit"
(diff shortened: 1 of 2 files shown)


code/modules/cargo/packs/silicon_modules.dm
```diff
@@ -0,0 +1,93 @@
...
+					/obj/item/bodypart/l_arm/robot,
+					/obj/item/bodypart/leg/left/robot,
+					/obj/item/bodypart/r_arm/robot,
+					/obj/item/bodypart/leg/right/robot,
+					/obj/item/assembly/flash/handheld,
+					/obj/item/assembly/flash/handheld,
+					/obj/item/stock_parts/cell/super)
+	crate_name = "Cyborg Construction Kit"
+
+/datum/supply_pack/silicon/mmi
```
thgvr, line 25: not so sure about this one. dunno if we really want people gettin borged

Burning02: Sorry if I come off as rude, but is there really any problems with the MMI? We kinda already have something similar to cyborgs having an organic brain (FBP's), so I personally don't see an issue with it, and it has been on the black market without any issues too

generalthrax: MMI's are prob fine. It'd be funny to FBP someone who got obliterated in round

thgvr: im 99% sure you cannot put a regular mmi into a borg, and the posi selectable mmi is a different type

thgvr:
really torn on if we want to actually include borgs more like this. medborgs are insane, so are most other iterations

thgvr:
decided ultimately this stuff shouldn't be super available, if it even should at all.

PR #5346, opened 2025-09-19
Miscgunchanges

generalthrax:
## About The Pull Request

- Increased PDW price slightly
- Decreased machine pistol price slightly
- Decreased SMG fire delay across the board to be lower (seems to have done nothing for some reason but i dunno how to fix it)
- Adds proper slowdowns to some long rifles (notably the Scout and Taipan)\
- Increased Scout's recoil
- Boomslang slows you less

## Why It's Good For The Game

Couple tiny adjustments. Noticed a lot of PDWs were the same price as SMGs. Thought it was silly. Also hikes up the Scouts recoil a fair sum since it had like Almost None before

## Changelog

:cl:
balance: Shuffled around some PDW / Machine pistol prices, decreased SMG fire delay, proper slowdowns for certain longrifles
balance: Scout recoil increased
balance: Boomslang and F90 slows you less
/:cl:


code/modules/projectiles/guns/manufacturer/clip_lanchester/ballistics.dm
```diff
@@ -443,7 +443,7 @@ NO_MAG_GUN_HELPER(automatic/marksman/f4/inteq)
 	spread_unwielded = 35
 	recoil = 2
 	recoil_unwielded = 10
-	wield_slowdown = SNIPER_SLOWDOWN
+	wield_slowdown = DMR_SLOWDOWN
```
Erikafox: what's up with this one the f90 is very much a sniper rifle

code/modules/projectiles/guns/manufacturer/scarborough/ballistics.dm
```diff
@@ -570,7 +570,7 @@ NO_MAG_GUN_HELPER(automatic/smg/sidewinder)
 	spread_unwielded = 35
 	recoil = 2
 	recoil_unwielded = 10
-	wield_slowdown = SNIPER_SLOWDOWN
+	wield_slowdown = DMR_SLOWDOWN
```
Erikafox: so is the boomslang

Erikafox (dismissed):
if the sniper slowdown define is too oppressive it should be tuned rather than just moving sniper rifles onto another speed define

code/modules/projectiles/guns/manufacturer/clip_lanchester/ballistics.dm
generalthrax: It's a fairly oppressive slowdown for a rifle that's statistically worse than similar snipers in every way. I can make a snowflake slowdown define just for that if needed.

To elaborate I think the slowdown is fine for other snipers (namely the Scout and others of similar strength), I don't think it warrants making snipers across the board quicker

Erikafox: I'm still not entirely sold. A lot of their power comes from the distance that you can engage at and adding more mobility to that is risky.

generalthrax: As a note, the Boomslang (notably the indie variant) does also have a much shorter scope while still maintaining the long slowdown

PR #5351, opened 2025-09-19
Bodycamera usability improvements

zimon9:
## About The Pull Request

This PR brings along a few quality of life changes to bodycameras. Namely:

- Bodycameras can now be used inhand in order to set their nametag and their network.
- Bodycameras can now be used on another bodycamera or bodycamera derivative in order to transfer their network. This should reduce the annoyance of having to use a tgui window every time a camera needs to be set.
- Bodycameras now display their tag on the item, making it easy to differentiate bodycameras after they've been set.
- Bodycameras can now be placed inside anything that it can fit inside, except for boots, and still work. This means that they work in bags now.
- Broadcast cameras do not inherit inhand configuration, but they do inherit their object getting renamed in accordance with their tag.
- Bodycamera crates have been modified so that they only carry one camera, and so that they no longer carry the fluff user's guide. Their price has also been decreased per unit camera to 100 credits.

Bodycameras:
image

Broadcast Cameras:
image

## Why It's Good For The Game
These quality of life changes should address the annoyances of using and configuring bodycameras. 
The guiding design principle here was that of reducing the number of clicks needed to get a particular task done. This should reduce the amount of busywork that setting a bodycamera currently requires. 
Having bodycameras be modifiable from inhand also makes it so that their end users can set their own tags, after their networks have been set by someone like, for instance, their captain. This parallelization of work should further reduce the trouble of getting a set of bodycameras up and running.

## Changelog

:cl:
add: Added bodycamera and broadcast camera obj tags
add: Added easy bodycamera network transfer
add: Added easy bodycamera configuration
balance: bodycameras now work anywhere except boots
balance: bodycameras now cost 100 credits each
/:cl:

### DIFF
--- a/code/game/machinery/computer/camera.dm
+++ b/code/game/machinery/computer/camera.dm
@@ -237,7 +237,7 @@
 		var/cam_location = active_camera_B.loc
 
 		// Is the camera in the following items? If so, let it transmit an image as normal
-		if((istype(cam_location, /obj/item/clothing/suit)) || (istype(cam_location, /obj/item/clothing/head/helmet)) || istype(cam_location, /obj/item/storage/belt) || istype(cam_location, /obj/item/storage/pouch)) //Should probably be refactored into excluding backpacks and boots instead of the current whitelist if more places need to be added
+		if(!(istype(cam_location, /obj/item/clothing/shoes)))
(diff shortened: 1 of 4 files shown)


FalloutFalcon (approved):
This PR has been rotting. 
image

PR #5353, opened 2025-09-20
Raleigh Update

Erikafox:
## About The Pull Request
Throws a beer bottle at my unloved child so it'll be better.

image

>armory lost ringneck, gained hellfire
>bandoliers
>redesign of rec spaces

## Why It's Good For The Game
Raleigh is better. It's not good enough though.

## Changelog

:cl:
add: The Raleigh has received a redesign. Please pay 250kcr to purchase a new one.
/:cl:


moffball19:
trial maptie trial reviewing this

i'd anchor these cabinets
image
<br>
they could use a few spare backpacks/satchels/duffels in the cabinets' contents
image
<br>
i don't think the orientation of the couch makes sense, i'd remove the middle portion and have 2 separate couches facing each other
image
<br>
these bottom doors are labelled incorrectly
image
<br>
the black sidings in this hall feel out of place to me. i'd extend them a bit further to the doors (look at how the venetia does it's central hallway)
image
<br>
i'd anchor the fridge
image
<br>
i think the black siding should layer over the red border here
image
<br>
consider giving the 2ic and cryo a nice view to space. and anchoring the 2ic's cabinet
image
<br>
image
<br>
a handrail in front of the desk here might be nice
image
<br>
cargo is missing a lightswitch
image
<br>
this room is missing a vent. i'd change this scrubber pipe to be the hidden subtype
image
<br>
robert is going to cause problems roundstart and eat wires. i'd replace them with a less disruptful pet
image
<br>

### armoury contents

they really don't need the hammer now that combat exosuits are 99% unobtainable
image
<br>
replace the .45 ammo can with 4.6 for the resolution
image
<br>
an empty subtype for the e-40 would keep it consistent with the other weapons here. put 1 more eoehoma cell into the mag locker if you do this
image
<br>

### misc

this area name seems unfitting to me. i'd recommend 'crew' (crew quarters) instead
image
<br>
consider rearranging the hall something like this to keep obstructions out of the main walkway
image
<br>
these airlocks are missing names. this isn't mandatory, but is notably absent when the rest of the airlocks have them
image

Apogee-dev:
after years of top 10 goliath moments happening to ships with windowed cryobays i’d advise against giving that particular room a view into space

Erikafox:
We have an advanced technique known as "closing the window"

PR #5354, opened 2025-09-20
Outpost medical vender hourly refills

zimon9:
## About The Pull Request

This PR adds an hourly refill function to vending machines, set OFF by default. The only vender type with this function active currently are OutpostMed vending machines.

## Why It's Good For The Game

It makes sense that an outpost would actively restock their vending machines -- and this also allows players to use the outpost medical vending machine for specific medical supplies without having to worry too much about someone else buying out stock before them. 

## Changelog

:cl:
add: Added an hourly restock function to vending machines, turned off by default
balance: OutpostMed venders now automatically restock themselves every hour
/:cl:


thgvr:
might need to be more often, but probablny not

PR #5359, opened 2025-09-20
Grinder packs are no longer huge

zimon9:
## About The Pull Request

This PR brings the weight class of grinder packs down from huge to bulky, which allows them to now fit inside backpacks and dufflebags.

## Why It's Good For The Game

Oftentimes when I'm on EVA with a grinder, I find it cumbersome to have to carry both it and my bag in my hands -- doubly so if I have to bring a jetpack with me, too. I think it makes sense for a dufflebag to be able to hold one, as they seem to be about torso sized in length, and about half the diameter of a typical torso.

## Changelog

:cl:
balance: grinder pack weight classes were brought down from huge to bulky
/:cl:


thgvr:
Don't think this is a good idea, it's intentionally very bulky

zimon9:
> Don't think this is a good idea, it's intentionally very bulky

Oh, alright -- I'll close this, then

PR #5360, opened 2025-09-20
Adds Sabres and Swords to the Outpost Market

Crystal1Blitz:
## About The Pull Request
Adds every faction's swords to their respective faction markets, with the base version of each faction's sword being 750, 3x the Machete.

## Why It's Good For The Game
Most factions have their own unique melee weapons, but they're barely used due to their rarity, which most ships only having one set of a given sword, and each sword being nigh unreplaceable unless you get lucky with the black market, which is a shame considering the work that went into some of these.

## Changelog

:cl:
add: Swords to factional cargo

/:cl:


Apogee-dev:
Can you remove the Nanotrasen saber and the SUNS Captain saber? Those two are exclusively for captains and shouldnt really be something you can just buy as many as you want even if you’re the same faction. The others are all issued more widely and are fine

Crystal1Blitz:
> Can you remove the Nanotrasen saber and the SUNS Captain saber? Those two are exclusively for captains and shouldnt really be something you can just buy as many as you want even if you’re the same faction. The others are all issued more widely and are fine

Makes sense, ill change it when I get the time

PR #5365, opened 2025-09-22
Removes Player-Accessible Rocket Launchers

firebudgy:
## About The Pull Request

Removes the Hammer and Hammer-DP from the Black Market, as well as the Hammer from the icemoon training facility. Raleigh remap is pending (with the Hammer already planned to be removed) so to avoid merge conflicts it isn't removed in this PR.

Also removes HEDP (rockets with devastation) from the Mako belt.

## Why It's Good For The Game

Rocket Launchers have, under every circumstance they've been used, never been seen positively.

https://github.com/user-attachments/assets/fd56de17-ba94-449a-b79c-07151be4d0c0

Despite being meant for use against mechs, the majority of their usage has been against players - both from Hammers obtained in-round and from MAKOs and Hammers given by admins. Usually, this results in someone just getting killed (which is realistic) and having to sit things out. Compared to just being in a firefight, however, these offer little to no counterplay or even time to react.

With the additional fact that combat exosuits are effectively unobtainable outside of niche circumstances and simplemobs not warranting them, the use case of having launchers is becoming less and less prevalent by the day.

## Changelog

:cl:
balance: Rocket launchers are no longer player accessible.
balance: Mako belts no longer have HEDP rockets.
/:cl:

### DIFF
--- a/_maps/RandomRuins/IceRuins/icemoon_tesla_lab.dmm
+++ b/_maps/RandomRuins/IceRuins/icemoon_tesla_lab.dmm
@@ -904,6 +904,7 @@
 	},
 /obj/effect/spawner/random/trash/mess,
 /obj/machinery/airalarm/directional/south,
+/obj/effect/decal/cleanable/dirt/dust,
 /turf/open/floor/plasteel/tech/grid,
 /area/ruin/icemoon/tesla_lab/armory)
 "dP" = (
@@ -4834,18 +4835,13 @@
 /obj/effect/turf_decal/trimline/transparent/blue/warning{
 	dir = 4
 	},
-/obj/effect/turf_decal/trimline/transparent/blue/filled/corner{
-	dir = 1
-	},
-/obj/effect/turf_decal/trimline/transparent/blue/filled/corner{
+/obj/effect/turf_decal/siding/thinplating/dark{
 	dir = 8
 	},
-/obj/effect/turf_decal/siding/thinplating/dark/corner{
-	dir = 1
-	},
-/obj/effect/turf_decal/siding/thinplating/dark/corner{
+/obj/effect/turf_decal/trimline/transparent/blue/filled/line{
 	dir = 8
 	},
+/obj/effect/decal/cleanable/dirt/dust,
 /turf/open/floor/plasteel/tech/grid,
 /area/ruin/icemoon/tesla_lab/armory)
 "tU" = (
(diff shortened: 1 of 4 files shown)


FalloutFalcon (approved):

thearbiber (approved):

thgvr:
IDK about the reasoning given here, could be applied to certainly a lot of weapons.
if it's being used enough to be annoying we definitely can look at its sources and balance of the item itself

If anything instead of pretty much removing them I'd like to explore aspects to make them more useful for planetary combat and used as primarily a rare loot item.

thgvr (dismissed):
what I said above. if you don't want to edit this PR to explore other changes than this, you can close and make another. I don't feel these are that common or a problem to remove all player access to it

i'm happy to discuss it but nobody has so far

thgvr:
if you have no intent to discuss this anywhere, the pr will be closed

zimon9:
If I may give my two cents -- I think the way that rocket launchers like the hammer differentiates itself from other weapons is in its ability to kill people in a single shot, and to destroy many of the items on them, no less. 

Sure, it's a player problem, but even if that were not the case, I don't see much of a use for them at present, personally. They're not worth using as a breaching tool due to their extravagant price, they go into overkill territory when used on another player, and there is nothing strong enough to take a hit from a rocket while keeping their occupants safe, if worse for wear.

If rockets were to remain, I'd imagine that they'd be nicer to work around if they didn't destroy items so easily, and/or if they didn't kill someone instantly on a single hit, but that would be fairly unrealistic.

thgvr:
there was a fairly significant nerf to item damage from explosions recently, could be lowered even further. better breaching capability is also a good idea, but neither of those will matter if they're just made unused.

moffball19:
> there was a fairly significant nerf to item damage from explosions recently, could be lowered even further. better breaching capability is also a good idea, but neither of those will matter if they're just made unused.

imo we don't need rocket tag, i've never seen these used in pvp in a healthy way and they'll continue to be used this way as long as they're accessible

https://github.com/user-attachments/assets/d2bbe9cb-245b-4bbc-acbc-fda14da12df3

https://github.com/user-attachments/assets/68ccdd48-ff54-48b7-89df-bd513eeb94d6

https://github.com/user-attachments/assets/4764c4a4-ef9d-4675-934d-1c2010918ec5

thgvr:
valid yeah, it'd be pretty difficult to make them interesting in ways that would be fun for the game and definitely out of scope for something like this.

anyways, if this pr wants to remove the availability it's probably best to fully remove them

Gristlebee:
Imo, I think just an accessibility removal is fine rather than a full codebase removal. I agree they're problematic in the wild, but I think they're fine as an effectively aspawn item to be used at admin discretion if the situation calls for it. If down the line we figure out a fun way to rework it, that's great.

PR #5367, opened 2025-09-22
gruboid no longer explode

firebudgy:
## About The Pull Request

gruboid stop exploding now

## Why It's Good For The Game

This just isn't fun, ask anyone how fun running from exploding common simplemobs that can SET YOU ON FIRE is. Plus we already had a PR previously to do this that went stale like six months ago.

## Changelog

:cl:
balance: Gruboids are no longer classified as "Hazardous Explosive Materials" by CLIP.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/mining_mobs/elites/goliath_broodmother.dm
+++ b/code/modules/mob/living/simple_animal/hostile/mining_mobs/elites/goliath_broodmother.dm
@@ -210,12 +210,6 @@
 		visible_message(span_warning("[src] digs one of its tentacles under [target]!"))
 		new /obj/effect/temp_visual/goliath_tentacle/broodmother(tturf, src)
 
-/mob/living/simple_animal/hostile/asteroid/elite/broodmother_child/death()
-	. = ..()
-	visible_message(span_warning("[src] explodes!"))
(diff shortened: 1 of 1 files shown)


Mirag1993:
Kinda questionable to just strip the explosion and leave it as yet another generic mob. If you’re removing a unique mechanic, it should really be replaced with something equally interesting, not just nothing.

Apogee-dev:
Would be neat, if definitely higher-effort, to “deactivate” the gruboids for about a second or two when killed before they actually explode, so players have a moment to get clear / an obvious telegraph like with other mob attacks such as goliaths and brimdemons. iirc CLAW does something kind of like this when it enters its second phase? that’s if the effort would be worth it to keep some form of the mechanic in anyway.

generalthrax:
prob out of scope. them exploding is already questionable and is pretty much only cause they inherited the behaviour from broodmother babies

Apogee-dev:
I’d say “maybe later” but I know I wouldn’t have the time to code it, lol. all good.

thgvr:
> Kinda questionable to just strip the explosion and leave it as yet another generic mob. If you’re removing a unique mechanic, it should really be replaced with something equally interesting, not just nothing.

yeah it's being replaced with nothing which is as interesting as them exploding for no reason

PR #5368, opened 2025-09-24
ERTs/Inspectors get proper access

Gristlebee:
## About The Pull Request

ERTs and inspectors were using normal job datums for access. Which meant they couldnt use the ERT outpost bays and associated areas. Adds proper job datums and some general define sanity for region access.

Also removes the microbombs from Ramzi ERT

## Why It's Good For The Game

ERTs should have ERT access I think. And the Ramzi ERT microbombs made them a bit of a pain to use. 

## Changelog

:cl:
fix: ERT access
del: Ramzi ert microbomb
/:cl:


meemofcourse (approved):
womp

PR #5369, opened 2025-09-24
Excess damage on limbs can transfer to the torso

Gristlebee:
## About The Pull Request

Excess damage dealt limbs can transfer over to the main body's torso. Excess brute and burn is transferred at half effectiveness. (E.g. If your limb has 40 brute damage and a damage cap of 50. And it gets hit by a bullet that deals 20 brute damage, 10 damage is dealt to the limb, and the excess 10 brute is halved to 5, and applied to the torso.)

Excess stamina damage is transferred at 100% effectiveness as you're tiring/exhausting your target.

Also fixes rubber rounds not apply stamina damage properly to limbs.

## Why It's Good For The Game

Targeting limbs is a generally frustrating experience, due to lack of easily noticeable feedback in combat and the fact any excess damage you deal is effectively wasted. Overdamaging a limb now is still less efficient in most scenarios, but now not entirely worthless.

Stamina transferring at full effectiveness I think should also give it an interesting niche for circumventing armor if you can land hits on the limbs at the cost of not dealing "real" damage.

## Changelog

:cl:
add: Excess damage can transfer to the torso.
fix: Rubber rounds not applying stamina damage to limbs.
/:cl:


thgvr:
one thing here that you're trying to solve with over damaged limbs is technically a bug right now, a wound won't upgrade into a bone break and I haven't had time to fix it. if that were working - an over damaged limb would be dismembered

thgvr:
i'd like to see that bug fixed rather than merge this, fwiw

PR #5378, opened 2025-09-26
Plasmacutter adjustments

Erikafox:
## About The Pull Request

Resprites + Reflavors the plasmacutter as a proper industrial tool. 
The cutter now has two modes, welder and deconstruction, which you can switch between too. It also has a short range projectile that you can fire off. It works a bit faster and can be recharged with power cells, rather than plasma.

## Changelog

:cl:
balance: The plasmacutter has been reworked!
/:cl:


rye-rice (approved):

code/game/objects/items/melee/plasma_cutter.dm
```diff
@@ -0,0 +1,188 @@
...
+		return
+
+/obj/item/plasmacutter/use_tool(atom/target, mob/living/user, delay, amount=1, volume=0, datum/callback/extra_checks)
+	if(amount)
+		if(tool_behaviour==TOOL_DECONSTRUCT)
+			target.add_overlay(GLOB.advanced_cutting_effect)
+		else
+			target.add_overlay(GLOB.welding_sparks)
+		. = ..()
+		if(tool_behaviour==TOOL_DECONSTRUCT)
```
iamthesilly: If the tool mode is toggled during the parent's do-after it won't remove the correct overlay.

SapphicOverload (changes requested):

SapphicOverload (approved):

PR #5384, opened 2025-09-27
fix html decodes in like 2 spots

Erikafox:
## About The Pull Request
holopads will no longer blast you with html encodes
hailing will no  longer blast you with html encodes

## Changelog

:cl:
fix: hailing/holopad sanitization
/:cl:

### DIFF
--- a/code/modules/tgui/tgui_input_text.dm
+++ b/code/modules/tgui/tgui_input_text.dm
@@ -142,7 +142,7 @@
 			if(max_length)
 				if(length_char(params["entry"]) > max_length)
 					CRASH("[usr] typed a text string longer than the max length")
-				if(encode && (length_char(trim(params["entry"])) > max_length))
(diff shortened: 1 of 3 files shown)


code/modules/tgui/tgui_input_text.dm
```diff
@@ -142,7 +142,7 @@
 			if(max_length)
 				if(length_char(params["entry"]) > max_length)
 					CRASH("[usr] typed a text string longer than the max length")
-				if(encode && (length_char(html_encode(params["entry"])) > max_length))
+				if(encode && (length_char(trim(params["entry"])) > max_length))
```
FalloutFalcon: what is the point of the trim here that is replacing the html_encode. Also def need to document this as it _could_ effect other random things that use tgui_input_text

Erikafox: this is called the "3 am forgot i did this in the pr"

PR #5403, opened 2025-10-02
adds ramzi hideout, new moon ruin

thearbiber:
## About The Pull Request

Yeap. Notable loot is just boomslang and blasting hammer
image
image
image

## Why It's Good For The Game

more ruins for the planet we just opened.

## Changelog

:cl:
add: Ramzi Hideout Ruin
/:cl:


code/game/area/areas/ruins/rockplanet.dm
```diff
@@ -131,7 +131,7 @@
 	icon_state = "hallP"
 
 /area/ruin/rockplanet/rust_base/bridge
-	name = "ISV Rust Base Bridge"
```
generalthrax, line 134: Think I accidentally did this somehow initially

code/modules/mob/living/simple_animal/corpse.dm
```diff
@@ -54,6 +54,18 @@
...
 	outfit = /datum/outfit/ramzi/soft/surplus
 
+/obj/effect/mob_spawn/human/corpse/ramzi/towel
+	name = "Ramzi Showerer"
+	outfit = /datum/outfit/ramzi/towel
+
+/datum/outfit/ramzi/towel
+	name = "Ramzi Clique Showerer"
+	uniform = null
+	head = null
```
generalthrax, line 76: You'll want to set preeeetty much every slot here to null (except gas mask prob). Right now checks are failing cause its trying to give them an ID when they can't fit an ID

Addust:
the showerers should just be screaming ones that verbally complain about their shower being interrupted

i mean if you were a pirate and someone interrupted your shower (quite probably after having killed a lot of your buddies) you'd be pretty pissed too

Gristlebee (dismissed):
- The front gate holofields are not wired.
image

- The right half of the engineering room could use more lighting.
image

- Is this locker meant to be empty aside from a pair of gloves?
image

- A light near the doors could be nice, the room's a bit dark otherwise.
image

- This washroom has no lights.
image

- A minor pistol like a ringneck or a cottonneck could be nice loot for the armory, it's a little bit empty otherwise.
image

- Some good food in the fancy canteen could be thematic.
image

- Some trash and used towels/laundry in these bins would be thematic.
image
image

- Some construction supplies in the cargo room or somewhere else would be flavourful I think, to really put the message across this is a place under construction.
image

- The base turfs under rock are airless, and will start attempting to drain any atmos in the ruin if broken. I'd suggest making a subtype turf for the wall's baseturfs that has the planetary atmos var set to false so they can fill and won't constantly drain.
image

- The edges here could use a little smoothing so isn't just flat and straight, and blends better if it touches a different biome.
image
image

- Every single personal locker aside from the officer's room is empty. Is that intended? I think filling them in would be nice, and adding some personal effects here and there would help make the ruin feel lived in.

thearbiber:
should be good.

Erikafox:
dreamseeker_IbLLfevC2T

was not fond of the of the tile merging here, might wanna use a different tile around the base of the platform

dreamseeker_gPcHFh1FSN

thoughts about making these two shutters still functional? You already added the button and it could be a neat interaction to see.

dreamseeker_OZkWlYb8b5
Probably want a way to access the cargo from the outside - crates are heavy and it's the difference between 2/3 doors and 1

PR #5410, opened 2025-10-04
godawful hack

Erikafox:
## About The Pull Request

"fixes" the verb loss on logout by just re-initializing the entire menu

as it turns out init verbs doesn't catch verbs that are just GONE which is what's HAPPENING I don't KNOW WHYYYY

## Changelog

:cl:
fix: verb loss
/:cl:


thgvr:
sure this probably has no consequences at all im sure

FalloutFalcon:
image

PR #5412, opened 2025-10-04
Make radiation stop forcing baldness

Retrino:
## About The Pull Request

Removes radiation's balding effect. 

## Why It's Good For The Game

This mechanic is pretty annoying, and very easy to be inflicted with without noticing (radioactive materials not always being stored correctly, waste planets, etc.) The damage type itself doesn't have much going for it but this current implementation just feels Not Great

## Changelog

:cl:
del: Removed radiation causing hair to fall out
/:cl:


code/modules/mob/living/carbon/human/species.dm
```diff
@@ -1435,18 +1435,6 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 	if(radiation > RAD_MOB_VOMIT && prob(RAD_MOB_VOMIT_PROB))
 		H.vomit(10, TRUE)
 
-	if(radiation > RAD_MOB_HAIRLOSS)
-		if(prob(15) && !(H.hairstyle == "Bald") && (HAIR in species_traits))
-			to_chat(H, span_danger("Your hair starts to fall out in clumps..."))
```
Erikafox, line 1453: I think there being an indication of lower levels of radiation damage is good. I'd just remove the proc

Erikafox (dismissed):

code/modules/mob/living/carbon/human/species.dm
thgvr: needs a further rework later instead for feedback considering most species dont have hair on their head

Erikafox: Change the to_chat to be related to ARS symptoms - dizziness, headache, so forth - have it throw that message as a warning on the prob(15) chance. Will be good to merge then.

FalloutFalcon (approved):
Satisfied Erika's request from what I can tell.

PR #5437, opened 2025-10-10
Changes the orientation of the briefing room on the Kali Andhi

zimon9:
## About The Pull Request

This map change makes it so that the briefing room has its chairs and podium moved so that the sidestates of characters in the room are shown instead of the front state and the backstate.

Before:
image

After:
image

image

## Why It's Good For The Game

It can be a little awkward to have your backsprite be displayed when you're in a briefing, or giving one. Changing the orientation of the briefing room enables everyone to show their sidestate, which tends to be better.

## Changelog

:cl:
refactor: modified the orientation of the Kali Andhi's briefing room.
/:cl:


Erikafox (approved):

Addust:
also while you're at it can you fix the areas in the back? one of the engine bays has a duplicated area type (both use port engines instead of one being port and one being starboard)

zimon9:
> also while you're at it can you fix the areas in the back? one of the engine bays has a duplicated area type (both use port engines instead of one being port and one being starboard)

I'd rather atomize a change like that.

moffball19:
trial maptie trial maptaining. i'd suggest moving the operative lockers to the left wall and the trashcan up to the top left corner. the way the lockers are placed now offers no room to move past them without stepping onto their occupied tiles or the chairs, which isn't ideal when 3 people are trying to access them at once. other than that, this change looks fine to me
image

moffball19:
on further insight, this room is also lacking an apc. if you could add that while you're at it, it'd be great

zimon9:
Thank you - I've set up the room up now as shown below:
image

Does this look okay? I'm unsure if it's a little silly for the APC to be where it is, but I couldn't think of a more symmetric spot for it

PR #5445, opened 2025-10-11
tweaks the Sabers

Crystal1Blitz:
## About The Pull Request

Ups the sabers force from 18 to 25, and lowers their blocking chance from 60% to 30%
## Why It's Good For The Game

The saber, as it is in game, does 18 force, 2 less then a combat knife, meaning, it takes one more hit to crit a unarmored person with a saber then a knife, and one less hit for a saber compared to the knife, adding this with the fact that sabers take up the belt or back slot, and they're almost never used except for fashion, this change makes it so sabers can crit in 4-5 hits on unarmored, and 5 on basic armor, giving it a edge against the combat knife, while still being a hit weaker then the Boarding cutlass.

The block chance was also insane, with 60% I was blocking 13 hits in a row on average, making melee mobs harmless, I hope 30% is at least more reasonable

## Changelog

:cl:
balance: increased damage on sabers, lowered block chance

/:cl:


DIB-DOG:
With this rebalance, the PGF Cutlass would not be notably more powerful than the normal saber but would still suffer from 15 less block chance. Would it be prudent with this re-balance to just bring the cutlass down to 25 like the normal saber and raise its block chance to 30 to match?

Either that or buff its damage more, but I fear that runs the risk of getting ridiculous.

Crystal1Blitz:
> With this rebalance, the PGF Cutlass would not be notably more powerful than the normal saber but would still suffer from 15 less block chance. Would it be prudent with this re-balance to just bring the cutlass down to 25 like the normal saber and raise its block chance to 30 to match?
> 
> Either that or buff its damage more, but I fear that runs the risk of getting ridiculous.

I feel like letting the cutlass have its own niche would be for the best, I believe upping the cutlass's destruction modifier and lowering its attack cooldown from 8 to 6 (cutlasses are meant to be faster then sabers anyways)

PR #5447, opened 2025-10-11
Grenade price parity with the black market

zimon9:
## About The Pull Request

The price of frag grenades and HE grenades on the black market has been adjusted to match the outpost catalog's prices more closely.

## Why It's Good For The Game

The black market is, from what I understand, meant to be a low quantity, variable cost option to source items from, with items generally costing more than what you'd get for them from the outpost. This change should help grenades fit that theme.

## Changelog

:cl:
balance: made grenades on the black market have prices similar to their outpost variants
/:cl:

### DIFF
--- a/code/game/objects/items/grenades/grenade.dm
+++ b/code/game/objects/items/grenades/grenade.dm
@@ -1,7 +1,7 @@
 /obj/item/grenade
 	name = "grenade"
 	desc = "It has an adjustable timer."
-	w_class = WEIGHT_CLASS_SMALL
+	w_class = WEIGHT_CLASS_NORMAL
 	icon = 'icons/obj/grenade.dmi'
 	icon_state = "grenade"
 	item_state = "flashbang"
--- a/code/game/objects/items/grenades/grenade.dm
+++ b/code/game/objects/items/grenades/grenade.dm
@@ -1,7 +1,7 @@
 /obj/item/grenade
 	name = "grenade"
 	desc = "It has an adjustable timer."
-	w_class = WEIGHT_CLASS_NORMAL
+	w_class = WEIGHT_CLASS_SMALL
 	icon = 'icons/obj/grenade.dmi'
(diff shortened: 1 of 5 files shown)


DIB-DOG:
At normal size can they still fit in webbing/armor pockets? I feel like those methods of storage shouldn't be punished. With webbing you're giving up space for magazines and quick access tools, and with armor pockets, the maximum is already two.

Would be a fairly straight-forward addition, would just have to add the grenade object to the list of accepted items in the exo-pocket component in "_code\datums\components\storage\concrete\pockets.dm_".

zimon9:
> At normal size can they still fit in webbing/armor pockets? I feel like those methods of storage shouldn't be punished.

I've sent a push that allows them to fit in military webbing now. They should be able to fit in belts as well. They don't fit in armor storage, but if they were made to fit, armor storage would need to be able to take normal items, or they'd need to have a specific exception for grenades, which might be a little silly. Do you feel like they should fit in armor?

DIB-DOG:
> Do you feel like they should fit in armor

Clipping grenades to your armor is a pretty normal thing IRL and being able to put the grenades into your armor slots so you don't unintentionally pull them is good for inventory management and safety. Adding the exception for armor storage and grenades isn't really much different than the exception that exists for webbing to allow normal sized pistols. I think it should be added.

zimon9:
> > Do you feel like they should fit in armor
> 
> Clipping grenades to your armor is a pretty normal thing IRL and being able to put the grenades into your armor slots so you don't unintentionally pull them is good for inventory management and safety. Adding the exception for armor storage and grenades isn't really much different than the exception that exists for webbing to allow normal sized pistols. I think it should be added.

Oh, my mistake, I wasn't aware. I think I'm going to revert them to being small sized again. The price hike will go back to normal as well, so all that this PR will really do is bring the black market price up to match the outpost market.

PR #5453, opened 2025-10-13
Pinscher back in cargo

SomeguyManperson:
## About The Pull Request

Undocumented removal

## Changelog

:cl:
add: The pinscher is orderable in cargo again
/:cl:

### DIFF
--- a/code/modules/cargo/packs/gun.dm
+++ b/code/modules/cargo/packs/gun.dm
@@ -103,6 +103,16 @@
 	faction_discount = 0
 	faction_locked = TRUE
 
+/datum/supply_pack/gun/pinscher
+	name = "Pinscher Heavy Pistol Crate"
+	desc = "Contains a heavy pistol manufactured by Serene Outdoors and modified to fit IRMG's standards. Chambered in .44 Roumain."
+	cost = 1250
+	contains = list(/obj/item/storage/guncase/pistol/pinscher)
+	crate_name = "pistol crate"
+	faction = /datum/faction/inteq
+	faction_discount = 0
+	faction_locked = TRUE
+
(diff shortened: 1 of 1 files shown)


generalthrax:
Lopped it in with the Schnauzer implementation as part of the Nanotrasen Advantage pr. Seems it was forgotten to be added in with changelog. It seemed poor taste to have two pistols with similar niches for the same faction so I opted to keep it for Honourable Vanguard use since it's the Token Serene Sporting retrofit. 

In hindsight it was probably unnecessary since .44 got buffed (by myself) and there'd be no reason to keep it out other than giving Inteq 3 semi-auto pistols

I'd probably increase the price to like 1500 though

Apogee-dev:
> Lopped it in with the Schnauzer implementation as part of the Nanotrasen Advantage pr. Seems it was forgotten to be added in with changelog. It seemed poor taste to have two pistols with similar niches for the same faction so I opted to keep it for Honourable Vanguard use since it's the Token Serene Sporting retrofit.
> 
> In hindsight it was probably unnecessary since .44 got buffed (by myself) and there'd be no reason to keep it out other than giving Inteq 3 semi-auto pistols
> 
> I'd probably increase the price to like 1500 though

In retrospect Inteq probably could have gone without the Schnauzer existing at all, honestly, seeing as (like you mentioned) they had the pinscher already and of the two, the pinscher is by far the more interesting pick. That's on me for asking rye to make an inteq podium variant in the first place, oops.

SomeguyManperson:
could put it back in the vanguard closets instead

thearbiber:
Think its fine as is right now honestly.

PR #5470, opened 2025-10-20
removes stuff that shouldn't be there

Goat-Real:
## About The Pull Request
noticed these salvaging and they annoyed me
the shutters in the n+s kitchen in the mining facility were doubled up
there was a wall inside of the bridge windows in the crashed starwalker
## Why It's Good For The Game
immersion ruining mapping mistake that must be purged
## Changelog

:cl: Goat
fix: removed random wall that was inside of a window on crashed starwalker
fix: the shutters are no longer doubled up in the n+s rock planet mining facility kitchen
/:cl:


Ratvarr:
look at singolo lab and turn off the rocks and you will find a rouge table

PR #5481, opened 2025-10-25
replaces all moonbase indestructible walls with normal

generalthrax:
## Why It's Good For The Game

idiots disease

## Changelog

:cl:
fix: moonbase is no longer indestructible
/:cl:


Zevotech (approved):
millions must scrap

LePiromano:
i don't know if the map's changed but i remember the ruin had 2 rooms that are main loot rooms but have no enemies in them so if the walls aren't indestructible you can just bust in and tkae everything (unless i'm thinking of the wrong moon ruin)

generalthrax:
Pretty sure I have every lootroom covered with enemies here. I think the other moon ruin also has it all covered, most of that ruins loot is in the hands of enemies

PR #5482, opened 2025-10-25
decreases bluespace jump to 20 seconds

generalthrax:
## About The Pull Request

decreases the bluespace jump delay from 1 minute to 20 seconds

## Why It's Good For The Game

kinda really long for no reason. its still enough time to intercept someone if theyre trying to flee from you too so should be fine

## Changelog

:cl:
balance: bluespace jump delay decreased from 1 minute to 20 seconds
/:cl:


Erikafox (dismissed):
take a look at some of the other announces that this ends up calling. It's kinda noise-spam and it'll be worse w/this.

generalthrax:
This seems to affect only this and from testing the sound isn't any more or less noise-spammy with 40 less seconds

PR #5496, opened 2025-10-27
Increases emergency tank volumes, and adds a pressure alert system

zimon9:
## About The Pull Request
Basic emergency tanks are effectively removed, with extended-capacity emergency tanks replacing them. Actual extended capacity emergency tanks now use the previously unused 'emergency-tst' sprite in order to differentiate them, but they don't have any actual differences. Their volumes were kept at 2. 

A visual and auditory alert system was also implemented to warn a user of a tank if their tank is getting low in pressure.
image
In the above image, the tank makes a sound and sends a voice alert regarding its pressure status.

This alert only makes a sound and notification when the pressure drops below 200 kPa, and again when the pressure approaches 0 kPa. The idea for the vocal notifications are that these alarms aren't standardized, so they can be dismissed if a user is unaware to their significance -- a vocal notification makes it so that the purpose of the alarm is clear.

dreamseeker_VQrpMIARTa
This gif shows the various overlay states for the pressure alert system, going from empty, to full pressure, respectively.

There's also some minor unrelated cleanup of commented out code regarding energy latches

## Why It's Good For The Game

A minor complaint I've heard a couple of times had to do with how readily emergency tanks emptied during use. I've also seen people only realize that their tank's run out when they get the status indicator that their character is lacking oxygen. This should help alleviate instances of that.

## Changelog

:cl:
add: Added an emergency tank pressure alert system
balance: Increased emergency tank volumes
/:cl:


LePiromano:
could just be me but aren't they emergency tanks for a reason? doesn't this kind of ruin the point of large tanks? the pressure warning is awesome but i feel like people should just learn to manage their oxygen better instead of babyproofing them by doubling/tripling their o2 time lmao, but it's understandable if people prefer to have a weapon slinged on their shoulder rather than a big ass o2 tank

DIB-DOG:
I agree with the other comment here. All the added features you’ve included are great, very well done, but I feel like increasing their capacity like this negates the purpose/trade off of using large back mounted tanks.

zimon9:
Yes, you're both right here. Originally I planned on just doubling the volumes, but I decided to triple them in order to bring extended capacity emergency tanks inline with phorid tanks, so they're now both equivalent in volume.
You're also both right that it's not hard to mitigate the challenge of keeping oxygen on you (especially just by carrying an extra tank or two) -- however, I have something almost complete that may make the extra volume worth having.

DIB-DOG:
>however, I have something almost complete that may make the extra volume worth having.

That is?

DIB-DOG:
Alright well I see the jetpacks. Also a cool implementation, but I still feel like that's one of those tradeoffs you should be expected to make for increased mobility/less need for air management. 

Choosing to sacrifice your backpack slot or suit storage slot is a meaningful choice which we have the items to make feasible. By utilizing pouches/rigging and using your boot/armor slots you can effectively remove your need to wear a backpack, which frees that slot up for a weapon, mobility tool, or air supply. I just fear that making emergency tanks even more universally effective will disincentivize making those kinds of choices.

zimon9:
> Alright well I see the jetpacks. Also a cool implementation, but I still feel like that's one of those tradeoffs you should be expected to make for increased mobility/less need for air management.
> 
> Choosing to sacrifice your backpack slot or suit storage slot is a meaningful choice which we have the items to make feasible. By utilizing pouches/rigging and using your boot/armor slots you can effectively remove your need to wear a backpack, which frees that slot up for a weapon, mobility tool, or air supply. I just fear that making emergency tanks even more universally effective will disincentivize making those kinds of choices.

That's a perfectly valid concern, and the storage choices you highlight are certainly part of the fun of the storage system -- perhaps I can drop the volumes down to 2x their original capacities, instead of 3x like now? --- This is kind of in lieu of having a system to just slot in a tank into a suit itself, as I'm having trouble getting something like that implemented, though it is something that I'd like to see done

thgvr:
na this is a good start, the tiny tanks were completely worthless being used up in like less than 15 mins

DIB-DOG:
> na this is a good start, the tiny tanks were completely worthless being used up in like less than 15 mins

I would argue that is exactly the point of them though. They're emergency survival tanks. A small package that has enough air for you to get to safety/reach an oxygenated environment and nothing more.

thgvr:
well I don't really think it should be that

DIB-DOG:
I guess I'm vetoed then, but in reality o2 tanks the size of our emergency tanks last like, 5-10 breaths. 

Not saying we should strive for that level of realism, because that would be annoying and not fun for anyone, but the fact that they last 10 or so minutes as they are, is already fairly impressive.

I just don't wanna see back mounted tanks become completely unnecessary and with too much of a buff to emergency tanks, I see them basically never being used.

LePiromano:
yeah if you're gonna change their purpose from being emergency to actual eva use at least change the name from "emergency" to "small" or something

Erikafox:
> na this is a good start, the tiny tanks were completely worthless being used up in like less than 15 mins

Most people overfill the tanks with air pumps - or purchase an oxygen can to refill them. I don't think making them bigger is smart. Maybe just make it so their full potential is a little more broadcast.

generalthrax:
Could also just send the blue tiny tanks to hell and make the yellow ones the norm

ritorizo:
The tank beeping is a good indicator but speaking may be a bit annoying.
stfu

zimon9:
> > na this is a good start, the tiny tanks were completely worthless being used up in like less than 15 mins
> 
> Most people overfill the tanks with air pumps - or purchase an oxygen can to refill them. I don't think making them bigger is smart. Maybe just make it so their full potential is a little more broadcast.

Like I said before, phorid tanks already have a volume of 6 -- all I'm doing is bringing engineering tanks to their level. It feels rather weird to print them out because they're objectively better, instead of using a proper engineering tank. It wouldn't make sense to reduce the volume of phorid tanks either because phorids need the extra volume.

DIB-DOG:
> Like I said before, phorid tanks already have a volume of 6 -- all I'm doing is bringing engineering tanks to their level. It feels rather weird to print them out because they're objectively better, instead of using a proper engineering tank. It wouldn't make sense to reduce the volume of phorid tanks either because phorids need the extra volume.

I’d almost just say remove the phorid tanks from being printable on the autolathe. We get tons of them in emergency air closets and you can buy 2 full ones for 100 credits in the outpost market. Would discourage using them for normal air just based on the inconvenience of emptying them out to refill, as opposed to having the already empty printed one.

zimon9:
> > Like I said before, phorid tanks already have a volume of 6 -- all I'm doing is bringing engineering tanks to their level. It feels rather weird to print them out because they're objectively better, instead of using a proper engineering tank. It wouldn't make sense to reduce the volume of phorid tanks either because phorids need the extra volume.
> 
> I’d almost just say remove the phorid tanks from being printable on the autolathe. We get tons of them in emergency air closets and you can buy 2 full ones for 100 credits in the outpost market. Would discourage using them for normal air just based on the inconvenience of emptying them out to refill, as opposed to having the already empty printed one.

This is reasonable and fair -- a lot of good ideas here and in the rest of the thread, really

zimon9:
Feedback was addressed, and tank volumes were left unchanged. Standard emergency tanks were effectively removed, and the alerts were made a little less chatty -- they should only fire when their pressure becomes perilously low.

code/game/objects/items/tanks/tank_types.dm
```diff
@@ -130,8 +130,8 @@
 
 /obj/item/tank/internals/emergency_oxygen/engi
```
thgvr, line 131: this tank exists and sets `volume` to the same as its parent, same name as well. intended?

zimon9: The name being the same is intended, but I can probably remove the redundant volume variable call. The reason why one wasn't removed outright is to prevent problems in maps and code where the two types of tanks spawn in (e.g., spawned in inventories on bodies, loadouts, oxygen lockers, ectera)

PR #5523, opened 2025-11-04
Die Woke 2: Die Woker (Adds pronouns as something seperate from gender)

Goat-Real:
## About The Pull Request
image

`/mob/living/carbon/human` now defines `pronouns`. This is `NULL` by default. All human specific human pronoun procs in `pronouns.dm` use the new `get_pronouns()` proc if `temp_gender` is false to grab pronouns. The pronoun helper will use the applicable gender of that specific pronoun  if non-null, using `gender` if null. `get_gender()` will use pronouns to determine noun (Male/Female/Person/Thing) but can be changed to a toggle between reading gender or pronouns if enough people want that.

Savefile version is now `43` and will match gender to pronouns on updated savefiles for ease of updating. 

This should probably be test merged seeing how many problems might come up that I could not see. 
## Why It's Good For The Game
~increases shiptest's wokeness by 100%, bring our wokeness up to modern day standards of other codebases.~

Tying pronouns to gender is restrictive due to some player customization options being restricted behind a specific gender. This allows players to pick whatever gender for the character allowing for more player customization. Humans and lizards (as far as I'm aware) also have sexual dimorphism so if people want to have a male body with they/them pronouns or a female body with he/him pronouns this allows for that. More dimorphism with body types could also be allowed as players won't have to worry about having gender force a specific body type. 

Should work fine for things like pirates if pronouns aren't defined it'll just read gender for pronoun usage as before.

:cl: Goat
add: Pronouns. Your pronouns are now separate from your gender. Be as woke as you want. Pronouns are your gender's default when your save file gets updated so don't worry about having to set them. 
/:cl:

### DIFF
--- a/code/__HELPERS/pronouns.dm
+++ b/code/__HELPERS/pronouns.dm
@@ -215,7 +215,33 @@
 		. = "es"
 
 //humans need special handling, because they can have their gender hidden
+/mob/living/carbon/human/proc/get_pronouns()
+	if(pronouns)
+		if(pronouns == "He")
+			return MALE
+		else if(pronouns == "She")
+			return FEMALE
+		else if(pronouns == "It")
+			return NEUTER
+		else
+			return PLURAL
+	return gender
+
+/mob/living/carbon/human/proc/get_default_pronouns()
+	switch(gender)
+		if(MALE)
+			return "He"
+		if(FEMALE)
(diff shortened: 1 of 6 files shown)


thgvr:
this hsould be fine

PR #5525, opened 2025-11-05
fix genemod tails disappearing on wag, adds bunny wag animation

Retrino:
## About The Pull Request
fixes a previous bug where wagging your tail as a human with a bunny or horse tail made your tail vanish. this is because the line of code did not exist for their wag state so they stopped existing. 
also adds a bunny tail wag animation, horses have no such luck yet sorry

https://github.com/user-attachments/assets/36887ad8-ef56-41f7-a49d-caf7d3b32679

## Why It's Good For The Game

less bugs, more content for tail


thgvr:
`can_wag_tail()` check only checks if there is a tail organ present. I hate ite here

PR #5527, opened 2025-11-05
Fixes most random-ass height filter runtimes

Erikafox:
## Changelog

:cl:
fix: height filter runtimes for kepori/vox
/:cl:


thgvr:
this is so fucked this can't be the answer

thgvr:
this makes shit janky and way more difficult, if we're gonna have to do this for like every file we need to revert height

thgvr:
fine for now

PR #5536, opened 2025-11-08
Sets Zohilan as CLIP's official language

SapphicOverload:
## About The Pull Request

Sets Zohilan as CLIP's official language, because I forgot to when adding it to the game.

## Why It's Good For The Game

CLIP's capital and industrial core are former Zohil colonies and should have this as their official language.

## Changelog

:cl:
fix: CLIP now uses Zohilan as its official language
/:cl:


generalthrax:
Believe they're written into lore as Common being their actual official language? They may have their roots in Zohil but they have other planets where Zohilians were not the majority, like Ryuunosuke having Solarian Roots. They put a lot of emphasis on the League itself, not sure I see two member states enforcing that on the rest.

meemofcourse (approved):
will address the league kalixcian issue later

PR #5550, opened 2025-11-12
remes does not cook

Erikafox:
## About The Pull Request

its not supposed to

## Changelog

:cl:
fix: you can no longer cook remes meat
/:cl:


Addust:
itd be funnier to just make it cook into a unique item that tastes like shit if youre from a culture that regularly consumes remes 

pineapple pizza of shiptest

PR #5565, opened 2025-11-15
random ship events actually pick a target

Erikafox:
## Changelog

:cl:
fix: ship targetting random events should actually fire
/:cl:


FalloutFalcon:
I think when I wrote the events, `/datum/event_admin_setup` was ran regradless of wether admins acctually forced events, this meant most randomization was just handled by those. Im trying to figure out why that has since changed....

FalloutFalcon (approved):

PR #5572, opened 2025-11-17
Allows users in hardsuits to equip knives and sledgehammers in the suit storage slot

zimon9:
## About The Pull Request

This adds the list of allowed objects from the global list of allowed objects for armor vests to hardsuits -- this allows hardsuits to hold things like sledgehammers in their suit storage slots now

## Why It's Good For The Game

The allowed list for hardsuits only allowed for a small list of engineering items previously -- this made it so that one was unable to hold a sledgehammer in the suit storage slot of a biege-red hardsuit, whereas a standard armor vest could hold one without issue. This change should bring a bit more parity between the two types of armor

## Changelog

:cl:
balance: updated the allowed list for hardsuits to include things that armor vests can hold in their suit storage slot
/:cl:


thgvr:
suit slot is basically back slot:2 so sure

PR #5573, opened 2025-11-17
Gives mounted widebands a dedicated use key, and prevents intercom use from sending messages through a wideband as well

zimon9:
## About The Pull Request

This pr adds a new use key for widebands, this being :w -- a few modifications were also made so that using the :i key when next to both a wideband and a intercom no longer sends a message through both of them

image

## Why It's Good For The Game

Previously, when you wanted to use _just_ a wideband or _just_ an intercom when multiple were near each other, you'd need to activate the microphone for each one manually. This had its own problems, as you had to ensure that no one else was speaking when the microphone was on, and it was a hassle to do, requiring three separate actions to send one message. This should make it so that widebands are now just as easy to use as an intercom.

## Changelog

:cl:
add: Added a dedicated wideband intercom key
balance: prevents the intercom key from sending messages through widebands
/:cl:


code/game/objects/items/devices/radio/intercom.dm
```diff
@@ -27,7 +27,10 @@ MAPPING_DIRECTIONAL_HELPERS(/obj/item/radio/intercom, 31)
 
 /obj/item/radio/intercom/examine(mob/user)
 	. = ..()
-	. += span_notice("Use [MODE_TOKEN_INTERCOM] when nearby to speak into it.")
+	if(istype(src, /obj/item/radio/intercom/wideband))
```
Erikafox: You could probably make the relevant mode a variable on the intercom.

zimon9: sure thing, I've set up a variable for it now

PR #5585, opened 2025-11-21
Rocks can now be used as cover

zimon9:
## About The Pull Request

This PR gives planetary rocks the passthrough flag so items (like grenades) can be thrown past them, and it gives them a 50% chance to allow a projectile to whizz past them. Just like with sandbags, if you're adjacent to a rock, you can hunker down next to it and fire over/to the side of the rock. 

Smaller rocks can be walked past entirely, but they offer no cover.

| Can be walked through/Provides no Cover | Cannot be walked through/Provides Cover |
|-----------------------------------------|-----------------------------------------|
|image|image|

The base passThrough behavior was moved to structures, so it can be used in the future for custom barricades/cover types.

## Why It's Good For The Game

I thought it would be fun to allow someone to take advantage of natural terrain in firefights -- I've also noticed that some rock sprites are a little small, so being able to walk on top of them might feel nicer.

## Changelog

:cl:
balance: rocks now act like sandbags
/:cl:

### DIFF
--- a/code/game/objects/structures.dm
+++ b/code/game/objects/structures.dm
@@ -19,6 +19,10 @@
 	var/climb_stun = 0
 
 	var/broken = 0 //similar to machinery's stat BROKEN
+	///The chance of a projectile to pass through the structure
+	var/passchance = 0
+	///If projectiles are allowed to pass through a structure at all
+	var/pass_through = FALSE
 
 /obj/structure/Initialize()
 	if (!armor)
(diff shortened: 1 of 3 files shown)


FalloutFalcon (approved):
might be good to make this an element at some point

Aquidu:
do you think you could make the tiny tiny piles of 3 rocks have a higher chance to passthrough so it's worse cover, or just make it provide zero cover at all and always passthrough? those things always get me and they're easy to overlook
Screenshot (4778)

zimon9:
> might be good to make this an element at some point

Alright, I'll try to see if I can do that

zimon9:
> do you think you could make the tiny tiny piles of 3 rocks have a higher chance to passthrough so it's worse cover, or just make it provide zero cover at all and always passthrough? those things always get me and they're easy to overlook Screenshot (4778)

I might just do that and make it so you can step over them, due to how tiny they are. It'll be a little manual, but it can be done yea

rye-rice (commented):
Would you mind moving `var/passchance` to `/obj/structure`? Theres a couple of other structures that could use this var, doing that would mean someone in the future could add values to other structures, even if that is outside the scope of this PR

zimon9:
> Would you mind moving `var/passchance` to `/obj/structure`? Theres a couple of other structures that could use this var, doing that would mean someone in the future could add values to other structures, even if that is outside the scope of this PR

Oh, I wouldn't mind doing that at all - good idea

zimon9:
Ready for review now

PR #5554, opened 2025-11-13
Minor Ruin and Outpost Kitchen Changes

generalthrax:
## About The Pull Request

image

image

image

image

by @DIB-DOG 
image

Also removes titanium requirement from oven + griddle construction

Also fixes some access problems on Beach Gunsmith

## Why It's Good For The Game

newfood. Also don't want titanium buy-in tax for getting kitchen stuff

## Changelog

:cl:
add: Added new kitchen stuff to CLIP moon ruin, Beach gunsmith, Arrowsong, and NT Ice (and NGR rock)
balance: Griddle and oven no longer requires titanium to build
fix: Bad access on Gunsmith
fix: NGR rock no longer uses away mission APCs
/:cl:

PR #5624, opened 2025-11-29
switch building material for griddle and oven

Retrino:
## About The Pull Request

Make the circuitboard to build griddles and ovens require iron instead of titanium

## Why It's Good For The Game

It's a weirdly expensive recipe on top of the thousands to purchase them both anyway


Erikafox:
#5554

Retrino:
💔

PR #5628, opened 2025-11-30
CTF Retheme - CLIP vs Frontiersmen

firebudgy:
## About The Pull Request

Rethemes CTF to be CLIP vs Frontiersmen. Someone else has started a map to go along with this, but I am PRing the outfit changes ahead of time as CTF is currently broken.

Frontiersmen get an SKM while CLIP gets a CM-82, both have standard sidearms (CM-23 and Semi-Auto Mauler) + 1 Frag Grenade.

## Why It's Good For The Game

CTF is ancient and horribly horribly broken (at least in terms of the guns it spawns). This rethemes it to be more- fitting, and also gives people actually working guns so admins don't have to spawn in piles to make it work.

## Changelog

:cl:
balance: CTF Retheme to Frontiersman vs CLIP. Main weapons are now the SKM/CM-82 assault rifles and small sidearms.
balance: You can now be any species in CTF, for better or for worse.
/:cl:


code/modules/awaymissions/capture_the_flag.dm
```diff
@@ -402,109 +401,122 @@
...
+		qdel(src)
+
+/proc/is_ctf_target(atom/target)
+	. = FALSE
+	if(istype(target, /obj/structure/barricade/security/ctf))
+		. = TRUE
+	if(ishuman(target))
+		var/mob/living/carbon/human/H = target
+		if(istype(H.wear_suit, /obj/item/clothing/suit/armor/vest/bulletproof))
+			. = TRUE
```
FalloutFalcon: you removed the code that checks this for damage. I would rename it and remove the barricade check as its unreachable with the new usages. (You will need to find and repalce the other usages)
```suggestion
/proc/is_ctf_player(atom/target)
	. = FALSE
	if(ishuman(target))
		var/mob/living/carbon/human/H = target
		if(istype(H.wear_suit, /obj/item/clothing/suit/armor/vest/bulletproof))
			return TRUE
```

FalloutFalcon (approved):
Verified that ammo pickups still work (oh god the code is so bad it just requips your whole ass outfit, but it works.)

code/modules/awaymissions/capture_the_flag.dm
firebudgy: > you removed the code that checks this for damage. I would rename it and remove the barricade check as its unreachable with the new usages. (You will need to find and repalce the other usages)

Change made, by the by

code/modules/awaymissions/capture_the_flag.dm
```diff
@@ -402,109 +401,134 @@
...
-/obj/projectile/beam/ctf/red
-	icon_state = "laser"
-	impact_effect_type = /obj/effect/temp_visual/impact_effect/red_laser
+/obj/item/gun/ballistic/automatic/pistol/cm23/ctf
+	desc = "CLIP's standard service pistol, chambered in 10mm. This sidearm will disintegrate if dropped."
 
-// BLUE TEAM GUNS
+/obj/item/gun/ballistic/automatic/pistol/cm23/ctf/dropped()
+	. = ..()
+	addtimer(CALLBACK(src, PROC_REF(floor_vanish)), 1)
```
Aquidu: is there any particular reason why we need dropped guns to vanish? last time we played every time my arm got shot my gun was just gone for good, and it's going to be a LOT of extra code when i make classes. also i think the idea of like, picking up an enemy's gun to fire the last of their magazine if you run out of your own ammo is kinda cool, i don't think we'll see a whole lot of "gun-swapping" because whoever picks it up will only have one magazine worth

firebudgy: extended the timers significantly so that you can still pick them up if you drop them on accident, but yes they need to vanish so that the battlefield isn't clogged with dropped guns

code/modules/awaymissions/capture_the_flag.dm
```diff
@@ -402,109 +401,160 @@
...
+/obj/item/gun/ballistic/automatic/assault/skm/ctf
+	desc = "An obsolete model of assault rifle once used by CLIP. This rifle will disintegrate if dropped."
 
-/obj/projectile/bullet/ctf/prehit_pierce(atom/target)
-	if(is_ctf_target(target))
-		damage = 60
-		return PROJECTILE_PIERCE_NONE	/// hey uhh don't hit anyone behind them
+/obj/item/gun/ballistic/automatic/assault/skm/ctf/dropped()
 	. = ..()
+	addtimer(CALLBACK(src, PROC_REF(floor_vanish)), 30)
```
SapphicOverload: This behavior should really be made into an element so it doesn't need to be copied and pasted so many times.

code/modules/awaymissions/capture_the_flag.dm
```diff
@@ -402,109 +401,160 @@
...
+// BLUE TEAM GUNS
+
+/obj/item/gun/ballistic/automatic/assault/cm82/ctf
+	desc = "CLIP's standard assault rifle, a relatively new service weapon. This rifle will disintegrate if dropped."
+
+/obj/item/gun/ballistic/automatic/assault/cm82/dropped()
+	. = ..()
+	addtimer(CALLBACK(src, PROC_REF(floor_vanish)), 30)
+
+/obj/item/gun/ballistic/automatic/assault/cm82/proc/floor_vanish()
```
KOCMODECAHTHUK, line 475: Please fix that, I'm too lazy to do a PR that will be reviewed for a month.
```
/obj/item/gun/ballistic/automatic/assault/cm82/dropped() - > /obj/item/gun/ballistic/automatic/assault/cm82/ctf/dropped()
/obj/item/gun/ballistic/automatic/assault/cm82/proc/floor_vanish() -> /obj/item/gun/ballistic/automatic/assault/cm82/ctf/proc/floor_vanish()
```
CM-82 - Qdeleted. NOT CTF gun.

firebudgy: You are aware this PR's already merged, right? I can't push changes to a merged PR.

PR #5631, opened 2025-12-01
Replaces the use of ambrosia in salads with cabbage

Aquidu:
## About The Pull Request
Replaces Ambrosia Vulgaris and Ambrosia Deus in the crafting recipe of salads with cabbage, and replaces Ambrosia in fridges with cabbage as well.

## Why It's Good For The Game
Last time I was on a Sunskipper and served a salad, it had drugs in it. I think drugging your customers is bad. Ambrosia is eluded to being a drug, so using it for most salads is probably bad. I would've used lettuce, but apparently we don't even have lettuce. No need to have drugs in your fridge if there's nothing to make with them.

## Changelog

:cl:
add: Salad recipies now use cabbage instead of ambrosia. Fridges appropriately modified.
/:cl:


code/game/objects/items/food/salad.dm
```diff
@@ -15,7 +15,7 @@
 
 /obj/item/food/salad/aesirsalad
 	name = "\improper Aesir salad"
```
FalloutFalcon, line 17: Its a salad very specific flavored as a salad with healing properties, I dont see why this one needs the ambrosia removed from it. Golden apples in code have no healing chems so its strange to make THEM whats giving it the panacea somehow.

FalloutFalcon (changes requested):
Other then that its a good change.

code/game/objects/items/food/salad.dm
generalthrax: It might be worth just removing or reflavouring to something that doesn't heal / doesn't include golden apples. Most of the tg salads are barely edible and extremely strange

Aquidu: > It might be worth just removing or reflavouring to something that doesn't heal / doesn't include golden apples. Most of the tg salads are barely edible and extremely strange

I liked the idea of Aesir salad being a expensive/high cost salad that uses "golden apples" for no reason other than vanity, like rich people who eat bread or other foods with legitimate gold shavings on them. Maybe just removing the panacea would be good?

FalloutFalcon (approved):

PR #5633, opened 2025-12-01
Adds Ship Cloaking

SapphicOverload:
## About The Pull Request

Adds ship cloaking devices, allowing a ship to hide from other ships' long-range sensors while active. Consumes 200 watts of stored power for each tile the ship has, roughly 126 kW for a Leyak-class frigate, with a recharge rate of 20 kW per capacitor tier and a maximum charge equal to the installed power cell.

Recharge power draw can be adjusted from the machine itself, and cloaking can be toggled from either the machine or from the ship's helm.

image

When hidden, others on the same tile can still detect the ship's class, but not its designation. It can still be docked to.

image

Currently only the Leyak starts off with cloaking functionality, but it can be purchased from the outpost for 10000 credits (or 7000 for Gorlex Hardliners). A Zohilan version of the cloaking system exists within the code along with their faction but is unobtainable until someone makes a ship for them, or adds it to a ruin.

If you intend to conceal yourself from other ships but don't have access to a cloaking system, hiding in a nebula produces the same effect.

## Why It's Good For The Game

Being able to hide a ship makes chases significantly more interesting, and is a step towards a proper ship-to-ship combat system.

## Changelog

:cl:
add: Added ship cloaking devices.
add: The Leyak-class Covert Frigate is now equipped with a cloaking system.
add: Hiding in a nebula obscures ships from long-range sensors.
/:cl:


thearbiber:
welcome back FTL dodge builds

Addust:
- 126kw of power for a carina-sized ship
im pretty sure even if you maxed the leyak's uranium gen out youd still need to feed it more to get constant cloak, but this also rewards engineering crackheadery since if you have a supermatter you can run a cloak pretty much until the end of time

generalthrax:
15000 might be a little steep for something that probably won't be utilized super often in the hands of players (especially with the power cost), since crew-to-crew player conflict isn't very common (thankfully). Might wanna try and swing for the 8000-10000 price range? It's still very high and likely won't end up on people's shopping lists.

> 126kw of power for a carina-sized ship

I think the high power drain is actually pretty good design choice. Ideally you aren't running that thing constantly only for short bursts on most ships. If you Do want to run it constantly you need the power to back it up

thgvr:
i'd like to have a little bit of testing with this perhaps from an event at least

Aquidu:
ghosts are able to activate/deactivate cloaking

Zevotech:
FRONTIER'S HAUNTED

PR #5640, opened 2025-12-02
fix neoxathin deplete + add overdose effect

Goat-Real:
## About The Pull Request

adds missing `..()` to the end of `on_mob_life()` of nexoathin and gives it an overdose effect where you have a 30% chance to vomit

## Why It's Good For The Game
fixes nexoathin not depleteting and being stuck in the body forever permanently healing toxin

gives nexoathin a proper OD effect where you vomit because this will (most likely) be consumed orally as it's a flora medicine and also used in soups so your body will try to vomit it out (because it felt bad for it to just be a mood debuff and not do anything else when it had a defined overdose amount).
## Changelog

:cl: Goat
fix: nexoathin depletes now
balance: nexoathin makes you vomit now when you overdose on it
/:cl:


code/modules/reagents/chemistry/reagents/medical_reagents/flora_reagents.dm
```diff
@@ -194,3 +194,9 @@
 /datum/reagent/medicine/neoxanthin/on_mob_life(mob/living/carbon/M)
 	if(prob(80))
 		M.adjustToxLoss(-1*REM, 0)
+	..()
+
+/datum/reagent/medicine/neoxathin/overdose_process(mob/living/carbon/M)
```
Erikafox:
```suggestion
/datum/reagent/medicine/neoxanthin/overdose_process(mob/living/carbon/M)
```

PR #5651, opened 2025-12-03
max bodytemp define

Erikafox:
## About The Pull Request

adds a max bodytemp define so you can no longer be the temperature of a small star

## Changelog

:cl:
fix: bodyheat should no longer go over 1000~ whatever units we use for heat
/:cl:


thgvr:
who even thought max temp infinity was a good idea

PR #5656, opened 2025-12-04
Remove Dynamic Missions

Erikafox:
## About The Pull Request
Dynamic missions dilute the exploration element of our game by adding a consistent way to add money while going to places you know. Outpost missions and bounties should be expanded to add more benefits to exploration, rather than having dynamic missions make things more routine.

## Changelog

:cl:
add: Several new cargo missions: Anomaly scans, Class 4 drilling, more creature captures, and dogtag bounties
add: New exports on the outpost main floor: blackboxes, officer tags, documents
add: dogtag loadout item
balance: changed the price of a few exports
del: Dynamic missions have been removed and supplemented by traditional outpost missions
/:cl:


FalloutFalcon:
> while going to places you know

This is def a valid issue with the system, I'd obv recommend reworking it somehow to just not allow meta-gaming, but its just an unfortunate flaw from the explicit intent of the system and I cant really think of a decent route without a massive rework to how they are accepted (likely going with how the punch card system was outlined)

> dynamic missions make things more routine.

So if this is one of the goals of the pr is this, I fail to see how moving the types of missions that existed in it to bounties, or even outpost missions really alleviate the issue of them being routine. I would already consider stuff like scans and drills arguably even more repeated the most of the dynamic missions. (at-least from when I was regularly playing, I'm open to feedback from others on this point) Increasing the amount of outpost missions they could hold would likely only worsen this.

Assuming this is a full removal. `/datum/mission/outpost` can just be convereted back to `/datum/mission` as the only reason it exists is to be a seperation between ruin and outpost missions.

Breakingzap:
You no longer seem to get punchcards from drill missions which is bad.

Erikafox:
you should still get them for rare drill missions, but not for common planet types. Hoping it encourages more exploration.

code/controllers/subsystem/missions.dm
```diff
@@ -1,48 +1,8 @@
 SUBSYSTEM_DEF(missions)
 	name = "Missions"
-	flags = SS_NO_INIT
+	flags = SS_NO_FIRE
```
FalloutFalcon:
```suggestion
	flags = SS_NO_INIT|SS_NO_FIRE
```

code/modules/client/loadout/loadout_accessories.dm
```diff
@@ -184,3 +185,20 @@
 /datum/gear/accessory/pocketprotector
 	display_name = "pocket protector"
 	path = /obj/item/clothing/accessory/pocketprotector
+
+/datum/gear/accessory/dogtag
```
FalloutFalcon, line 195: undocumented

code/modules/cargo/exports/hunting.dm
```diff
@@ -59,6 +59,7 @@
 /datum/export/hunting/trophy/rare
 	cost = 1000
 	unit_name = "rare hunting trophy"
+	elasticity_coeff = 0
```
FalloutFalcon, line 62: undoucmented and something i mostly disagree with as they aren't sufficiently rare enough to not behavior similar to the common ones. (unless mega which are true rewards for really difficult fights for the most part.

code/modules/missions/outpost/industrial_drill.dm
```diff
@@ -0,0 +1,120 @@
...
+/datum/mission/acquire/industrial_drill/New(...)
+	if(!desc)
+		desc = "[SSmissions.get_researcher_name()] has requested that we locate mineral resources for development in the near future. \
+		The only suitable mineral deposits for heavy industry are typically far underground. Locate a vein of sufficent depth, place the provided drill, and let it dig until it's produced a mineral sample. Place the sample in the box and return it to us. \
+		A bonus will be provided for return of the drill."
+	..()
+
+/datum/mission/acquire/industrial_drill/accept(datum/overmap/ship/controlled/acceptor, turf/accept_loc, obj/hangar_crate_spawner/cargo_belt)
+	. = ..()
+	mission_drill = spawn_bound(/obj/machinery/drill/sampler_mission, cargo_belt, VARSET_CALLBACK(src, mission_drill, null))
```
FalloutFalcon: this does not implement the cargo belt logic you put in this pr.

FalloutFalcon (commented):
ill merge or you can self merge it tonight.
these are kinda nits/ i dont care enough.

FalloutFalcon (approved):

FalloutFalcon:
wait why am I "admins"

PR #5657, opened 2025-12-04
Super Sporter Resprite + Indie Hydra

generalthrax:
## About The Pull Request

Sprites by @thearbiber 
image

Decreases Super Sporter fire delay from 4 to 3

Resprites the Super Sporter and rechambers it to 7.62. Indie Hydra is semi-auto, has higher spread, and can only accept the small 20-round and 30-round hydra mags. It can be purchased for 3000 credits.

## Why It's Good For The Game

Super Sporter is a fairly unpopular gun, this might shine it up a little bit. Think it would be interesting to introduce a cheaper rifle from the Scarborough selection that isn't as solid as the rifle proper but can still hold its own

Fire delay so it has some advantages over the Hydra

## Changelog

:cl: thearbiber, generalthrax
add: Added new Super Sporter sprites and Indie Hydra to cargo (3000 credits)
balance: Super Sporter is now chambered in 7.62 and has a fire delay of 3 instead of 4
fix: Super Sporter now has normal Rifle wear
/:cl:


DIB-DOG:
Just popping in as food for thought in regard to the new rifle.

(Aware that the super sporter is only 2500, so maybe my scale is just off)

Putting the indie hydra at 3000 credits places it at the same price point as the indie cobra. Cost over time in regards to ammo is definitely something to consider of course, but when comparing it to other assault rifles which are priced at 5000, it seems like it could fit happily between the two at 4000? 

This is coming purely from me looking at numbers and not having felt/experience the performance difference between the guns, so perhaps the 3000 price tag is warranted, but I wanted to at least bring it up to conversation.

Imaginos16 (approved):
Fwiw the sprites are awesome

code/modules/projectiles/guns/manufacturer/scarborough/ballistics.dm
```diff
@@ -780,6 +780,31 @@ NO_MAG_GUN_HELPER(automatic/marksman/taipan)
 
 NO_MAG_GUN_HELPER(automatic/assault/hydra)
 
+/obj/item/gun/ballistic/automatic/assault/hydra/indie
+	name = "Hydra-80"
+	desc = "A dated variant of Scarborough Arm's premier assault rifle platform. Only accepts small magazines and is locked to semi-auto. Chambered in 5.56mm CLIP."
```
Addust: Scarborough Arms' would probably fit better, its Scarborough Arms not Scarborough Arm

generalthrax: Typo

Overteh:
Hello from the tester team (me), I found a problem, Super Sporter mags doesn't accept cartridges, because they have "7.62x40mm CLIP" in their caliber var, unfortunately (or fortunately?) CLIP is not yet label their cartridges with manufacturer's name, so it simply does not exist, to fix it, just need to remove the " CLIP" to make it will be "7.62x40mm" and then mags will be able to reload.

generalthrax:
Ah shit

PR #5672, opened 2025-12-08
Prevents fake deathgasping

zimon9:
## About The Pull Request

Makes the deathgasp emote unusable as an intentional emote. It still triggers after someone dies naturally.

## Why It's Good For The Game

It's pretty immersion breaking to see someone lying on the floor and deathgasping three times in a row in the slim chance that they can somehow survive and frag montage their way out of dying

## Changelog

:cl:
del: Removed intentional deathgasps
/:cl:


SapphicOverload (changes requested):
The emoting unit test needs to be updated to use a different emote instead of deathgasp that can be done in hard-crit.

zimon9:
> The emoting unit test needs to be updated to use a different emote instead of deathgasp that can be done in hard-crit.

Thank you, I've updated the unit test now. I had been a little hesitant because I was uncertain if I was allowed to modify the unit test, but it should be fine now

FalloutFalcon (approved):

SapphicOverload (approved):

PR #5717, opened 2025-12-18
Goat Horns for humans

thearbiber:
## About The Pull Request
no tail because the bunny tail already fits perfectly for that.
image

## Why It's Good For The Game

Should give some more diversity besides just ear choices? Dont think it strays too far from mammal?

## Changelog

:cl:
add: goat horns for humans
/:cl:


generalthrax:
goated with the s

PR #5718, opened 2025-12-18
Shemagh Masks + Neckwraps

generalthrax:
## About The Pull Request

Sprites  by @thearbiber 

Lets you alt click shemaghs to use them as a mask item instead similar to existing facemasks. This was kind of a bad implementation and I had to move factional shemaghs into the broader shemagh folder so if theres things that could be improved I would appreciate pointers

image

Also adds neckwraps recolourable to loadout

image

## Why It's Good For The Game

Desert fits

## Changelog

:cl: thearbiber, generalthrax
add: You can now alt click shemaghs to use them as a facemask
add: Added recolourable neckwraps to loadout
/:cl:


code/modules/clothing/neck/_neck.dm
```diff
@@ -531,6 +531,53 @@
 	icon_state = "shemagh"
 	supports_variations = VOX_VARIATION
 
+/obj/item/clothing/neck/shemagh/AltClick(mob/user)
```
thgvr, line 534: reskin system?

generalthrax: you asking if the shemaghs themselves should be recolourable or if both the shemaghs and the facemasks should be part of it

code/modules/clothing/neck/_neck.dm
```diff
@@ -531,6 +531,53 @@
...
+
+/obj/item/clothing/mask/shemagh
+	icon = 'icons/obj/clothing/neck.dmi'
+	mob_overlay_icon = 'icons/mob/clothing/neck.dmi'
+	flags_inv = HIDEEARS|HIDEHAIR|HIDEFACIALHAIR
+	clothing_flags = ALLOWINTERNALS
+	flags_cover = MASKCOVERSMOUTH
+	alternate_worn_layer = SUIT_LAYER
+	w_class = WEIGHT_CLASS_TINY
+	var/sourceShemaghType
```
thgvr: `snake_case` please, same for rest of vars you're using here

code/modules/clothing/neck/_neck.dm
thgvr: reskin system to change between items, but I don't think it'd just work out of the box like that

PR #5721, opened 2025-12-19
labcoat/armored duster resprites+black duster

thearbiber:
## About The Pull Request
yeup
image

## Why It's Good For The Game

god they were old+the armored coat had no visible armor+black duster is a cool option

## Changelog

:cl:
add: labcoat/armored duster resprite and black duster
/:cl:


Retrino:
the armored dusters sprites are a bit difficult to read, i would recommend going with a more reinforced look rather than a vest built into the duster. to my understanding the duster itself was meant to be made from heavier armor-adjacent material (forgive my wording its 3 am) instead of a traditional armor vest style. 

the labcoat is a great upgrade to the current, though i don't know if the forced illusion for the back is working quite like intended, assuming its meant to give a slimmer look? the south facing seems to do it a bit better. otherwise they look very nice. the labcoat was really due for a look over

PR #5729, opened 2025-12-21
Alphabetically sorts the orbit menu

zimon9:
## About The Pull Request

This PR makes it so that the orbit menu for alive, dead, and ghost players is sorted alphabetically, as shown in the following screenshot:

image

## Why It's Good For The Game

The current orbit menu is not sorted at all, and it can be a little difficult to quickly select someone to orbit without specifically searching for them. This should fix that issue.

## Changelog

:cl:
add: Adds sorting to the orbit menu
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/Orbit/OrbitContent.tsx
+++ b/tgui/packages/tgui/interfaces/Orbit/OrbitContent.tsx
@@ -25,16 +25,16 @@ export const OrbitContent = (props, context) => {
 
   const sections: readonly ContentSection[] = [
     {
-      content: data.alive,
+      content: data.alive.sort((a, b) => a.name.localeCompare(b.name)),
       title: 'Alive',
       color: 'good',
     },
     {
-      content: data.dead,
+      content: data.dead.sort((a, b) => a.name.localeCompare(b.name)),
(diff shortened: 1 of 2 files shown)


FalloutFalcon (changes requested):
If you do the sorting dm side, you move it to a server side operation once rather then it being sorted for every client (unless im misremembering the react server-client seperation) as well as byond being generally faster then js anyway

zimon9:
> If you do the sorting dm side, you move it to a server side operation once rather then it being sorted for every client (unless im misremembering the react server-client seperation) as well as byond being generally faster then js anyway

Oh, I'm surprised, I thought it was faster the other way around. Generally, I'd like to try and offload work from the server whenever possible, because it's already processing a lot, isn't it?

FalloutFalcon:
> Generally, I'd like to try and offload work from the server whenever possible, because it's already processing a lot, isn't it?

I thought about it for a minute, The server would actually have to sort the list once for every mob interacting with the UI, therefore my point is alot nuller as I was originally thinking it would only run once if it was server side. This is fine. its a pretty small list anyway.

FalloutFalcon (approved):

PR #5734, opened 2025-12-21
Allows water to be used to reduce disgust

zimon9:
## About The Pull Request

Much in the same way as sol dry, this PR makes it so that a glass of water can be used to quell nausea

## Why It's Good For The Game

I'm not totally sure if this should be as strong as sol dry, but I felt like it made for fun flavor for a cup of water to help with queasiness and all

## Changelog

:cl:
add: Added the ability for water to be used to reduce disgust levels
/:cl:


Erikafox:
make it weaker and we're good

zimon9:
> make it weaker and we're good

Alright, I've brought it down to reduce disgust by two instead of five now

PR #5736, opened 2025-12-21
Allows wigs to be worn in the neck slot

zimon9:
## About The Pull Request

This PR adds a slot flag for wigs so they can be worn in the neck slot, allowing for the capability to wear head items along with wigs. I plan on implementing a system to allow for hats to be placed on wigs, and for wigs to become an optional attachment, but this change should allow for a little bit of flexibility while I work on that.

dreamseeker_PEOFPfDJ1c

## Why It's Good For The Game

Having people with wigs be able to use head items along with their wigs allows for more fashion variability, and I think having more fashion options is great

## Changelog

:cl:
add: Added the ability for wigs be to equipped in the neck slot
/:cl:


EmptyLullaby:
my power grows

code/modules/clothing/head/misc_special.dm
```diff
@@ -142,6 +142,7 @@
 	icon_state = "hair_vlong"
```
iamthesilly, line 142: This icon state doesn't exist in the default neck overlay file and is causing the unit tests to fail, you'll have to manually define `mob_overlay_icon` here.

SapphicOverload (changes requested):

code/modules/clothing/head/misc_special.dm
zimon9: Thank you! I've pushed a commit correcting this now

SapphicOverload (approved):

PR #5768, opened 2026-01-01
Dwayne-remap

DIB-DOG:
## About The Pull Request
Remaps the Dwayne-class Long Range Mining Transport.
image
image
image

The armory is shifting from a bajillion flaming arrows to a Volt carbine, Conflagration shotgun, Ohm pistol, and a Candor.

## Why It's Good For The Game
Simultaneously brings the Dwayne in line with other NT vessels while making it enjoyable to play on. 

## Changelog

:cl:
add: remaps the Dwayne
/:cl:

### DIFF
--- a/_maps/configs/independent_mkii_dwayne.json
+++ b/_maps/configs/independent_mkii_dwayne.json
@@ -43,7 +43,7 @@
 		},
 		"Deckhand": {
 			"outfit": "/datum/outfit/job/independent/assistant",
-			"slots": 1
+			"slots": 2
 		}
 	},
 	"enabled": true
--- a/_maps/shuttles/independent/independent_dwayne.dmm
+++ b/_maps/shuttles/independent/independent_dwayne.dmm
@@ -178,8 +178,8 @@
 /area/ship/cargo/port)
 "bR" = (
 /obj/effect/decal/cleanable/dirt,
-/obj/effect/turf_decal/siding/thinplating,
-/obj/effect/turf_decal/borderfloor{
+/obj/effect/turf_decal/borderfloor,
+/obj/effect/turf_decal/siding/thinplating{
 	dir = 1
 	},
 /turf/open/floor/plasteel/patterned,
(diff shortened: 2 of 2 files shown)


thearbiber (dismissed):
* Would remove the windows here for normal walling. Usually find it odd to see a person hop into/out of cryo
image

* Would use a different tile type than grimey here. Tech tiles or mono grey would work well I believe
image

* Think the ship could fit an oven, its got a nice kitchen area. Moving around the cutting board might work to have it line up on the top?
image

dont see much else besides that. Looks pretty good

Randomguy523:
Passive vent on the outside leading into recycling is rough without a pump or anything to control it, leading to it taking in a bunch of cold/hot gas when you might not want it. Could maybe switch it to an active vent that can be switched on and off from the inside, similar to the Falmouth?

DIB-DOG:
> Passive vent on the outside leading into recycling is rough without a pump or anything to control it, leading to it taking in a bunch of cold/hot gas when you might not want it. Could maybe switch it to an active vent that can be switched on and off from the inside, similar to the Falmouth?

Think adding a temperature gate for the feedback would work?

Randomguy523:
> Think adding a temperature gate for the feedback would work?

Haven't messed with temperature gates much so I'm unsure. Not a lot of ships use them over pumps.

DIB-DOG:
> Passive vent on the outside leading into recycling is rough without a pump or anything to control it, leading to it taking in a bunch of cold/hot gas when you might not want it. Could maybe switch it to an active vent that can be switched on and off from the inside, similar to the Falmouth?

Appreciate the idea, spent some time looking through the map and thinking of ideas of how to solve this problem before it came to be pretty easily. Know it’s not ideal but if you absolutely do not want to be drawing overly hot/cold air from the outside you can disable the first filtering pump to achieve that result.

I know it will also stop the on board air from being filtered during that time, but with the reserve mixed air tank you should have more than enough air for any stop, after which point you can re-enable the filter pump.

The space we have to work with in that room sort of constricts how much can be done, so if you have an alternative piping floor plan you’d like to suggest I am more than open to it, but for the time being, the system is serviceable.

PR #5773, opened 2026-01-02
You can only take 1 High Priority Mission

Erikafox:
## About The Pull Request

Makes it so you can only take 1 high priority mission at a time

## Changelog

:cl:
balance: You can only take 1 high priority mission from the outpost at a time
balance: High priority missions pay out more now.
/:cl:


Erikafox:
what the fuck ever it works

code/modules/missions/mission.dm
```diff
@@ -267,8 +267,6 @@
 		"value" = src.value,
 		"progressStr" = get_progress_string(),
 		"progressPer" = get_progress_percent(),
-		"actStr" = act_str
+		"actStr" = act_str,
+		"highPriority" = src.high_priority
```
FalloutFalcon, line 271: idk why these use `src.` There is 0 reason and not even used consistantly here.
(i would suggest a change but it wont let me cause its lines outside your changes)

FalloutFalcon (approved):
You validate the input in ui_act which is good. (never trust tgui's outputs)

PR #5676, opened 2025-12-09
Robot Wounds

SapphicOverload:
## About The Pull Request

This adds several new wound types applicable to robotic limbs, and ports some wound refactoring from /tg/ to do so.

I've also made some additional code changes to allow rolling for and applying multiple wound types at once, and applying slash/pierce wounds with burn damage weapons that have the correct sharpness.

New wound types:
- Buckling - The limb is dented, buckled inward, or sheared, making movement more difficult. Caused by brute damage, more commonly by blunt weaponry.
- Heat Warping - The limb has been warped by thermal stress, reducing the limb's maximum integrity. Caused by burn damage.
- Electrical - The limb's internal electronics have been damaged, causing loss of limb function and damaging internal components. Caused by all damage but especially bullets and lasers.

<details>
<summary>Treatments:</summary>

BUCKLING:

 - Dented Plating (Moderate Buckling) - Apply a crowbar to bend the plating back into place.
 - Buckled Chassis (Severe Buckling) - 1) Cut off the external plating with a welding tool, angle grinder, or plasma cutter. 2) Add new titanium, plasteel, or plastitanium plating. 3) Weld the new plating on with a welding tool.
 - Sheared Frame (Critical Buckling) - 1) Cut off the external plating. 2) Disconnect the internal frame with a wrench. 3) Add new rods to replace the damaged frame. 4) Secure the new frame with a wrench. 5) Add the new plating. 6) Weld the new plating on.

 HEAT-WARPING:

- Surface Oxidization (Moderate Heat-Warping) - Apply a welding tool or similarly hot object to burn off the oxidized layer.
- Warped Plating (Severe Heat-Warping) - Apply a welding tool or other source of high heat to re-heat the plating, then bend it back into shape with a crowbar.
- Deformed Chassis (Critical Heat-Warping) - 1) Cut off the external plating. 2) Disconnect the internal frame with a wrench. 3) Add new rods to replace the damaged frame. 4) Secure the new frame with a wrench. 5) Add the new plating. 6) Weld the new plating on.

ELECTRICAL:

- Damaged Electronics (Severe Electrical) - 1) Unscrew the part's shell. 2) Open the hatch. 3) Replace the internal wiring. 4) Re-heat the solder with a welding tool, or apply liquid solder. 5) Close the hatch. 6) Screw the shell back together.
[...]

PR #5778, opened 2026-01-03
Changes shuttlegibbing to just violently killing someone

zimon9:
## About The Pull Request

This PR modifies shuttlegibbing so that someone that gets crushed by a shuttle doesn't get gibbed, and is instead just given 600 brute damage and severe wounding.

Queen of the road
image

## Why It's Good For The Game

As it is, shuttlegibbing always requires admin intervention, and this change should make it so that admin intervention is no longer required to fix a shuttlegibbing event, so instances of it happening can be kept IC. The consequences for getting crushed should still be severe, with 600 brute being applied, and with wounding being rolled for on every body part.

## Changelog

:cl:
balance: made shuttlegibbing nicer
/:cl:


code/modules/shuttle/on_move.dm
```diff
@@ -39,7 +39,15 @@ All ShuttleMove procs go here
 				if(isanimal(M) || isbasicmob(M))
 					qdel(M)
 				else
-					M.gib()
+					//you're going to get, unequivocally, fucked up
+					M.apply_damage(400, BRUTE, forced = TRUE, spread_damage = TRUE)
+					M.apply_damage(100, BRUTE, BODY_ZONE_CHEST, forced = TRUE)
+					M.apply_damage(100, BRUTE, BODY_ZONE_HEAD, forced = TRUE)
+					var/mob/living/carbon/mob = M
```
west3436: Add a typecheck after this to prevent a runtime happening whenever a non-carbon mob is smashed

zimon9: oh yeah, good point

FalloutFalcon (approved):

code/modules/shuttle/on_move.dm
```diff
@@ -39,7 +39,16 @@ All ShuttleMove procs go here
...
 				else
-					M.gib()
+					//you're going to get, unequivocally, fucked up
+					M.apply_damage(400, BRUTE, forced = TRUE, spread_damage = TRUE)
+					M.apply_damage(100, BRUTE, BODY_ZONE_CHEST, forced = TRUE)
+					M.apply_damage(100, BRUTE, BODY_ZONE_HEAD, forced = TRUE)
+					if(istype(M, /mob/living/carbon))
+						var/mob/living/carbon/mob = M
+						for(var/obj/item/bodypart/limb in mob.bodyparts)
+							limb.check_wounding(WOUND_BLUNT, 50, 50)
```
iamthesilly: `check_wounding` requires a list now as of #5676
```suggestion
							limb.check_wounding(list(WOUND_BLUNT = 50), 50)
```

SapphicOverload (changes requested):

code/modules/shuttle/on_move.dm
zimon9: Ohp, thank you!

SapphicOverload (approved):

PR #5792, opened 2026-01-08
Removes debraining via point-blank gunshots

zimon9:
## About The Pull Request

This PR removes the code responsible for braining someone with a gun, and it instead increases the bullet damage multiplier from 3x to 10x, so most calibers will end up killing or at least incapacitating someone that is executed from point blank.

## Why It's Good For The Game

I kind of dislike it when it's done to other players, or when it's done to simplemobs, but I understand that there may be advocates for it.

## Changelog

:cl:
balance: removes braining
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -816,9 +816,6 @@
 			safety_overlay.icon_state = "[safety_wording]-off"
 		. += safety_overlay
 
-#define BRAINS_BLOWN_THROW_RANGE 2
-#define BRAINS_BLOWN_THROW_SPEED 1
-
 /obj/item/gun/proc/handle_suicide(mob/living/carbon/human/user, mob/living/carbon/human/target, params, bypass_timer)
 	if(!ishuman(user) || !ishuman(target))
 		return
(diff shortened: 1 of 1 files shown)


thgvr:
it's intended to be really bad and drastic, which is why it takes so long. basically I don't think you should be able to easily recover from blowing your brains out or otherwise

zimon9:
> it's intended to be really bad and drastic, which is why it takes so long. basically I don't think you should be able to easily recover from blowing your brains out or otherwise

Hmm... what if it also did a bit of organ damage to the brain? I could also just close this if you'd prefer

thgvr:
i'd like to keep it pretty extreme like this. Imagine if you shoot someone point blank and the head and they just live

zimon9:
> i'd like to keep it pretty extreme like this. Imagine if you shoot someone point blank and the head and they just live

Yeah, I understand -- I actually increased the damage multiplier that the bullet does specifically because of that, and in my testing, the calibers I used did instantly kill the test dummy I used, but it's okay if you'd like this feature to be kept as it is

DIB-DOG:
> i'd like to keep it pretty extreme like this. Imagine if you shoot someone point blank and the head and they just live

If you shoot them with a laser they do.

thgvr:
hahaha yeah

Aquidu:
> > i'd like to keep it pretty extreme like this. Imagine if you shoot someone point blank and the head and they just live
> 
> If you shoot them with a laser they do.

Low calibers, too, or if they're wearing a helmet

PR #5796, opened 2026-01-09
Makes simplemobs not get stuck on railings

Erikafox:
## About The Pull Request
Simplemobs will no longer get stuck bashing platforms and railings they can get over
## Changelog

:cl:
fix: simplemobs will no longer get stuck on railings when they're trying to kill you
/:cl:

### DIFF
--- a/code/game/objects/structures/railings.dm
+++ b/code/game/objects/structures/railings.dm
@@ -105,7 +105,7 @@
 		return . || mover.throwing || mover.movement_type & (FLYING | FLOATING)
 	return TRUE
 
-/obj/structure/platform/CanAStarPass(obj/item/card/id/ID, to_dir, atom/movable/requester)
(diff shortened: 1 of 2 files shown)


code/game/objects/structures/railings.dm
```diff
@@ -105,6 +105,11 @@
 		return . || mover.throwing || mover.movement_type & (FLYING | FLOATING)
 	return TRUE
 
+/obj/structure/platform/CanAStarPass(obj/item/card/id/ID, to_dir, atom/movable/requester)
```
iamthesilly:
```suggestion
/obj/structure/railing/CanAStarPass(obj/item/card/id/ID, to_dir, atom/movable/requester)
```

SapphicOverload (changes requested):

SapphicOverload (approved):

PR #5800, opened 2026-01-10
emergency tank optimizations

zimon9:
## About The Pull Request

This PR adds a status overlay check to prevent tanks from changing their overlay state over and over to the same exact state they're already in -- which seems to be a performance sink. 

I figured that since this may solve the issue, the timer that updates their alert status can be brought down from 5 seconds to 1 second, but please let me know if that's a little too often.

I also fixed a pressure range visual bug that I forgot to put in the previous patch.

## Why It's Good For The Game

This should help reduce lag a little further by removing pointless operations, and it should help the tanks update a little more quickly, which might be a little nicer (as it might be confusing for tanks to stay on the same icon for five seconds even after the pressure is updated)

## Changelog

:cl:
refactor: tank alert code modified to reduce pointless operations, updates more rapidly
/:cl:

### DIFF
--- a/code/game/objects/items/tanks/tanks.dm
+++ b/code/game/objects/items/tanks/tanks.dm
@@ -78,7 +78,7 @@
 
 	START_PROCESSING(SSobj, src)
 
-	addtimer(CALLBACK(src, PROC_REF(pressure_alerts)), 5 SECONDS, TIMER_STOPPABLE|TIMER_LOOP|TIMER_DELETE_ME)
+	addtimer(CALLBACK(src, PROC_REF(pressure_alerts)), 1 SECONDS, TIMER_STOPPABLE|TIMER_LOOP|TIMER_DELETE_ME)
 
 /obj/item/tank/proc/populate_gas()
 	return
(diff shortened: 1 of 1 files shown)


code/game/objects/items/tanks/tanks.dm
```diff
@@ -78,7 +78,7 @@
 
 	START_PROCESSING(SSobj, src)
 
-	addtimer(CALLBACK(src, PROC_REF(pressure_alerts)), 5 SECONDS, TIMER_STOPPABLE|TIMER_LOOP|TIMER_DELETE_ME)
+	addtimer(CALLBACK(src, PROC_REF(pressure_alerts)), 1 SECONDS, TIMER_STOPPABLE|TIMER_LOOP|TIMER_DELETE_ME)
```
iamthesilly: Couldn't `pressure_alerts()` just be called in `process()` instead?

code/game/objects/items/tanks/tanks.dm
```diff
@@ -235,15 +235,20 @@
 	var/pressure = air_contents.return_pressure()
 
 	// Switches the pressure status overlay depending on which range the tank pressure lies in
+	// The extra icon state check prevents the icon state from being changed if it's already set to it
```
iamthesilly, line 233: The checks you added don't actually do anything, `status_overlay_icon_state` is a local variable defined as null just before which means it always just passes them every time. The real culprit is that `update_overlays()` is called every single time `pressure_alerts()` runs even when it doesn't change any alerts.

SapphicOverload (changes requested):

code/game/objects/items/tanks/tanks.dm
zimon9: Oh, I see, thank you! I've changed this behavior now so that update_overlay() only gets called once whenever a pressure alert is done

code/game/objects/items/tanks/tanks.dm
zimon9: Oh, I didn't know I could use that -- thank you! I've swapped the proc over to there now

code/game/objects/items/tanks/tanks.dm
```diff
@@ -330,22 +332,26 @@
 	// Binary variables are used here to prevent an alert from repeating more than once
 	switch(pressure)
 		if((5 * ONE_ATMOSPHERE) to (29 * ONE_ATMOSPHERE))
-			warning_alert = FALSE
-			critical_warning_alert = FALSE
-			empty_alert = FALSE
+			if(!nominal_alert)
+				nominal_alert = TRUE
```
iamthesilly: This isn't set back to false on the other alert levels. Also, I think it would be a lot easier to handle alerts with just one number (0-3) to represent the alert level, instead of having to bother with four separate boolean variables that are mutually exclusive anyway.

SapphicOverload (changes requested):

code/game/objects/items/tanks/tanks.dm
zimon9: oh, you're right, thank you -- I'll swap them over to using numbers instead

zimon9: Review should be addressed now

SapphicOverload (approved):

PR #5820, opened 2026-01-19
replaces roach syndicate headsets with pirate headsets

generalthrax:
## Why It's Good For The Game

this probably wont ever come into play but in the event it does prevents some stupid things from happening

## Changelog

:cl:
fix: roach now has pirate headsets instead of syndicate ones
/:cl:


Erikafox:
rust red headset

PR #5848, opened 2026-01-31
Removes the Spawner Menu Button from the Ghost HUD

firebudgy:
## About The Pull Request

Removes the spawner menu button from the ghost HUD. (It looks offcenter here but it's not, trust me.)
image

## Why It's Good For The Game

We removed ghostroles - and there's none remaining at this point in time, as well as from what I can remember there being no plans to _add_ new ghostroles in the future. Also it looks like shit.

## Changelog

:cl:
del: Ghostrole Spawner UI Button
/:cl:


KOCMODECAHTHUK:
Maybe we should also remove the pAI button?
It's less relevant than the ghost roles button.

firebudgy:
> Maybe we should also remove the pAI button? It's less relevant than the guest roles button.

I didn't immediately because there are, if very few, still pAI players existing. It probably wouldn't be a big loss if it went though.

PR #5853, opened 2026-02-01
Directional Attacks will refrain from self-targetting

Erikafox:
## About The Pull Request
directional attacks will no longer select your own mob as a valid target

## Changelog

:cl:
fix: you can no longer randomly melee yourself with directional melee
/:cl:

### DIFF
--- a/code/datums/elements/directional_attack.dm
+++ b/code/datums/elements/directional_attack.dm
@@ -60,6 +60,8 @@
 
 	var/mob/living/target_mob
 	for(target_mob in turf_to_check)
+		if(target_mob == source)
+			continue
 		if(!target_mob || target_mob.stat == DEAD)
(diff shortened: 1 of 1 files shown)


code/datums/elements/directional_attack.dm
iamthesilly: The clicked atom might be a screen object and have a null location, which causes problems trying to find an angle to it. In this case you can use `parse_caught_click_modifiers()` to get the turf underneath the client's mouse cursor instead.

SapphicOverload (changes requested):
This does solve the issue of hitting yourself, but the fact it's still targeting the user's tile at all means that something else has gone wrong to even get to this point.

SapphicOverload (approved):

PR #5876, opened 2026-02-09
Broadcast Camera/Body Camera improvements/bugfix

zimon9:
## About The Pull Request

Functional changes:
- increases view distance of unwielded broadcast cameras to 7, the same as what a player can see
- Wielded broadcast cameras are given a view range of 10
- Listening range of broadcast cameras was increased to 7
- allows broadcast cameras to be affixed to utility, medical, mining, and security belts
- changes the description of bodycameras so that their in-hand usage is emphasized better
- disallows areas from being set to the current camera location, preventing the server from being locked up due to having to render an extremely large area
- reduces the weight class of broadcast cameras from "bulky" to "normal"
- adds random character initialization for broadcast camera names, to prevent two of them from sharing the same name

Map changes:
- Multitools were removed from maps that previously had them for usage with bodycameras

## Why It's Good For The Game

This PR addresses common complaints/feature requests for bodycameras, and it also makes broadcast cameras more usable. This also prevents cameras from lagging the server out (again). The map changes were done in order to encourage handheld use of the camera. I can revert those if they conflict with the feel the designers of particular maps might have wanted to convey.

## Changelog

:cl:
del: removed multitools from maps if they were on the same tile as bodycameras
balance: increased view and listening distances for bodycameras/broadcast cameras
balance: allows bodycameras and broadcast cameras to be worn in belts
fix: fixed an incorrect area render bug with cameras
/:cl:


generalthrax:
Chat, is this real?

Erikafox:
I don't know how to feel about making the base bodycameras being moved up in range. I like that they can give imperfect information.

zimon9:
> I don't know how to feel about making the base bodycameras being moved up in range. I like that they can give imperfect information.

Oh, alright -- should I bump them down to 6 for their view range, or should I keep it at their base level of 5?

Erikafox:
I'd prefer it stay at 5. Also leaves room for potential upgrades to purchase.

zimon9:
> I'd prefer it stay at 5. Also leaves room for potential upgrades to purchase.

Sounds good to me -- let me push a change for that

code/game/machinery/computer/camera.dm
```diff
@@ -237,7 +237,7 @@
 		var/cam_location = active_camera_B.loc
 
 		// Is the camera in the following items? If so, let it transmit an image as normal
-		if(!(istype(cam_location, /obj/item/clothing/shoes)))
+		if(!((istype(cam_location, /obj/item/clothing/shoes)) || (istype(cam_location, /turf/))))
```
Erikafox:
```suggestion
		if(!((istype(cam_location, /obj/item/clothing/shoes)) || (isturf(cam_location))))
```

code/game/objects/items/bodycamera.dm
```diff
@@ -47,7 +47,10 @@
 	if(in_range(src, user))
 		. += span_notice("The camera is set to a nametag of '<b>[c_tag]</b>'.")
 		. += span_notice("The camera is set to transmit on the '<b>[network[1]]</b>' network.")
-		. += span_notice("It looks like you can modify the camera settings by using it in your hand, or by using a <b>multitool</b> on it.")
+		if(!(istype(src, /obj/item/bodycamera/broadcast_camera)))
```
Erikafox: could you make a var instead of using a typecheck?

code/modules/modular_computers/file_system/programs/secureye.dm
```diff
@@ -175,7 +175,7 @@
 
 	var/cam_location = active_camera.loc
 
-	if(!(istype(cam_location, /obj/item/clothing/shoes)))
+	if(!((istype(cam_location, /obj/item/clothing/shoes)) || (istype(cam_location, /turf/))))
```
Erikafox:
```suggestion
	if(!((istype(cam_location, /obj/item/clothing/shoes)) || (isturf(cam_location))))
```

code/game/objects/items/storage/belt.dm
```diff
@@ -337,6 +344,10 @@
 	STR.can_hold_max_of_items = typecacheof(list(
 		/obj/item/gun = 1,
 	))
+	var/static/list/exception_cache = typecacheof(list(
```
Erikafox: cameras are weight_class_small and can probably be added to the normal holdable list here

code/game/objects/items/storage/belt.dm
```diff
@@ -427,6 +438,10 @@
 		/obj/item/tank/internals/emergency_oxygen,
 		/obj/item/trench_tool,
 		))
+	var/static/list/exception_cache = typecacheof(list(
```
Erikafox: same here

Erikafox (dismissed):

code/game/objects/items/storage/belt.dm
zimon9: That was done specifically so that broadcast cameras can fit in the belt -- it being flavored as hanging _off_ of the belt.

They're too big to fit otherwise, being bulky items.

code/game/objects/items/bodycamera.dm
zimon9: yeah, let me do that

code/game/objects/items/storage/belt.dm
Erikafox: If they're hanging off the belt they should take up the whole belt slot - like a weapon

generalthrax: I think it's semi-reasonable. If we're okay with allowing plasmacutters in toolbelts I see no issue with a rather purposeless item in the gameplay loop utilized only for fun fluff

zimon9:
I've addressed the reviews now -- hopefully this is better

PR #5887, opened 2026-02-12
Various Overmap Fixes and QOL

rye-rice:
## About The Pull Request
Meant to fix the various docking issues ive noticed over the past few uptimes, along with making some player facing docking stuff easier to comprehend. Meant to be a minor PR, you know the old song and dance
In no particular order:
- Makes docking error messages less fucking vauge and weird. This isn't even because they're player facing, this is because i geniuely couldnt figure out wtf they meant without looking up the code. This means an admin might actually figure out what is causing a particular docking issue
- - Some more docking errors for for adjustable docks since those handle docking different from regular docks.
- Fixes improperly mapped docking ports on some ships
- - Firefly docking port is no longer extremely dangerous due to missing `disable_on_owner_ship_dock`
- - June's Docking port now can dock to other ships
- Fixes slimegirl that occurs due to ships sometimes stealing the mobile or even transit docking port from other ships. This should in theory make ship to ship docking a lot less fucked up, but also long rounds (>4 hours) should be a lot more stable
- - Should make docking ports a lot less prone to random breakages and errors
- Adds edgecase handling for a fair bit of ship code
- - Makes Very Big ships have less janky transit zones
- - Checking if a ship can dock to an adjustable dock doesn't actually said dock anymore, saving perf and making adjust. docking port bugs less common
- Quick dock is now Actually Quick. it should search all docking ports on an overmap object (barring outposts, those have their own handling) to find one that you can actually dock to
- Uploading shuttle map templates no longer breaks the overmap sprite due to missing faction datum

Needs TM During an early round due to it touching pretty vital code. Otherwise, it shouldn't be noticable too much.

## Why It's Good For The Game
Working shipcode that works better is good.

## Changelog

:cl:
add: Quick dock now searches for docking ports your ship can land at, rather than picking the first option and failing. Quicker!
add: Makes docking port messages less vauge.
add: Adjusts transit vlevel sizes for larger ships
add: In case you are unable to dock to something, you should be given more feedback
fix: Shuttle errors due to docking ports being stolen
fix: Edgecase handling for docking ports, should fix very rare shuttle bugs
fix: Kali, Corona, Dwayne, Firefly, Scarab, Venetia, Typhoon, June, and especially the Quasar have either proper docking ports or fixed docking ports
fix: Firefly is now not extremely dangerous to dock to
fix: Shuttle code should be less prone to breaking in general

/:cl:


code/modules/admin/verbs/map_template_loadverb.dm
```diff
@@ -73,6 +73,7 @@
 	SSmapping.map_templates[map] = new_map
 	if(template_type == "Shuttle")
 		var/datum/map_template/shuttle/shuttle_template = new_map
+		shuttle_template.faction = new /datum/faction/independent //as factions were moved to map templates from datum/overmap/ship, uploading shuttle templates was broken for a while
```
iamthesilly:
```suggestion
		shuttle_template.faction = SSfactions.factions[/datum/faction/independent] //as factions were moved to map templates from datum/overmap/ship, uploading shuttle templates was broken for a while
```

SapphicOverload (changes requested):

SapphicOverload (approved):

Erikafox (approved):

PR #5907, opened 2026-02-19
Elaborates on IPC wound treatment in health analyzer wound scan

generalthrax:
## About The Pull Request

It was a little vague before. I don't think a player could learn how to fix these wounds without perusing the pr or the code or without a lot of trial and error.

Goes from this
image

To this
image

May attempt to shorten them if necessary but I think knowing all your options is good for treatment

## Why It's Good For The Game

More intuitive way to find out how to treat new IPC wounds in game

## Changelog

:cl:
add: Added more in-depth explanations on how to treat new IPC wounds on health analyzer
/:cl:

### DIFF
--- a/code/datums/wounds/robotic/buckling.dm
+++ b/code/datums/wounds/robotic/buckling.dm
@@ -23,8 +23,8 @@
 
 /datum/wound/blunt/buckling/moderate
 	name = "Bent Plating"
-	desc = "Patient's external plating is bent out of shape. Applying duct tape can temporarily secure the limb until proper repairs."
-	treat_text = "Recommend applying crowbar or wirecutters to bend plating back into place."
+	desc = "Patient's external plating is bent out of shape."
+	treat_text = "Recommend applying crowbar or wirecutters to bend plating back into place. Applying duct tape can temporarily secure the limb until proper repairs."
 	examine_desc = "is bent out of shape"
 	occur_text = "suddenly bends out of shape"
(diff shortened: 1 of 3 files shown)


Erikafox:
I think there's a way to make it a little more informative without writing down step by step instructions - flesh wounds try to keep it more vague (I.E. ruptured cavity states: "Bandaging, cauterization, or surgical repair followed by potential transfusion.")

Having the relevant surgeries on the wiki would also alleviate this.

generalthrax:
It works for flesh wounds because they're quite simple. For a compound fracture, the most complicated treatment is the surgical tools. Scalpel -> retractor -> hemostat. Then you apply the obvious tools of either bonesetter, bone gel, or surgical tape.

It's even simpler for weeping avulsions or the highest tier pierce wound. You just bandage and apply cautery or sutures. You don't even need a surgery.

Here, some of the high tier wounds are upwards of eight step processes. It's vague for flesh wounds because they're simple to treat. I'm not sure it translates nearly as much to these. Personally, I came into shiptest wounds after being out of practice with them on tg for a long time. I didn't look at the wiki, I just used the health analyzer and was able to intuit because of their simplicity. These are anything but (which is fine, but I think it should be possible to learn how to fix them without being forced to navigate the wiki) (of which it is not actually present on the wiki)

generalthrax:
Erika came up with a more ideal alternative so this'll go down

PR #5909, opened 2026-02-19
Brings Bear Speed down to Polar Bear Speed

generalthrax:
## Why It's Good For The Game

Feel like these are more supposed to be an ambush mob rather than a runs-you-down-in-the-open mob. Not sure why brown bears would be much faster than Polar bears anyways

## Changelog

:cl:
balance: Brown bears are now the same speed as polar bears
/:cl:


thgvr:
u can take the var off polar if its on base type

FalloutFalcon:
bears are fast as fuck irl. This is a good change tho. They should fulfill the same role as goliaths for the most part.

I think the speed difference tho just comes from there original TG differences.
Polar bears were mining mobs while normal bears were like random station bullshit.

PR #5936, opened 2026-02-25
Proper outpost baseturfs

### LINKED ISSUE #2526: Digging up the new Nanotrasen Outpost snow tiles creates space tiles
Issue reported from  Round ID: 2033 (Shiptest)

If you dig up the snow tiles it creates space
Pickaxing seems to create icy rock that drains air
image
image

rye-rice:
## About The Pull Request

outposts now have proper baseturfs minus Trifuge, since that would require actual map changes and i just want to ping someone and have this merged

resolves #2526 

## Why It's Good For The Game

Any kind of situation where a turf on an outpost gets destroyed for any reasons reveals that every outpost doesn't have a proper baseturf set up

## Changelog

:cl:
fix: Outposts now have proper baseturfs
/:cl:

### DIFF
--- a/code/modules/overmap/objects/outpost/outpost_types.dm
+++ b/code/modules/overmap/objects/outpost/outpost_types.dm
@@ -225,6 +225,19 @@
 	)
 	faction = /datum/faction/nt
 
+	main_level_ztraits = list(
+		ZTRAIT_STATION = TRUE,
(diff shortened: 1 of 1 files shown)


FalloutFalcon:
This prob fixes #2526 which is ancient but i somehow still remember making it.

PR #5941, opened 2026-02-28
Increases capacity of fridges

zimon9:
## About The Pull Request

This PR increases the amount of items that fridges can hold from 30 items, same as a standard closet, up to 50 items. It also doubles its dragging slowdown, up from the standard of 1.5 from closets, up to 3.

Special Dependencies: None

## Why It's Good For The Game

Food items and ingredients can end up piling up quite a bit, and oftentimes I've seen a fridge get filled to the brim. I think an increase to 50 items can end up being a nice QOL change. 

The slowdown increase is intended to make it so that fridges don't become an attractive option for moving around items other than food. It would be a little silly to do so, anyway.

## Changelog

:cl:
balance: increases the amount of items fridges can hold from 30 items to 50, and doubles their dragging slowdown
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/secure/freezer.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/freezer.dm
@@ -1,5 +1,7 @@
 /obj/structure/closet/secure_closet/freezer
 	icon_state = "freezer"
+	drag_slowdown = 3
+	storage_capacity = 50
 	var/jones = FALSE
 
 /obj/structure/closet/secure_closet/freezer/Destroy()


Apogee-dev:
Does this apply to freezer crates too (such as blood freezers or the crates food orders arrive in), or only fridge closets?

zimon9:
> Does this apply to freezer crates too (such as blood freezers or the crates food orders arrive in), or only fridge closets?

Ahh, only fridge closets -- do you believe it would be good for freezer crates to have a bigger capacity as well? I initially avoided it because fridge closets aren't constructable or easily obtainable, aside from spawning with one or finding one in a ruin -- I figured this would mitigate the potential for them to be misused

PR #5942, opened 2026-02-28
Swaps the glove type for the Derecho Foreman

zimon9:
## About The Pull Request

This PR swaps the type of glove that Derecho Foremen wake up with from the red insulated variant to the yellow variant.

Comparison:
| Original (Red) | New (Yellow) |
|----------------|--------------|
|image|image|

Special Dependencies: None

## Why It's Good For The Game

Super duper minor thing, but I just think that the newer yellow glove sprites look a bit better than the red ones. Ultimately not really even a big deal because the yellow gloves are easily accessible enough to be swapped out in-game.

## Changelog

:cl:
code: swapped the type of insulated glove that derecho foremen wake up with
/:cl:


EmptyLullaby:
im free

PR #5965, opened 2026-03-06
Adds command shortwave radios

zimon9:
## About The Pull Request

This PR adds a shortwave radio subtype that has the command mode enabled. This allows the radio to send large, bold text into the chatbox without requiring the use of a megaphone. This subtype was added into the starter outfit of various faction captain outfits. In addition, it can be purchased from the outpost for 750 credits.

A right click interaction was also added to allow the loudness boost to be toggled without having to use the UI. I'm experiencing a minor bug where the right click toggle doesn't work if it's the active hand, but it does work when it's not in the active hand.

Spritework for the radio is by @firebudgy 

image

image

## Why It's Good For The Game

This can prove to be a minor QOL change for captains when they're trying to grab the attention of their crew. It's already possible to use a megaphone to this effect, but this makes it a little easier.

## Changelog

:cl: @firebudgy 
balance: adds command radios and gives them to faction captains
/:cl:


EmptyLullaby:
ZIMON MY LOVE

Erikafox (changes requested):
rather than making this a backpack content, could you subtype the survival box so that command survival boxes spawn with this radio instead of the normal one?

zimon9:
> rather than making this a backpack content, could you subtype the survival box so that command survival boxes spawn with this radio instead of the normal one?

Oh, good idea -- I've swapped them now

Erikafox (approved):

PR #5968, opened 2026-03-06
Adds factional webbings to cargo

zimon9:
## About The Pull Request

This PR adds factional webbings/belts for Inteq, Hardliners, CLIP, and the NGR to the outpost market

## Why It's Good For The Game

I feel like this might be helpful in cases where one may want to purchase more webbings, if they don't have enough or are missing one. Functionally identical to assault belts, these only serve to help faction cohesion.

## Changelog

:cl:
add: Added faction webbings for inteq, hardliners, CLIP, and the NGR to cargo
/:cl:


BaseballDuster:
Do you think you could also add on CLIP webbing if possible in this PR please?

zimon9:
> Do you think you could also add on CLIP webbing if possible in this PR please?

Oh yeah, I should do that

PR #5972, opened 2026-03-06
Fixes Messy Chop Haircut

Imaginos16:
## About The Pull Request
Literal 2 pixel change to make it so the eye is fully visible on the left (uncovered) side.

## Why It's Good For The Game
One dot eyes bad, we have sclera for a reason!!!

## Changelog

:cl: PositiveEntropy
fix: Messy Chop properly shows the full eye.
/:cl:


Retrino:
the bangs were kind of intentional when i made the sprite, i would kind of prefer if there was just a pushed back option like with the other bang alt hairs

PR #6004, opened 2026-03-13
Spellcheck other_reagents.dm

OnlineGirlfriend:
## About The Pull Request

Spellchecks other_reagents.dm.  Also a little formatting.

## Why It's Good For The Game

Spelling and consistent capitalization scheme = good aesthetics.

## Changelog

:cl:
spellcheck: other_reagents.dm
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -1431,7 +1431,7 @@
 
 /datum/reagent/stable_plasma
 	name = "Stable Plasma"
-	description = "Non-flammable plasma locked in a liquid form that cannot ignite or become gaseous/solid."
+	description = "Non-flammable plasma locked in a liquid form that cannot ignite or change states."
 	reagent_state = LIQUID
 	color = "#2D2D2D"
 	taste_description = "bitterness"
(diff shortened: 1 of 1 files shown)


code/modules/reagents/chemistry/reagents/other_reagents.dm
```diff
@@ -1887,10 +1887,10 @@
 	name = "Romerol"
 	// the REAL zombie powder
 	description = "Romerol is a highly experimental bioterror agent \
-		which causes dormant nodules to be etched into the grey matter of \
+		that causes dormant nodules to be etched into the grey matter of \
 		the subject. These nodules only become active upon death of the \
-		host, upon which, the secondary structures activate and take control \
-		of the host body."
+		host, upon which time the secondary structures activate and take \
```
Erikafox:
```suggestion
		host, upon which, the secondary structures activate and take \
```

OnlineGirlfriend: This isn't correct; there should not be a comma after the "which." Another possible alternative is "These nodules only become active upon the death of the host, when the secondary structures activate and take control." But I would not leave it as-is

Erikafox: Upon which makes more sense than upon which time so remove the comma if you want

OnlineGirlfriend: It doesn’t, it already says “upon the death of the host,” following that with “upon which” is redundant unless “upon which” takes a new object

PR #6014, opened 2026-03-15
Fix examine text pronouns for low blood

AnomalousPotato:
## About The Pull Request
Adds proper pronoun use for the following:
Changes "they has look a little pale" to "they look a little pale" in the examine text if blood level is slightly low.
Changes "they look like they is going to faint" to "they look like they are going to faint"

## Why It's Good For The Game
SPaG good

## Changelog
:cl:
spellcheck: No longer will you "has look a little pale" in your examine text
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -256,7 +256,7 @@
 		apparent_blood_volume -= 150 // enough to knock you down one tier
 	switch(apparent_blood_volume)
 		if(BLOOD_VOLUME_OKAY to BLOOD_VOLUME_SAFE)
-			msg += "[t_He] [t_has] looks a little pale.\n"
+			msg += "[t_He] looks a little pale.\n"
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -256,7 +256,7 @@
 		apparent_blood_volume -= 150 // enough to knock you down one tier
 	switch(apparent_blood_volume)
 		if(BLOOD_VOLUME_OKAY to BLOOD_VOLUME_SAFE)
-			msg += "[t_He] [t_has] looks a little pale.\n"
+			msg += "[t_He] looks a little pale.\n"
 		if(BLOOD_VOLUME_BAD to BLOOD_VOLUME_OKAY)
 			msg += "<b>[t_He] look[p_s()] like [t_he] is going to faint.</b>\n"
```
Erikafox: could you catch this line too? 
can read as "They look like they is going to faint"

AnomalousPotato: Ye

PR #6015, opened 2026-03-15
Suit storage fixes

AnomalousPotato:
## About The Pull Request
- When unequipping your outer suit, you will no longer drop the item in your suit storage if it doesn't need an outer suit to fit there.
- You can now attempt to strip from/to the suit storage slot even if the person isn't wearing a suit, removing the untouchable magic gun storage.
- Items that don't require a suit to fit in suit storage can now be equipped as long as you have a uniform OR a suit, instead of requiring a uniform.
- Unequipping a uniform will also unequip suit storage if they are not wearing a suit and the item cannot be worn naked.

I have tested attempting to equip and unequip items with and without force suit storage trait, while wearing only uniform, only suit, and uniform and suit. I have also tested all of that on another mob via strip. Everything seems to work fine.

## Why It's Good For The Game
Makes suit storage more consistent 

## Changelog

:cl:
fix: Fixed various weird suit storage interactions
fix: You can now always attempt to strip and equip from suit storage regardless if a suit is equipped
/:cl:


Erikafox (approved):

Erikafox:
worked saturday

PR #6031, opened 2026-03-17
Refactors machinery power code

Ical92:
## About The Pull Request
Refactors machine power code to allow machines to draw from area power and power through cables. Currently, all `/obj/machinery` draw from area power and all `/obj/machinery/power` draw power from cables. This merges the power code and switches between the two power sources with the `power_flags` variable of `/obj/machinery`. There should not be any changes to in-game behavior from this PR, as machines are set to POWER_ALLOW_AREA only and power machines are set to POWER_ALLOW_WIRE only. This PR only adds the functionality for machines to draw from both sources, but does not change any machines to do so.

Current functionality is:
- POWER_ALLOW_AREA: Same functionality as a `/obj/machinery` before
- POWER_ALLOW_WIRE: Same functionality as a `/obj/machinery/power` before
- POWER_ALLOW_AREA | POWER_ALLOW_WIRE: Allows both, preferring a cable if present

This will likely need further testing on live due to modifying power code, and could have some edge cases that I have not caught in my local testing.

image

(Possible behaviors that could be implemented following this PR, though are not implemented in this PR.)

## Why It's Good For The Game
Will allow future PRs to modify machine behavior to allow machines, such as chargers, turrets, airlocks, computers, etc. to be able to draw power from cables. This will make it easier for players to power machines in absence of an APC or enclosed area. It will also allow for turrets used in ruins currently 'wired' and powered by a 1-tile area to be properly powered over cable, letting players have an opportunity to cut power to the turret if the map allows.

## Changelog

:cl:
refactor: refactored machine code to allow modifying power source
/:cl:


code/modules/power/power.dm
```diff
@@ -166,26 +137,87 @@
...
+	// powernet = null
+	var/turf/T = get_turf(src)
+	if(!T)
+		return
+	var/obj/structure/cable/attached_wire = T.get_cable_node()
+	if((power_flags & POWER_ALLOW_WIRE) && attached_wire?.powernet)
+		attached_wire.powernet.add_machine(src)
+		. = TRUE
+	if((power_flags & POWER_ALLOW_AREA) && get_area(src))
+		disconnect_from_network()
```
FalloutFalcon, line 162: a little curious why we only disconnect if we have the area power flag, It seems to me like you would want to refresh/disconnect your connects regardless here? I am just asking, I haven't read up on power code in a while.

FalloutFalcon (approved):

code/modules/power/power.dm
Ical92: At that point in the code it means that the machine is either: Not on a wire anymore, not allowed to draw power anymore, or that the wire doesn't have a powernet (for whatever reason). So, this is just to make sure that we're disconnected from the wire we were previously connected to so we aren't drawing from the wire *and* the area power.

Erikafox:
vDe77MYO6s
believe related

Erikafox:
Seems to have broken solars (more)

FalloutFalcon:
> Seems to have broken solars (more)

has this been reproducible or addressed? otherwise pr is good.

Ical92:
> has this been reproducible or addressed? otherwise pr is good.

Both are addressed now

thearbiber:
PR seems to mess up drills. Players cant turn them on, works fine in AI interact

Ical92:
Should be resolved

generalthrax:
Machines with POWER_ALLOW_WIRE vv'd on it seem to not receive power specifically from wires pointing in diagonal directions anywhere in the powernet. They seem to only work along cardinal directions

Ical92:
> Machines with POWER_ALLOW_WIRE vv'd on it seem to not receive power specifically from wires pointing in diagonal directions anywhere in the powernet. They seem to only work along cardinal directions

@generalthrax 
Wasn't able to reproduce, before though with VV you had to manually call power_change() to update it but now it calls automatically after VV. May have been the issue but not sure if it's specifically with diagonal.

generalthrax:
Weird. We'll see wednesday

zimon9:
During testing, it was found that interaction with drills and gas canisters was not possible when spawned in as a character. When attempting to use these machines, AI Interaction worked, however

Ical92:
> During testing, it was found that interaction with drills and gas canisters was not possible when spawned in as a character. When attempting to use these machines, AI Interaction worked, however

Drills were fixed, canisters were just fixed

In the future if this happens it's likely beecause the machine doesn't have the `INTERACT_MACHINE_UNPOWERED` flag set to true on `interaction_flags_machine`

zimon9:
> > During testing, it was found that interaction with drills and gas canisters was not possible when spawned in as a character. When attempting to use these machines, AI Interaction worked, however
> 
> Drills were fixed, canisters were just fixed
> 
> In the future if this happens it's likely beecause the machine doesn't have the `INTERACT_MACHINE_UNPOWERED` flag set to true on `interaction_flags_machine`

Thank you, and good work

PR #6039, opened 2026-03-19
Adds Vigilitas factional armor and clothing crates.

conk444:
## About The Pull Request
Adds a crate to the armor category with the special Vigilitas branded armor vest and helmet, as well as a crate in security supplies with the Vigilitas security slacks and hat (plus jackboots and gloves). Vigilitas armor crate costs the same as the basic unbranded armor crate (because they have the same armor rating) and the clothing crate costs 150 credits.
I also added a period to a description that didn't have a period at the end of it.

## Why It's Good For The Game
It's flavorful, and there are a few examples I can think of off the top of my head wherein this may be useful:

1. You're the director of a Vigilitas patrol vessel, and two of your lackeys just fell in a pit of lava and burned their clothes off. You can buy new slacks and armor to replace what they lost.
2. You're the director of an N+S/ MW freighter, and you need security detail. Some character from Vigilitas is on loan to your freighter as a deckhand, so you promote them to security detail and make them wear the standard uniform, rather than just having a guy dressed like a deckhand with milsurp armor running around.

## Changelog

:cl:
add: Vigilitas now allows staff to buy replacement security uniforms.
spellcheck: added a period.
/:cl:


conk444:
yay

PR #6063, opened 2026-03-24
Balances the Yari

generalthrax:
## About The Pull Request

Doubles the energy cost of disabler + laser so it has 10 shots on a normal cell 20 shots on an upgraded shell (its base)

Also increased the energy cost of its ion so you can only do it twice on a normal four times on an upgraded

Also adds it to Holster

## Why It's Good For The Game

Gun is too good to map onto roundstart ships

## Changelog

:cl:
balance: Nerfed energy cost of Yari projectiles
fix: Yari can now fit in either belt or shoulder holster
/:cl:

### DIFF
--- a/code/modules/projectiles/ammunition/energy/stun.dm
+++ b/code/modules/projectiles/ammunition/energy/stun.dm
@@ -36,8 +36,8 @@
 	e_cost = 625
 
 /obj/item/ammo_casing/energy/disabler/sharplite/hos
-	e_cost = 500
-
+	e_cost = 1000
+g
 /obj/item/ammo_casing/energy/disabler/scatter	//WS edit, scatter repathing
 	pellets = 3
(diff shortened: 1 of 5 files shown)


code/modules/projectiles/ammunition/energy/stun.dm
```diff
@@ -36,8 +36,8 @@
 	e_cost = 625
 
 /obj/item/ammo_casing/energy/disabler/sharplite/hos
-	e_cost = 500
-
+	e_cost = 1000
+g
```
Erikafox:
```suggestion

```

Imaginos16:
Could you maybe add it to the shoulder holster? I'm tired of having to store it in my backpack...

generalthrax:
I got you

PR #6081, opened 2026-03-27
Further increases loadout item slots from 20 to 30

SkyClearingFowl:
## About The Pull Request

Increases the number of loadout item slots that you can take from 20 to 30.

## Why It's Good For The Game

Having more loadout slots can allow a player to have greater diversity in clothing and fluff for more character expression. Personally, many of my characters have a wide diversity of outfits, but I've noticed that 20 slots wasn't enough to portray more than two or three of them at a time. 30 slots should round it out comfortably.
 
## Changelog

:cl:
balance: increases loadout item slots from 20 to 30
/:cl:

### DIFF
--- a/code/controllers/configuration/entries/game_options.dm
+++ b/code/controllers/configuration/entries/game_options.dm
@@ -131,7 +131,7 @@
 	value_mode = VALUE_MODE_FLAG
 
 /datum/config_entry/number/max_loadout_items	//maximum number of items that can be in a player's loadout
-	config_entry_value = 20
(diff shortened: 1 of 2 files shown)


Erikafox (changes requested):
We are not increasing loadout size further without adding an alternative method for loadout to spawn that isn't a box in a bag.

Also I'm not really sold on increasing it further at all.

SkyClearingFowl:
Got it, those are fair points. I'll close this, thank you for the review

PR #6090, opened 2026-03-29
Hail optimizations

Gristlebee:
## About The Pull Request

Hails show who sent them, and dont repeat if you're docked to the same overmap object

## Why It's Good For The Game

Useful info to know who's hailing you and less chat spam

## Changelog

:cl:
add: Incoming hails display ship name.
add: hails dont repeat if you're docked to the same overmap location


Erikafox:
while you're mucking with this do you think you could add a check so that it doesn't repeat a hail if you're both on the same vz?

moffball19:
image

> IMV Jararaca >> VISV Tornado's Anvil

incoming hails seem to display your own ship's name, this was from the pov of someone on the VISV

Gristlebee:
Fixed. Also ships shouldnt repeat hails if they're on the same planet now.

PR #6098, opened 2026-03-30
Himehabu is no longer less accurate when wielded

generalthrax:
## About The Pull Request

Used to suddenly become less accurate when you held it in two hands

## Why It's Good For The Game

Fixes good

## Changelog

:cl:
fix: Himehabu is no longer less accurate when wielded
/:cl:


EmptyLullaby:
^_^

PR #6103, opened 2026-03-31
Adds CM-82 extended magazines 

firebudgy:
## About The Pull Request

Adds CM-82 magazines to the game, with sprites done by the wonderful @Imaginos16!
magimage
casketmagimage
Extended with a 45-round capacity, and Casket with a 60-round capacity.

Extended is in cargo for 750 credits (with the price of the SKM extended magazine lowered to match, it was a bit much really).

## Why It's Good For The Game

They didn't have any special goodies for their assault rifle, leaving them a bit behind in comparison to pretty much everyone else. This helps mitigate that and gives their minutemen some more encouragement on sticking with the standard service rifle.

## Changelog

:cl: Cloudbreak, Imaginos16
add: New extended magazines for CLIP! 
balance: Extended magazine cargo cost reduced to 750 credits from 1250 credits.
/:cl:


Imaginos16 (approved):
Peak...

Apogee-dev:
Extended is a great add. I worry that the casket mag might end up stepping on the CM-40’s toes a little bit. The only reason the SKM and Hydra have drum and casket magazines available is because they both have SAW variants after all.

firebudgy:
> Extended is a great add. I worry that the casket mag might end up stepping on the CM-40’s toes a little bit. The only reason the SKM and Hydra have drum and casket magazines available is because they both have SAW variants after all.

Maybe, but it's not nearly as much firepower as those can output + only the extended is actually buyable without intervention. It was just something added because one of the sprites looked good as an extra-high capacity magazine.

thgvr:
the tradeoff is you have to load it by hand

IHaveNoOriginality:
Maybe the casket magazine could be medium sized to be in line with the Hydra casket and other drum magazines? 60 rounds you can fit easily in toolboxes, boots, pockets and your bag could be rough and easy to abuse. Just noticed this.

PR #6122, opened 2026-04-03
Camera network refactor

zimon9:
## About The Pull Request

This PR refactors cameras and camera consoles slightly. The "ss13" camera network is gone, and is now replaced with the "ship" camera network. 

Mapped in camera consoles, and structural cameras will default to this ship network, prefixed with a pointer to the ship. 
Example: 
image
Mapped in body cameras are also supposed to default to the ship's network on init, but they don't do this, it seems. I can't quite figure out how to resolve this right now, but it looks like the mobile port is null during mapload, which causes the issue.

If a mapper wishes to change the network for a camera console, structural camera, or body camera, they can set the `mapload_network` variable as they wish. For instance, an input of "banana" for the mapload_network variable will set the network as `[ship_pointer]banana`.

Body camera initialization was changed slightly such that changing the name variable in a map will set the c_tag to this name. If one wishes to change the c_tag only, the name of the resulting body camera will be in the format "body camera - (`c_tag`)".

In addition, camera consoles and bodycameras can resync to a ship's camera network via their respective menus.

## Why It's Good For The Game

I hadn't considered how mappers might use these devices in their maps, and I only recently realized that they're coded in such a way that they prevent certain usecases. This PR should hopefully address this and provide a bit more power for mappers to create the vision they want for their maps. If I miss anything with this refactor, feel free to let me know.

## Changelog

:cl:
add: ability to sync with ship's camera network, bodycam init on mapload
refactor: refactored camera backend to allow for mappers to set multiple custom networks for isolated camera networks within the same ship
/:cl:

### DIFF
--- a/code/game/machinery/camera/camera.dm
+++ b/code/game/machinery/camera/camera.dm
@@ -19,7 +19,7 @@
 	integrity_failure = 0.5
 	var/default_camera_icon = "camera" //the camera's base icon used by update_appearance - icon_state is primarily used for mapping display purposes.
 	var/list/network = list("ship")
-	var/mapload_network
+	var/mapload_network = "ship"
 	var/c_tag = null
 	var/status = TRUE
 	var/start_active = FALSE //If it ignores the random chance to start broken on round start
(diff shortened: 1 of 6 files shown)


Erikafox (changes requested):
Did you try using connect_to_shuttle for camera network init stuff?

PR #6127, opened 2026-04-04
CLIP industrial hardsuit

thearbiber:
## About The Pull Request

image

## Why It's Good For The Game
Please god civilian clip content.

## Changelog

:cl:
add: CLIP industrial hardsuit
/:cl:


Apogee-dev:
Big fan of this design. The highlighter green stripes go incredibly well with CLIP grey-blue.

DIB-DOG:
Not sure how it compares to the indie mining hardsuit so take this with a grain of salt.

It’s not like 20 bullet/laser armor is gonna do much, but it might be worthwhile to drop the values down to 10 or so? 

Again I might be totally wrong and if this is comparable to the indie mining hardsuit then just ignore me.

thearbiber:
Its identical to the indie mining in everything except rad and slowdown.

Apogee-dev:
The slowdown should probably just be the same as the normal mining suit, shouldn’t it? It’s not like it’s a sleek lightweight combat suit or anything, it’s every bit as clunky as the indie one.

thearbiber:
.5 slowdown is pretty standard for anything thats going to get hit without explicit heavy armor/being cheap.

PR #6133, opened 2026-04-05
Pouches fit in Vests

firebudgy:
## About The Pull Request

Pouches now fit in vests, allowing for more flexibility. Adjusts nesting code to account for exception lists as well.

## Why It's Good For The Game

Turns out, even trying to make pouches work for pockets - that space is valuable enough that people tend to stick the same few items in there, with oxygen tanks taking up a mandatory slot.

This should add flexibility to people's loadouts and make them consider a bit more what they do with their vests - and allow for customization. And snack pouches. Everyone needs one of those.

## Changelog

:cl:
balance: Pouches fit in vests now.
code: Storage nesting code now accounts for exception lists, in the specific cases this would be relevant.
/:cl:

### DIFF
--- a/code/datums/components/storage/concrete/pockets.dm
+++ b/code/datums/components/storage/concrete/pockets.dm
@@ -30,6 +30,13 @@
 	quickdraw = FALSE
 	silent = FALSE
 
+/datum/component/storage/concrete/pockets/exo/Initialize()
+	. = ..()
+	var/static/list/exception_cache = typecacheof(list(
+		/obj/item/storage/pouch
+		))
+	exception_hold = exception_cache
+
 /datum/component/storage/concrete/pockets/exo/large
 	max_items = 3
 
--- a/code/datums/components/storage/storage.dm
+++ b/code/datums/components/storage/storage.dm
@@ -640,7 +640,7 @@
 	if(isitem(host))
 		var/obj/item/IP = host
 		var/datum/component/storage/STR_I = I.GetComponent(/datum/component/storage)
-		if((I.w_class >= IP.w_class) && STR_I && !allow_big_nesting)
+		if((I.w_class >= IP.w_class) && STR_I && !allow_big_nesting && !is_type_in_list(I, exception_hold))
 			if(!stop_messages)
 				to_chat(M, span_warning("[IP] cannot hold [I] as it's a storage item of the same size!"))
 			return FALSE //To prevent the stacking of same sized storage items.


thgvr:
this is almost certainly not a good change

Overteh:
Would be great to make a separate storage slot dedicated for small oxygen tank. Unfortutanelly, don't know the code much to do it myself.

DIB-DOG:
> Would be great to make a separate storage slot dedicated for small oxygen tank. Unfortutanelly, don't know the code much to do it myself.

Would have to disagree somewhat there? Having to sacrifice a slot or decide whether or not you’re gonna bring something is a good thing IMO and makes operations in a no-air environment actually feel bulky.

Besides, air can fit on your belt, suit storage, back, or either of your two pockets. If you can’t make air fit in at least one of those, you should probably re-evaluate how much you’re carrying.

DIB-DOG:
> this is almost certainly not a good change

I was worried at first when I first read it too, but it’s not going into webbing/vests on your belt slot, which could admittedly lead to a pretty crazy nested storage situation.

Since it’s going into your actual armor vest’s storage, and those only have 2 slots total, I don’t think this is too outrageous. It makes sense to be able to attach them to the armor and shouldn’t be too OP IMO.

Overteh:
makes it a storage in 2 clicks away, also have to keep equipment doll open, I think it will be bulky to use

Overteh:
> > Would be great to make a separate storage slot dedicated for small oxygen tank. Unfortutanelly, don't know the code much to do it myself.
> 
> Would have to disagree somewhat there? Having to sacrifice a slot or decide whether or not you’re gonna bring something is a good thing IMO and makes operations in a no-air environment actually feel bulky.
> 
> Besides, air can fit on your belt, suit storage, back, or either of your two pockets. If you can’t make air fit in at least one of those, you should probably re-evaluate how much you’re carrying.

Then I can just sacrifice my flesh and become IPC to have my pocket free. On station builds, you don't need to use oxygen tank and flashlight most of time, they are tools for emergency situations, unlike shiptest where you need them 90% of the time. We have pocket flashlight replacements like rail light and helmet mount, now we just need something for oxygen tanks. I would prefer to have a further discussion not in the comments of this pull request.

thgvr:
what we should be looking at more here is stuff like having oxygen tanks be inserted into hardsuits to function

Apogee-dev:
It wouldn't hurt if oxygen tanks could function out of specific additional storage slots, maybe? Vest/jacket pockets come to mind. Those aren't exactly competitive, as storage space goes.

firebudgy:
> It wouldn't hurt if oxygen tanks could function out of specific additional storage slots, maybe? Vest/jacket pockets come to mind. Those aren't exactly competitive, as storage space goes.

If you can figure out how to do it. Both I and Zimon tried.

thgvr:
something i think could work well is a mask of some sort that accepts an air tank and functions off of that

Erikafox (approved):

PR #6136, opened 2026-04-06
Weapon Cell Price Revisit

DIB-DOG:
Sharplite mini cells (Ohm cells) were dropped in price to 250 credits

**ETHERBOR:**

A premium option in the world of energy weapons, and for good reason. Their cells are more expensive than their competition from sharplite due to their marginally greater capacity and production cost

Civilian Etherbor cells were set in price to 500 credits

**EXTENDED CAPACITY:**

Akin to the drop in price for extended magazines for the SKM/CM-82/Hydra

Extended capacity weapon cells across the board have been dropped in price to 750 credits.

It's important to note that Extended capacity energy weapon cells are comparable in shot count to standard magazines in equivalent ballistic options.

## Why It's Good For The Game
Energy weapons have the notable advantage of not requiring the purchase of ammunition, and for this reason their cells must be more expensive than a ballistic weapon magazine.

However, they also suffer the disadvantage of being unable to be topped off or refilled away from charging infrastructure.

To balance, this change keeps energy cells fairly priced, but makes the aquisition of further cells more reasonable.

## Changelog

:cl:
add: Added sharplite plus cells to the black market
balance: rebalanced energy cell prices
/:cl:


DIB-DOG:
losing my mind, okay one more time

PR #6149, opened 2026-04-10
PGFN Armor

DIB-DOG:
## About The Pull Request
**Game on!**

https://github.com/user-attachments/assets/c326cd33-e6a6-418b-9874-d5bf732ca2aa

Adds some PGFN Variant armors to the PGF Lineup to finally pull in their look. 

This includes a set of armor on level with M10 equipment as well as a matching face mask.

PGFN navywear coats have also recieved a bit of a nerf in the process. The coats have retained their melee armor but have been reduced to 10 from 35 in all other fields. Trading off, the coats do protect the wearers arms.

image
image

image
image

image
image

image
image

## Why It's Good For The Game
Really solidifies the identity of the Navy crew and gives them a thematic light combat equipment option.

## Changelog

:cl:
add: Added new PGFN armor
/:cl:


thgvr:
helmet seems way big, is it not just a recolor of the existing one? that'd probably be better.
the transparent visors here and the missing area on the outfits around the neck don't exactly make a look reminiscent of pgf stuff, as well

but yeah helmet just looks giant, one of the reasons I made the pgf set the way it is is to combat that. especially really silly here on kepori. doesn't feel like it matches the mask here as well

DIB-DOG:
> helmet seems way big, is it not just a recolor of the existing one? that'd probably be better. the transparent visors here and the missing area on the outfits around the neck don't exactly make a look reminiscent of pgf stuff, as well

The helmet is in fact just a recolor of the M10 plus the reflector on the back. The open neck is a normal thing for how the M10 looks. There are no transparent visors, on the Sarathi model, it's wearing a pair of safety goggles that I forgot to remove.


> but yeah helmet just looks giant, one of the reasons I made the pgf set the way it is is to combat that. especially really silly here on kepori. doesn't feel like it matches the mask here as well

The helmets being large is intentional to differentiate them from the marines and give them a notably different look. Taking inspiration from the Navy "talker helmet" (ofc not _**that**_ big)
image

Recolored marine armor wouldn't fit that bill. The "basic" m10 helmets are mentioned to be a generally accepted design galaxy-wide and while the PGF Marine Corps uses a unique design, I don't think that the PGF Navy would find such an investment to be necessary or worth the cost. Navy crew aren't expecting to be directly engaged in a firefight and as such, a sleek tight design isn't necessary. They're more concerned about shrapnel protection within the confines of a ship.

PR #6184, opened 2026-04-19
Retools the Banshee into a Combat Ship

generalthrax:
## About The Pull Request

Lot of style discussion in deadchat so I went and did it

image

Additionally added a Sidewinder to the ship armoury to make up for losing a hammer and the Salvage Ripley

## Why It's Good For The Game

Stylistically not purpose-built for salvage like a number of other salvage ships. Simply has them more set for shooting and killing and so forth

## Changelog

:cl:
balance: Banshee retooled into a combat ship. Removed salvage ripley, one hammer, and added a Sidewinder and replaced wreckers with Troopers
/:cl:

### DIFF
--- a/_maps/configs/hardliners_banshee.json
+++ b/_maps/configs/hardliners_banshee.json
@@ -3,7 +3,7 @@
 	"faction": "/datum/faction/syndicate/hardliners",
 	"prefix": "ISV",
 	"manufacturer": "ISF Spacecraft",
-	"map_name": "Y450-H Banshee-class Salvage Clipper",
+	"map_name": "Y450-H Banshee-class Security Clipper",
 	"namelists": [
 		"GORLEX",
 		"NATURAL_AGGRESSIVE",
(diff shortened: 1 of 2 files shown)


Madtankdriver:
After playing on this vessel for quite some time now, I don't see a really good reason why we should change the Banshee into more of a combative vessel. Currently, apart from the one other Hardliner mining vessel being mapped, which isn't even really associated heavily with Cybersun. It is one of the only current salvaging Hardliner vessel. Basically, you would be removing an already good niche. Want to do combat primarily? Play on the Leyak.

Part of what makes the Banshee work already as it is is that it isn’t really that combat oriented for regular players playing on it. It's a nice, quiet, specialized shuttle for tearing down ruins, and I don't think it'd be a good idea to remove the Ripley or change much at all. Honestly, it already works really well as is! Changing it to be more mainly combative would strip away what makes it unique. It’s a specialized vessel, a refurbished salvaging vessel funded by Cybersun, hence why they get the Ripley.

But for whatever reason, if you do change it, remember that the Ripley is worth a heck of an investment of money due to its tools and utility. Perhaps think about increasing the starting amount of money to 3000+ credits, and or adding another Hardsuit if you indeed want to lean more towards being a combative vessel to offset some losses. Personally, I'd say leave the Banshee alone. 

All in all, the Banshee already does its job well and what it was designed to do! Try it, think about it, and imagine how your new idea of the Banshee could work well enough to replace the current Banshee. In doing so, you may just be removing variety rather than improving it.

generalthrax:
I'm just the messenger from a maptainer commentary on the vessel not being particularly purpose-built for salvage. If the whole team thinks it isn't good this will just get closed and it'll remain as-is. Otherwise I'm sure the other Hardliner ship being produced will take a similar niche

Madtankdriver:
Okie okie :)
Just think about it.

generalthrax:
Maptainer consensus has ended up not being in favour. I will close this now

PR #6217, opened 2026-04-26
Add Sink to Corona

generalthrax:
## About The Pull Request

image

## Changelog

:cl:
fix: Add sink to corona
/:cl:


moffball19:
that spot feels a bit hacky. could you try either a chem sink on the table here or replacing the filing cabinet with a kitchen sink?
image

generalthrax:
Do you think the equipment within could fit within the cabinet?

Ah, it should lovely

generalthrax:
image

Revised. I think the tables are too cluttered to fit a chem table so this will do better

PR #6222, opened 2026-04-27
Adds a cart to the outpost console and other qol

ritorizo:
## About The Pull Request

Adds a cart to the cargo console which allow to remove/add and set the amount of each individual packs.
screenshot-2026-04-27-22-30-25
screenshot-2026-04-27-22-30-12
screenshot-2026-04-27-22-30-00

The description of the packs and individual costs are visible when hovering the pack name and total cost.
Fix the formating of upchage/reduction.
Makes the packs static (optimisation). 

## Why It's Good For The Game

Should make large orders easier.

## Changelog

:cl:
add: The cargo console now has a cart + fix upcharge formating.
fix: the cargo console now updates when it's ship dock/undock.
:cl:


Erikafox (approved):
needs tm

Overteh:
I'm not completely sure, but apparently now the console doesn't understand well when the ship departs from port and continues to show information as if it's still docked at the outpost, you can even make a cargo order, but nothing will arrive on conveyor belt where the ship was

ritorizo:
> I'm not completely sure, but apparently now the console doesn't understand well when the ship departs from port and continues to show information as if it's still docked at the outpost, you can even make a cargo order, but nothing will arrive on conveyor belt where the ship was

the last commits should fix the bug (ty for reporting it)

Ical92:
awa thank you this is what I wanted but could not wrangle tgui to do

PR #6233, opened 2026-05-01
Fixes wading bug

googobbug:
## About The Pull Request

Fixes the bug where you get slowed forever when landing on water while buckled, which did not previously clear your movespeed modifier.

## Why It's Good For The Game

Nobody likes being slow forever.

## Changelog

:cl: Bumtickley00
fix: You are no longer permanently slowed when landing on water.
/:cl:

### DIFF
--- a/code/datums/elements/immerse.dm
+++ b/code/datums/elements/immerse.dm
@@ -131,8 +131,6 @@ GLOBAL_LIST_INIT(immerse_ignored_movable, typecacheof(list(
 	if(isliving(movable))
 		var/mob/living/living_mob = movable
 		living_mob.remove_movespeed_modifier(/datum/movespeed_modifier/wading)
-	if(buckled)
-		return
 	UnregisterSignal(movable, list(
 		COMSIG_MOVETYPE_FLAG_ENABLED,
 		COMSIG_MOVETYPE_FLAG_DISABLED,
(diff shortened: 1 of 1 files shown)


thearbiber:
this seems to cause another movement bug with equipment slowdown

googobbug:
odd. what happens? and how can I trigger it?

thearbiber:
started happening when i tried equipping items while moving towards the end of the round. just increases slowdown cache by 1 thereefor slower Not sure much else on it

Erikafox:
Im pretty sure that's unrelated to wading. Has it occurred again?

code/datums/elements/immerse.dm
```diff
@@ -128,11 +128,11 @@ GLOBAL_LIST_INIT(immerse_ignored_movable, typecacheof(list(
 	var/atom/movable/to_check = buckled || movable
 	if(!(to_check.movement_type & MOVETYPES_NOT_TOUCHING_GROUND) && !movable.throwing)
 		remove_immerse_overlay(movable)
-	if(buckled)
-		return
 	if(isliving(movable))
 		var/mob/living/living_mob = movable
 		living_mob.remove_movespeed_modifier(/datum/movespeed_modifier/wading)
+	if(buckled)
+		return
```
iamthesilly: This check should be removed entirely, the signals still need to be unregistered as well.
```suggestion
```

SapphicOverload (changes requested):

code/datums/elements/immerse.dm
googobbug: fixed

FalloutFalcon: > the signals still need to be unregistered as well.

~~I dont see the signals in question that arent already unregistered? It seems like everything we register gets unregistered?~~
Ah no you mean the return is stopping it from getting unregistered.

PR #6255, opened 2026-05-08
Agressive intent mouse facing fixes

ritorizo:
## About The Pull Request

The player will now always face the mouse when in agressive intent and not just when the keybinds are used to switch.
Players will no longer face the object they are pulling when moving which will prevent them from quickly looking at the object then at the mouse.

## Why It's Good For The Game

bugfix

## Changelog

:cl:
fix: Players now always face the mouse when in agressive mode.
:cl:


thgvr (approved):

Erikafox (changes requested):
The behavior should be that the player is always facing a crate if they're pulling it.

ritorizo:
> The behavior should be that the player is always facing a crate if they're pulling it.

Implemented it, sorry for the delay I didin't look at the pr once I saw the first aprove.

code/modules/mob/living/living.dm
```diff
@@ -350,6 +350,8 @@
 			update_pull_movespeed()
 
 		set_pull_offsets(M, state)
+	else if(AM.density)
```
Erikafox: potential case: open closet

ritorizo: spent some time thinking about it since I didin't want to add a typecheck, now the code will also check if the object has a `drag_slowdown` value.

Erikafox (approved):
gonna tm tomorrow

PR #6258, opened 2026-05-08
reagent scanners will now indicate what an item grinds into and adds the scanner to cargo

ritorizo:
## About The Pull Request

screenshot-2026-05-08-01-41-09

The medical analyzers will now act like a reagent scanner when in chem mode. 
Scanning an item will now indicate what it grinds into.
The results of scans is now handeled in the atom/reagent_scan() proc to make it easier to exend the behavior of reagent scanners.
Allows the content of rations to be put back in the box (ended up in this branch by accident)

## Why It's Good For The Game

The reagent analyze feels redundant and having to craft it everytime we want to do chemistry is tedious.   

## Changelog

:cl:
add: Medical analyzers can now be used as chem scanners. You can also scan an item to know what it grinds into.
fix: The content of ration packs now fits back inside the box.
:cl:


Erikafox:
rather reagent scanners be something you can purchase than adding even more functionality onto a very strong item.

thgvr (changes requested):
they're intentionally separate, aquisition can be changed however (maybe even just kill the crafting recipe - lol crafting menu)

Erikafox:
rather the crafting recipe exists (no reason to remove it.. alternate methods are good), but it should be purchasable.

ritorizo:
Would it be okay if the advanced scanner could be used as a chem scanner ?

Erikafox:
no

PR #6259, opened 2026-05-08
Hardliner Holdout and Buried Shrine removal

Luftkommando:
## About The Pull Request

A Hardliner installation is besieged by a Ramzi force who don't know that they're fighting ghosts.

<details><summary>Ruin Images</summary>
StrongDMM-2026-05-07 20 13 34
StrongDMM-2026-05-07 20 13 35

</details>

## Why It's Good For The Game

Buried shrine is something of a dud in the Lava lineup alongside Biodome (already being removed), and while the lava steps are neat Indiana Jones and the Lost Temple probably isn't making the cut. Holdout aims to provide something with bite but which isn't as grueling as other combat heavy ruins because you can pick your engagements.

## Changelog

:cl:
add: Hardliner Holdout
del: Buried Shrine
fix: Donk! Co. mention in syndicate_base.dm
/:cl:


thearbiber (changes requested):
* Turret will be good enough with just the bags. Table bag combo does not look that good map wise

Image

* Goes for this one also I'd pick one or the other. The hallway this is in also doesnt have an APC

Image

* Barracks is missing an APC+Air alarm. Window in the circled area may make it look more homely even if shuttered.

Image

*  Kitchen missing an air alarm

Image

* Would make the crate here visible from the start of the room. Otherwise its very out of the way and mostly would be found with metaknowledge

Image

* More of a suggestion, perhaps put the breaker in the office so its more of a hazard you need to look around for a solution?

Image

* Think the front turrets should have their range reduced to 10 or 9. They are quite oppressive at the front and while rushing in like a dumbo should be punished its a bit much to even get close to the ruin at all right now.

Image

* Engineering also needs an alarm

Image

* Need to remove the last of buried shrine here

Image

* Would probably dump the access locks on the lockers unless its intended here. Think you could put more misc hardliner gear in that closet also to make it feel more filled, webbing/boots etc.

Image

* Would probably fill this with some spare clothes also 

Image

kaylexis:
you should probably move the button for the sergeant's shutters, it's easily missed since it's behind the sandbags and next to the turrets.

Crystal1Blitz:
Ruin is nice, fun to deal with all the defensive spots but the turret control does not work and it makes it a real pain when you think the turrets are gonna be off but they arent, I got shot 23 times

PR #6265, opened 2026-05-09
[MDB IGNORE] Makes outpost ingame mapsize match the mapped size, allows larger outpost maps, kind of

rye-rice:
## About The Pull Request
This in short, fixes some weird mapzone stuff to prevent some deadspace inbetween the outpost map ending and the border. This means you can have larger outpost maps, in theory. However, one would have to prevent the map from loading if FULL_INIT isnt on to prevent errors when loading on a local build. But it should now load without any bugs.

## Why It's Good For The Game

This has always bugged me and i heard rumors that maybe someone wanted a larger sized outpost.

## Changelog

:cl:
fix: Outpost maps should no longer have space in between their borders and edges.
/:cl:


thearbiber:
thank you so much

rye-rice:
Made this better to resemble static overmap objects better

code/modules/overmap/objects/outpost/outpost.dm
```diff
@@ -185,18 +187,22 @@
...
+	vlevel = SSmapping.create_virtual_level(
 		name,
 		main_level_ztraits,
 		mapzone,
-		QUADRANT_MAP_SIZE,
-		QUADRANT_MAP_SIZE,
+		main_template.width,
+		main_template.height,
 		ALLOCATION_QUADRANT,
 		QUADRANT_MAP_SIZE
```
Erikafox, line 225: The allocation and quadrant size need to actually match the size of the outpost. I suggest you just make the outpost spawn an entire z level for itself so it's not a weird annoying pain in the ass.

code/modules/overmap/objects/outpost/outpost.dm
```diff
@@ -185,18 +187,22 @@
...
 	)
-	vlevel.reserve_margin(QUADRANT_SIZE_BORDER)
+	vlevel.reserve_margin(border_size)
+
+	var/turf/spawn_turf = locate(vlevel.low_x,vlevel.low_y,vlevel.z_value)
 
-	main_template.load(vlevel.get_unreserved_bottom_left_turf())
```
Erikafox, line 227: why not just use this proc?

code/modules/overmap/objects/outpost/outpost.dm
```diff
@@ -185,18 +187,22 @@
 	log_game("[src] [REF(src)] OUTPOST MAP LEVEL INIT")
 	log_shuttle("[src] [REF(src)] OUTPOST MAP LEVEL INIT")
 
-	var/datum/virtual_level/vlevel = SSmapping.create_virtual_level(
+	var/datum/virtual_level/vlevel
+
+	vlevel = SSmapping.create_virtual_level(
```
Erikafox:
```suggestion
	var/datum/virtual_level/vlevel = SSmapping.create_virtual_level(
```

code/modules/overmap/objects/outpost/outpost.dm
```diff
@@ -43,6 +43,8 @@
 	/// The mapzone used by the outpost level and hangars. Using a single mapzone means networked radio messages.
 	var/datum/map_zone/mapzone
 	var/list/datum/hangar_shaft/shaft_datums = list()
+	///The border size to use. It's recommended to set this to 0 if your map is above 121 tiles wide/high, otherwise it may fail to load on a local build.
+	var/border_size = QUADRANT_SIZE_BORDER
```
Erikafox: I don't like this it should have an edge if it's expected to share space. If it's not then w/e

Erikafox (changes requested):
This is jank and weird and it needs to not be so we don't have weird mapgen problems. Just make the outpost explicitly use a whole z level or something.

rye-rice:
> This is jank and weird and it needs to not be so we don't have weird mapgen problems. Just make the outpost explicitly use a whole z level or something.

To be honest, I probably should have said this when review was sent, but I think thats out of scope, it's not that i disagree with you, but this was meant to be a small and quick improvement; it is now is absolutely much better than how outposts are handled ingame rn; if you look at agini's map edge on live you will notice a lot of empty space as the game generates map borders wayyy outside the map's height. This tries to make outposts have parity with how punchcards load maps, hence some of the changes seeming odd

code/modules/overmap/objects/outpost/outpost.dm
rye-rice: get_unreserved_bottom_left_turf() is for getting the most bottom left non-border turf, since we map in the border with the map, we need to get the true bottom left of the vlevel to load the map on

code/modules/overmap/objects/outpost/outpost.dm
rye-rice: I think this is a holdover from the first iteration of the pr, ill change comment

code/modules/overmap/objects/outpost/outpost.dm
rye-rice: As the borders are mapped in with the map, it IS matching the outpost map size

PR #6267, opened 2026-05-09
Purchase Turret Boards from Cargo

generalthrax:
## Why It's Good For The Game

In case you need replacements or want to make more. Still requires a weapon to construct

## Changelog

:cl:
add: Added turret boards to cargo
/:cl:


Erikafox:
Thinking on it I don't think it'd be an awful thing if turrets were sold in the relevant faction market. I have an idea for making them need some ammunition supplied anyways

PR #6286, opened 2026-05-15
removes the OD threshold for Determination, and lets you keep on going while under its effects

zimon9:
## About The Pull Request

This PR removes the OD threshold for determination, and leans more into the "second wind" aspect by allowing a player to do things while past critical threshold (like using an emergency medipen on themselves).

## Why It's Good For The Game

Determination can be obtained by a player getting wounds, and there have been instances where it has caused an overdose just from wound procs. I also think that it can be nice for it to act like "adrenaline", in the sense that someone under its effects can ignore pain enough to get themselves to safety before collapsing properly.

## Changelog

:cl:
balance: gives determination similar characteristics to carfencadrizine
balance: prevents determination OD on wound procs
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -2201,7 +2201,7 @@
 	metabolization_rate = 0.75 * REAGENTS_METABOLISM // 5u (WOUND_DETERMINATION_CRITICAL) will last for ~17 ticks
 	self_consuming = TRUE
 	taste_description = "pure determination"
-	overdose_threshold = 45
+	overdose_threshold = 200
 	process_flags = ALL
 	/// Whether we've had at least WOUND_DETERMINATION_SEVERE (2.5u) of determination at any given time. No damage slowdown immunity or indication we're having a second wind if it's just a single moderate wound
 	var/significant = FALSE
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -2226,10 +2226,7 @@
 	..()
 
 /datum/reagent/determination/on_mob_end_metabolize(mob/living/carbon/M)
-	REMOVE_TRAIT(L, TRAIT_HARDLY_WOUNDED, /datum/reagent/medicine/carfencadrizine)
-	REMOVE_TRAIT(L, TRAIT_PINPOINT_EYES, type)
 	REMOVE_TRAIT(L, TRAIT_NOSOFTCRIT, type)
-	REMOVE_TRAIT(L, TRAIT_NOHARDCRIT, type)
 
 	if(significant)
 		var/stam_crash = 0
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -2226,7 +2226,7 @@
 	..()
 
 /datum/reagent/determination/on_mob_end_metabolize(mob/living/carbon/M)
-	REMOVE_TRAIT(L, TRAIT_NOSOFTCRIT, type)
+	REMOVE_TRAIT(M, TRAIT_NOSOFTCRIT, type)
 
 	if(significant)
 		var/stam_crash = 0
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -2206,6 +2206,13 @@
 	/// Whether we've had at least WOUND_DETERMINATION_SEVERE (2.5u) of determination at any given time. No damage slowdown immunity or indication we're having a second wind if it's just a single moderate wound
 	var/significant = FALSE
 
+/datum/reagent/determination/on_mob_metabolize(mob/living/L)
+	. = ..()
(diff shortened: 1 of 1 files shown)


code/modules/reagents/chemistry/reagents/other_reagents.dm
```diff
@@ -2201,11 +2201,18 @@
...
 	process_flags = ALL
 	/// Whether we've had at least WOUND_DETERMINATION_SEVERE (2.5u) of determination at any given time. No damage slowdown immunity or indication we're having a second wind if it's just a single moderate wound
 	var/significant = FALSE
 
+/datum/reagent/determination/on_mob_metabolize(mob/living/L)
+	. = ..()
+	ADD_TRAIT(L, TRAIT_HARDLY_WOUNDED, /datum/reagent/determination)
+	ADD_TRAIT(L, TRAIT_PINPOINT_EYES, type)
+	ADD_TRAIT(L, TRAIT_NOSOFTCRIT, type)
+	ADD_TRAIT(L, TRAIT_NOHARDCRIT, type)
```
Erikafox: keep hardcrit because it's the body failing to stay up.

code/modules/reagents/chemistry/reagents/other_reagents.dm
```diff
@@ -2201,11 +2201,18 @@
...
-	overdose_threshold = 45
+	overdose_threshold = 200
 	process_flags = ALL
 	/// Whether we've had at least WOUND_DETERMINATION_SEVERE (2.5u) of determination at any given time. No damage slowdown immunity or indication we're having a second wind if it's just a single moderate wound
 	var/significant = FALSE
 
+/datum/reagent/determination/on_mob_metabolize(mob/living/L)
+	. = ..()
+	ADD_TRAIT(L, TRAIT_HARDLY_WOUNDED, /datum/reagent/determination)
+	ADD_TRAIT(L, TRAIT_PINPOINT_EYES, type)
```
Erikafox: this was in carfen from the opioids

code/modules/reagents/chemistry/reagents/other_reagents.dm
```diff
@@ -2201,11 +2201,18 @@
...
 	taste_description = "pure determination"
-	overdose_threshold = 45
+	overdose_threshold = 200
 	process_flags = ALL
 	/// Whether we've had at least WOUND_DETERMINATION_SEVERE (2.5u) of determination at any given time. No damage slowdown immunity or indication we're having a second wind if it's just a single moderate wound
 	var/significant = FALSE
 
+/datum/reagent/determination/on_mob_metabolize(mob/living/L)
+	. = ..()
+	ADD_TRAIT(L, TRAIT_HARDLY_WOUNDED, /datum/reagent/determination)
```
Erikafox: unsure about this because it'll make wounding by rapid fire Weird

Erikafox (changes requested):
The 'combat high' effects of determination are handled via the determined status effect - which currently just makes limping less bad.

code/modules/reagents/chemistry/reagents/other_reagents.dm
```diff
@@ -2201,11 +2201,18 @@
 	metabolization_rate = 0.75 * REAGENTS_METABOLISM // 5u (WOUND_DETERMINATION_CRITICAL) will last for ~17 ticks
 	self_consuming = TRUE
 	taste_description = "pure determination"
-	overdose_threshold = 45
+	overdose_threshold = 200
```
Erikafox: instead of making this a High Number we probably won't hit you can just set it to 0 for 'no overdose'.

zimon9: Thank you, I've set it to zero now. Do you want me to remove the overdose_process define as well?

code/modules/reagents/chemistry/reagents/other_reagents.dm
zimon9: Thank you, I'll remove it

code/modules/reagents/chemistry/reagents/other_reagents.dm
zimon9: yea, it is kind of out of place, thank you

code/modules/reagents/chemistry/reagents/other_reagents.dm
zimon9: good point, I'll remove it

Erikafox:
So my major concern with this is that it increases the time to an effective kill pretty heavily. I don't know if that's going to be healthy for the game.

generalthrax:
Would be worth testing. I can see the merits of a potential higher time to kill after hearing many arguments about the preferred "longer" pace of Shiptest combat rather than engagements ending shortly after the opening.

I doubt people on the brink of death with dozens of wounds and suffering from projectile recoil will be particularly resilient in combat, either

Erikafox:
The way we achieve a longer ttk should not be by tying it into a chemical. It should be in actually altering the underlaying system. If we tie it to this and start making baseline assumptions it'll get weird and tough to balance around very quickly.

zimon9:
> So my major concern with this is that it increases the time to an effective kill pretty heavily. I don't know if that's going to be healthy for the game.

I'm not too sure about this -- in PVP and PVE, any determination that's proc'd will be rendered out of the system relatively quickly -- once it's all gone, someone past crit threshold will be unable to do anything -- meaning in protracted engagements, it'll make combat a bit more interesting. 

If someone instead uses it to bumrush enemies, they'll end up in a worse spot than they started in, I think.

PR #6288, opened 2026-05-15
Increases food disgust thresholds

zimon9:
## About The Pull Request

Increases disliked/toxic food thresholds so a single bite of a disliked/toxic food no longer causes vomiting.

## Why It's Good For The Game

Vomiting after eating one bite of a disliked/toxic food for a species is a tiny bit silly, I think

## Changelog

:cl:
balance: increases disgust thresholds
/:cl:


Erikafox (changes requested):
Given that Rachnids should be extended a similar amnesty if they bite into something shitty, you'll need to change the underlaying logic for handling toxic foods.

zimon9:
> Given that Rachnids should be extended a similar amnesty if they bite into something shitty, you'll need to change the underlaying logic for handling toxic foods.

Sure, gimme a second..

PR #6298, opened 2026-05-19
Mark's Weather Patch

Erikafox:
## About The Pull Request
Makes weather act a little more sane by making the various chances in it use seconds per tick & updating the how it fires

## Changelog

:cl: Mark Suckerberg
code: optimized ssWeather
/:cl:


Gristlebee:
The storm sounds appear to be stacking on each other and getting pretty loud during testing.

PR #6301, opened 2026-05-21
Crying Sun pump settings fix

DIB-DOG:
## About The Pull Request
Changes a pump's roundstart settings from O2 to N2.

## Why It's Good For The Game
It's meant to fill containers with O2, it now sends N2. That's not a good thing. Fixes it.

## Changelog

:cl:
fix: Fixed the Oxygen pumping station to actually send oxygen.
/:cl:

### DIFF
--- a/_maps/shuttles/pgf/pgf_crying_sun.dmm
+++ b/_maps/shuttles/pgf/pgf_crying_sun.dmm
@@ -43,11 +43,10 @@
 	pixel_y = 32
 	},
 /obj/structure/catwalk/over/plated_catwalk/dark,
-/obj/machinery/atmospherics/components/trinary/filter/atmos/o2{
+/obj/machinery/atmospherics/components/trinary/filter/atmos/n2{
 	dir = 4;
 	piping_layer = 2;
-	on = 0;
-	layer = 2.435
+	name = "oxygen pump"
 	},
 /turf/open/floor/plating,
 /area/ship/storage/equip)


_maps/shuttles/pgf/pgf_crying_sun.dmm
```diff
@@ -43,11 +43,10 @@
 	pixel_y = 32
 	},
 /obj/structure/catwalk/over/plated_catwalk/dark,
-/obj/machinery/atmospherics/components/trinary/filter/atmos/o2{
+/obj/machinery/atmospherics/components/trinary/filter/atmos/n2{
 	dir = 4;
 	piping_layer = 2;
-	on = 0;
-	layer = 2.435
+	name = "oxygen pump"
```
Erikafox, line 49: you sure this is right?

DIB-DOG: Thank you, wasn't paying attention. The layer is correct but it should be off. Changing that now.

Erikafox: I was referring more to the N2 pump being named Oxygen Pump. It feels like weird labeling.

DIB-DOG: It is a bit odd, but there is a reason for it.

The filter is placed and labeled in a location where you are able to fill up a canister with gas from the distribution loop. The purpose is to pull nitrogen out of the standard air mix and send/pump the remaining oxygen into the tank.

While it is functionally filtering out nitrogen, its purpose is to pump in oxygen. Hence the name.

PR #6321, opened 2026-05-27
Pixel perfect scaling minor granularization

zimon9:
## About The Pull Request

Adds pixel perfect scaling between 2.1x, 2.2x, 2.3x, 2.4x, and 2.5x the standard viewport size.

## Why It's Good For The Game

This can be helpful for making the most out of your screen without using stretch to fit (which, as I understand, stretches the pixels some?)

## Changelog

:cl:
add: Added a few more pixel perfect scaling options
/:cl:


thgvr:
i don't think values other than .0 and .5 will scale cleanly

zimon9:
> i don't think values other than .0 and .5 will scale cleanly

Oh, when I tested them, they seemed okay to me, but I'm not too familiar with the system really -- if the values between .0 and .5 won't scale nicely though, I can absolutely remove those options

PR #6322, opened 2026-05-27
Goliath 2

miniusAreas:
## About The Pull Request

Redoes the Goliath, in all it's glory. I wasn't happy with it the first time, so consider this a revision.

image

image

Massive changes to the layout. Swaps a engineer out for a miner role, and changes the June up to be a little different. Crew now spawn with gowns.

## Why It's Good For The Game

Makes the goliath less of a nightmare in spots, re-arranges crew to have less overlap in duties, and adds a lot of soul into it.

## Changelog

:cl:
add: Remaps the Goliath to be significantly more robust.
/:cl:


thearbiber (dismissed):
Review. Woaw.

* Think a tank setup/rack would fit better in between these SSU's. Looks kinda empty otherwise

Image

* Is this meant to start with no fuel? 

Image

* Think the normal stairs look better here over the old.

Image

* Not really a fan of hallway lathe here. Think moving it to disposals room might be better?

Image

Image

* Probably fit an intercom under the TV for now so you dont need to listen to livestreams with your handheld

Image

* The pump is being smothered by a locker 

Image

* Make sure the exterior plating here is the /airless variant

Image

* Slight cheating and more of a suggestion, but what about putting a handrail above the wall here on the june so you dont fall over on the holopad? Can name it overhead handrail or something like that 

Image

* Cryo feels a bit empty waking up in it. Some seating or tables down here might help

Image

* Both of these buttons only effect the bridge exterior windows

Image

*  Big blast doors holofields arent wired here. Also a suggestion but I think you could get away with just the two north blast doors.

Image

* Space tile under the junes thruster here

Image

* Hydro tray doesnt fit in the box 

Image

*  No wire under these tiles

Image

* Cables the wrong direction under the security door 

Image

* No APC in this room

Image

Bit long but its mostly just technical stuff. Looks good overall I'd say

PR #6323, opened 2026-05-27
Adds the Gravitational Jaunter

zimon9:
## About The Pull Request

This is a port/adaption of the syndicate experimental jaunter from this PR:
https://github.com/BeeStation/BeeStation-Hornet/pull/9155

image

The Gravitational Jaunter allows a user to jaunt in a bubble of bent space, much like that of an Alcubierre drive. The bubble dissipates upon hitting an object of sufficient density, be that a wall or a person, and causes localized damage to whatever is hit.

The below link will take you to a video of it in action.
https://github.com/user-attachments/assets/a8b742ae-5118-423b-a22b-6e3255ff8f18

I'm open to any changes to this as needed -- be that to make it use an internal battery like the flyssa, to adjusting the structural/mob damage, or the recharge rate.

This PR also adds a proc to check if a teleport target turf is valid -- a dependency for this port.
A few checks to prevent teleports were also removed because teleports in CC are wanted, but people are still unable to teleport out of bounds/past a map border.

## Why It's Good For The Game

This was a requested port from a user that wished to use it as a reward item for going through a ruin -- I also think it's a pretty fun tool overall.

## Changelog

:cl: @PowerfulBacon 
add: Added gravitational jaunter
refactor: refactored some teleportation code
/:cl:


code/game/objects/items/teleportation.dm
```diff
@@ -209,3 +209,125 @@
...
+		else
+			target.apply_damage(25, BRUTE, blocked = armor_block)
+		target.Paralyze(10 * (100 - armor_block) / 100)
+		target.Knockdown(40 * (100 - armor_block) / 100)
+		// Check if we successfully knocked them down
+		if (target.body_position == LYING_DOWN)
+			to_chat(target, span_userdanger("[user] teleports into you, knocking you to the floor with the bluespace wave!"))
+		else
+			to_chat(user, span_userdanger("[target] resists the force of your jaunt's wake, bringing you to stop!"))
+			to_chat(target, span_userdanger("[user] slams into you, falling out of their bluespace jaunt tunnel!"))
```
Erikafox, line 328: gravity-ify your fluff a bit

Erikafox:
So. The way you've written the fluff out probably doesn't work with the implementation of a *teleporter*. You could try using charge logic like the stupid ramzi hammer does, and making a visual effect. Feel free to ping me if you want ideas for the effect.

zimon9:
> So. The way you've written the fluff out probably doesn't work with the implementation of a _teleporter_. You could try using charge logic like the stupid ramzi hammer does, and making a visual effect. Feel free to ping me if you want ideas for the effect.

I'd be down for that -- I'm not very good with fluff really

zimon9:
Putting this on draft while the fluff is worked on

code/game/objects/items/teleportation.dm
```diff
@@ -209,3 +209,125 @@
...
+	//stores the recharge timer id
+	var/recharge_timer
+
+/obj/item/gravitational_jaunter/examine(mob/user)
+	. = ..()
+	. += span_notice("[src] has [charges] out of [max_charges] charges left.")
+	if(recharging)
+		. += span_notice("<b>A small display on the back reads:</b>")
+		var/timeleft = timeleft(recharge_timer)
+		var/loadingbar = num2loadingbar(timeleft/recharge_time, reverse=TRUE)
```
thgvr, line 251: had no idea this proc even existed.... this is quite cool we should probably use it more lol.

code/game/objects/items/teleportation.dm
```diff
@@ -209,3 +209,125 @@
...
+	//Uses of the device left
+	var/charges = 4
+	//The maximum number of stored uses
+	var/max_charges = 4
+	var/minimum_teleport_distance = 6
+	var/maximum_teleport_distance = 8
+	//How far the emergency teleport checks for a safe position
+	var/parallel_teleport_distance = 3
+	//How long it takes to replenish a charge
+	var/recharge_time = 15 SECONDS
```
thgvr, line 239: it could probably be good to make it battery operated instead,maybe?

Erikafox: Having it be anomaly tech is cool enough on its own. I'm unsure it needs to be cell based too.

code/game/objects/items/teleportation.dm
zimon9: Right? It's pretty neat

code/game/objects/items/teleportation.dm
zimon9: I could potentially do both? In that it could use an internal cell and also make use of the current system where an internal capacitor charges up charges -- giving 20 to 30 total jaunts with a maximum jaunt quantity of 4 at a time. I could potentially reduce the charge rate as well, if we take this avenue. My thinking here is that this implementation can imply that a link between gravity and the other three fundamental forces, much like how gravity generators currently do (since they require electricity to operate). I'm good with anything though and, I'm happy to let you both discuss it

PR #6334, opened 2026-05-29
Vox Pidgin language symbol shows up for speakers and doesnt show for non-speakers

Gristlebee:
## About The Pull Request

Title

## Why It's Good For The Game

The current implementation where the language symbol doesn't show for fluent speakers isn't very intuitive in practice. It's fairly easy to mix up what you're speaking when it appears to be plain text in your text box without a symbol. It also makes it consistent with the other languages.

## Changelog

:cl:
add: Vox pidgin symbol shows for people for understand the language.
/:cl:


generalthrax:
I opened a similar pr (https://github.com/shiptest-ss13/Shiptest/pull/6333) but your explanation is better so i'll close mine

PR #6342, opened 2026-06-01
Reflavours starting text and ending screen

zimon9:
## About The Pull Request

This PR changes text to move shiptest's appearance away from that of a station server

| **New** | **Old** |
|:-------:|:-------:|
|image|image|
|image|image|
|image|image|

## Why It's Good For The Game

Part of why I wanted to propose this change is because much of what I'm cutting or adjusting is never changed between rounds, (the gamemode will never not be secret, the station goals will never be done, ect), while the other part of why I wanted to do this is because it moves the appearance of the server further away from that of a station server. It also opens the door for more fun fluff end-of-the-round items to be shown (total money earned, number of explosives that went off during the round, bullets fired, total number of ships and their crews, ect)

## Changelog

:cl:
refactor: changed startup and end of round text to better reflect shiptest
/:cl:


code/controllers/subsystem/ticker.dm
```diff
@@ -277,7 +274,7 @@ SUBSYSTEM_DEF(ticker)
 	round_start_timeofday = world.timeofday
 	SSdbcore.SetRoundStart()
 
-	to_chat(world, span_notice("<B>Welcome to [station_name()], enjoy your stay!</B>"))
+	to_chat(world, span_notice("<B><font size=4>Welcome to Shiptest Pre-Alpha, enjoy your stay!</font></B>"))
```
Erikafox: This should still use station_name so that it doesn't fuck w/ downstreams. I'd suggest briefly refactoring station name to make a more lore compliant name

Erikafox (changes requested):

code/controllers/subsystem/ticker.dm
zimon9: Sounds like a good idea, I'll do that

zimon9: Changes implemented

zimon9: This is an open letter to you or anyone else to suggest prefixes but I think it's good to go

code/__HELPERS/names.dm
```diff
@@ -99,14 +98,10 @@ GLOBAL_VAR(command_name)
...
 		if(2)
 			new_station_name += pick(GLOB.greek_letters)
-		if(3)
-			new_station_name += "\Roman[rand(1,99)]"
 		if(4)
 			new_station_name += pick(GLOB.phonetic_alphabet)
 		if(5)
 			new_station_name += pick(GLOB.numbers_as_words)
-		if(13)
-			new_station_name += pick("13","XIII","Thirteen")
```
FalloutFalcon, line 109: If you remove this then you should remove the thing that acctually can cause it, which would be the
```
		if(istype(holiday, /datum/holiday/friday_thirteenth))
			random = 13
```

strings/station_names.txt
FalloutFalcon, line 1: If we are gonna have 1 entry in this it should just be hardcoded tbh...

FalloutFalcon (changes requested):

code/__HELPERS/names.dm
zimon9: Oh, okay, thank you -- I've removed it all now

strings/station_names.txt
zimon9: Ah well, I did hardcode it at first, but I decided to just change the entries in the string file for it to just the one entry, in case downstreams wanted to use a different string list for their station names

zimon9:
I'm going to go all the way and make the ending screen show interesting statistics, so I'll put this on draft for now

zimon9:
I'm just going to leave this ready as-is for now. I'll add fun statistics in another PR down the line

PR #6348, opened 2026-06-01
Agni Car Battery Fix

goober3:
## About The Pull Request
image

the objects submersed in tar in the maintenance sections of agni appear to be making this check fail:

image

this PR rectifies that, and also relocates the var-edited fluff objects to the outpost's mapdata file.

## Changelog

:cl:
fix: Agni's car battery room has been cleaned up to rectify mysterious check failures.
/:cl:


Erikafox (commented):
I have a feeling that the checks might've been erroring because the objects weren't proper types. Try placing them back in and running a few tests?

PR #6350, opened 2026-06-01
Beach Bunkers Check Fix

goober3:
## About The Pull Request

image

these mice in beach_bunkers.dmm like attempting to open the locked airlocks during checks, which allegedly causes the checks to fail. quote from rye:

> okay so durng ruin spawn during this one specific test, a rat spawns and attempts to open a airlock, however when a ruin is either being cleaned up or spawning, this airlock doesnt have a wire datum

image

i don't know a smart way of fixing this, so the mice have been murdered.

## Changelog

:cl:
fix: The living mice have been removed from beach_bunker.dmm's moonshine storage for crimes of esoteric code check failures.
/:cl:


Erikafox (changes requested):
Try disabling the AI controller during tests rather than removing the mice.

PR #6351, opened 2026-06-01
[DNM] Fixes security HUDs so that they show job icons again

zimon9:
## About The Pull Request

This PR changes the define for security huds so that they're an integer again, as job icons weren't showing up when a string entry was used for them instead

image

## Why It's Good For The Game

I know the solution is kind of terrible, but I tried to dig deep into the problem and I got as far as figuring out that hud_atoms weren't being added when the list entry was a string, but they _were_ being added when the entry was an integer. What's especially confusing is that, during debugging, I found that byond correctly loaded the right hud datum regardless of what the entry was, be it a string or an integer. It's just the fact that hud_atoms aren't being added for this datum unless its entry is given as an integer.

## Changelog

:cl:
fix: fixed hud job icons so that they show up again
/:cl:

### DIFF
--- a/code/__DEFINES/atom_hud.dm
+++ b/code/__DEFINES/atom_hud.dm
@@ -54,7 +54,7 @@
 //data HUD (medhud, sechud) defines
 //Don't forget to update human/New() if you change these!
 #define DATA_HUD_SECURITY_BASIC "sec_basic"
-#define DATA_HUD_SECURITY_ADVANCED "sec_adv"
+#define DATA_HUD_SECURITY_ADVANCED 2
 #define DATA_HUD_MEDICAL_BASIC "med_basic"
 #define DATA_HUD_MEDICAL_ADVANCED "med_adv"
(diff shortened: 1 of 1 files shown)


FalloutFalcon (changes requested):
I dug for a minute as well and cant see why this occurs but i suspect it would be an issue with the other huds if it is an issue. regardless, Instead of having a single entry in here be a number make the whole thing a proper alist and make them all numbers again. Otherwise anything running 516 has issue
(TG
Image
Image
)

from https://github.com/tgstation/tgstation/commit/d84a87f66840894a05049615787200ca94150d7a
```dm
/// Create an alist global with an initializer expression
#define GLOBAL_ALIST_INIT(X, InitValue) GLOBAL_RAW(/alist/##X); GLOBAL_MANAGED(X, InitValue)

// Create an alist global that is initialized as an empty list
#define GLOBAL_ALIST_EMPTY(X) GLOBAL_ALIST_INIT(X, alist())
```

PR #6352, opened 2026-06-02
allows screwdrivers to be autolathable via right click

zimon9:
## About The Pull Request

This PR makes it so that screwdrivers can be autolathed via right clicking on an autolathe with them. Standard deconstruction functionality via left click is retained.

dreamseeker_JSsy6drXwT

## Why It's Good For The Game

Have you or a loved one ever accidentally printed out 50 screwdrivers instead of 50 marker beacons? Now you can undo your mistake instead of stuffing them all in the trash bin. (This also reduces a bit of inconsistency with respect to lathing tools)

## Changelog

:cl:
add: Added ability to reclaim material on screwdrivers via autolathe
/:cl:

### DIFF
--- a/code/game/machinery/autolathe.dm
+++ b/code/game/machinery/autolathe.dm
@@ -268,6 +268,14 @@
 	var/datum/component/material_container/materials = GetComponent(/datum/component/material_container)
 	materials.retrieve_all()
 
+/obj/machinery/autolathe/attackby_secondary(obj/item/O, mob/living/user, params)
+	if(istype(O, /obj/item/screwdriver))
+		var/datum/component/material_container/materials = GetComponent(/datum/component/material_container)
+		materials.OnAttackBy(materials, O, user)
+	else
+		attackby(O, user, params)
+	return SECONDARY_ATTACK_CANCEL_ATTACK_CHAIN
+
 /obj/machinery/autolathe/attackby(obj/item/O, mob/living/user, params)
 	if (busy)
 		to_chat(user, "<span class=\"alert\">The autolathe is busy. Please wait for completion of previous operation.</span>")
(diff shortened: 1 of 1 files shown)


code/game/machinery/autolathe.dm
```diff
@@ -268,6 +268,14 @@
 	var/datum/component/material_container/materials = GetComponent(/datum/component/material_container)
 	materials.retrieve_all()
 
+/obj/machinery/autolathe/attackby_secondary(obj/item/O, mob/living/user, params)
+	if(istype(O, /obj/item/screwdriver))
```
iamthesilly: Not everything with screwdriver functionality is a screwdriver.
```suggestion
	if(O.tool_behaviour == TOOL_SCREWDRIVER)
```

code/game/machinery/autolathe.dm
```diff
@@ -268,6 +268,14 @@
...
+		var/datum/component/material_container/materials = GetComponent(/datum/component/material_container)
+		materials.OnAttackBy(materials, O, user)
+	else
+		attackby(O, user, params)
+	return SECONDARY_ATTACK_CANCEL_ATTACK_CHAIN
```
iamthesilly: You don't need to call the normal attackby here, returning `SECONDARY_ATTACK_CALL_NORMAL` already does that.
```suggestion
		return SECONDARY_ATTACK_CANCEL_ATTACK_CHAIN
	return SECONDARY_ATTACK_CALL_NORMAL
```

iamthesilly (dismissed):

code/game/machinery/autolathe.dm
zimon9: oh, my bad- thank you

code/game/machinery/autolathe.dm
zimon9: Thank you for catching these -- I've signed off on these changes now, thank you!

PR #6355, opened 2026-06-02
Electronic Voicebox quirk

zimon9:
## About The Pull Request

This PR adds a quirk that allows a non-IPC player to spawn in with a robot voicebox instead of their species tongue. It does not give a player the ability to speak Encoded.

A new span class, noticerobotic was made for this pr, and the Insert proc for tongues was modified so that it takes the ``drop_if_replaced`` argument to pass on to its parent proc

## Why It's Good For The Game

This was a requested feature in order to allow for players to make non-robotic characters with robotic voices -- be it due to a botched surgery, excessive smoking, or what have you.

## Changelog

:cl:
add: Added the electronic voicebox quirk
/:cl:

### DIFF
--- a/code/modules/surgery/organs/tongue.dm
+++ b/code/modules/surgery/organs/tongue.dm
@@ -34,7 +34,7 @@
 
 /obj/item/organ/tongue/proc/handle_speech(datum/source, list/speech_args)
 
-/obj/item/organ/tongue/Insert(mob/living/carbon/M, special = 0)
+/obj/item/organ/tongue/Insert(mob/living/carbon/M, special = 0, drop_if_replaced = TRUE)
 	..()
 	if (modifies_speech)
 		RegisterSignal(M, COMSIG_MOB_SAY, PROC_REF(handle_speech))
--- a/code/modules/surgery/organs/tongue.dm
+++ b/code/modules/surgery/organs/tongue.dm
@@ -149,6 +149,10 @@
 	. = ..()
 	languages_possible = languages_possible_robot
 
+/obj/item/organ/tongue/robot/Insert(mob/living/carbon/M, special = 0, drop_if_replaced = TRUE)
+	. = ..()
+	ADD_TRAIT(owner, TRAIT_ELECTRONIC_VOICEBOX, ROUNDSTART_TRAIT)
+
 /obj/item/organ/tongue/robot/emp_act(severity)
 	owner.apply_effect(EFFECT_STUTTER, 120)
 	owner.force_scream()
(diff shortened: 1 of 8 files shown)


NovyNovatic:
Honestly something like this might also be nice the other way around. Not every synthetic is going to sound exactly robotic after all. I kinda see it as more of a character personality thing than anything. I like it!

zimon9:
> Honestly something like this might also be nice the other way around. Not every synthetic is going to sound exactly robotic after all. I kinda see it as more of a character personality thing than anything. I like it!

Not a bad idea -- I'll have to think about its exact implementation but I'll put it on the backburner for now

Erikafox (dismissed):
Please have the voicebox trait apply to other mobs with a voicebox (IPCs)

zimon9:
> Please have the voicebox trait apply to other mobs with a voicebox (IPCs)

If I may ask, what do you mean by this? I restricted this quirk from IPCs because they already come with robot voiceboxes, so their being able to use this quirk would change nothing for them. The intention here was for fleshy vocal cords to be replaced with an implanted voicebox with this quirk

Erikafox:
So - the trait that's added to the mob by the quirk, TRAIT_ELECTRONIC_VOICEBOX, should be added to mobs that have a voicebox but don't have the trait (maybe have it done on the organ?), because later down the line, if someone wants to make something that checks for the trait (I.E. device that scrambles speech for voiceboxes), it'd only apply to people with the quirk, rather than everyone with an electronic voicebox.

zimon9:
> So - the trait that's added to the mob by the quirk, TRAIT_ELECTRONIC_VOICEBOX, should be added to mobs that have a voicebox but don't have the trait (maybe have it done on the organ?), because later down the line, if someone wants to make something that checks for the trait (I.E. device that scrambles speech for voiceboxes), it'd only apply to people with the quirk, rather than everyone with an electronic voicebox.

Oh, I see! Sure thing, I'll absolutely do that

zimon9:
> So - the trait that's added to the mob by the quirk, TRAIT_ELECTRONIC_VOICEBOX, should be added to mobs that have a voicebox but don't have the trait (maybe have it done on the organ?), because later down the line, if someone wants to make something that checks for the trait (I.E. device that scrambles speech for voiceboxes), it'd only apply to people with the quirk, rather than everyone with an electronic voicebox.

Thank you, I've tied the trait to the robot voicebox organ now

code/modules/surgery/organs/tongue.dm
```diff
@@ -149,6 +149,10 @@
 	. = ..()
 	languages_possible = languages_possible_robot
 
+/obj/item/organ/tongue/robot/Insert(mob/living/carbon/M, special = 0, drop_if_replaced = TRUE)
+	. = ..()
+	ADD_TRAIT(owner, TRAIT_ELECTRONIC_VOICEBOX, ROUNDSTART_TRAIT)
```
Erikafox: it's not gonna be a roundstart trait if it gets added midround

zimon9: oops, changed the source to ORGAN_TRAIT now, thank you -- I hope that's right?

PR #6360, opened 2026-06-02
adds salbutamol to the outpost med vender

zimon9:
## About The Pull Request

This PR allows salbutamol inhaler refills to be purchased from the outpost medical vender, and replaces inhaler purchases with their cartridge type instead. A generic empty inhaler is sold for use alongside these cartridges. It also reduces the price of salbutamol refills from 200 credits to 100 credits.

## Why It's Good For The Game

I think it'd be nice to have salbutamol inhalers be more easily accessible outside of oxygen kits, and it would also be nice for their cartridges to be a bit cheaper

## Changelog

:cl:
add: added salbutamol refills to the outpost med
remove: replaces prescription inhalers and sting inhalers with the relevant cartridge type
balance: reduced price of salbutamol refill kits
/:cl:

### DIFF
--- a/code/modules/vending/medical_wall.dm
+++ b/code/modules/vending/medical_wall.dm
@@ -40,8 +40,8 @@
 		/obj/item/storage/pill_bottle/placebatol = 10,
 		/obj/item/reagent_containers/hypospray/medipen/placebatol = 10,
 		/obj/item/inhaler/placebatol = 10,
-		/obj/item/inhaler/salbutamol = 5,
-		/obj/item/reagent_containers/inhaler_canister/salbutamol = 5,
+		/obj/item/inhaler/medical = 5,
+		/obj/item/reagent_containers/inhaler_canister/salbutamol = 10,
 	)
 	premium = list(
 		/obj/item/reagent_containers/medigel/hadrakine = 3,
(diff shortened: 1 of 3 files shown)


code/modules/vending/medical_wall.dm
```diff
@@ -39,7 +39,9 @@
 		/obj/item/reagent_containers/syringe/antiviral = 4,
 		/obj/item/storage/pill_bottle/placebatol = 10,
 		/obj/item/reagent_containers/hypospray/medipen/placebatol = 10,
-		/obj/item/inhaler/placebatol = 10
+		/obj/item/inhaler/placebatol = 10,
```
Erikafox: Could you edit this list so you can buy an unloaded inhaler and then the cans separately, so there's no preloaded inhaler, then more cartridges for the same inhaler

Erikafox (changes requested):

code/modules/vending/medical_wall.dm
zimon9: Sure thing, but my intention of doing it like this was more so for convenience -- the cost of purchasing a placebo injector and then a salbutamol refill would've equaled the same cost of a full salbutamol inhaler

zimon9: Nevermind, I misinterpreted -- the changes are implemented and flow nicer than my earlier implementation

Erikafox (approved):

PR #6362, opened 2026-06-03
Scales reward for completing additional lode scans of a drill

zimon9:
## About The Pull Request

This PR makes it so that instead of any additional lode being equal to 200 credits, it scales by a factor of the drill mission class. Class 2 drills net 400 credits per additional lode, and class 3 drills net 600 credits.

Additional ore lodes are capped at 4 extra lodes.

## Why It's Good For The Game

This change should make it more worthwhile to do extra scans on higher-tiered drills. As it is now, obtaining additional lodes becomes less and less worth doing the higher the class of drill you're doing, and this change should address that.

## Changelog

:cl:
balance: scales additional lode redemption based on drill class, up to a maximum of 4 additional lodes
/:cl:

### DIFF
--- a/code/modules/missions/outpost/drill_mission.dm
+++ b/code/modules/missions/outpost/drill_mission.dm
@@ -228,3 +228,8 @@
 		say("Required samples gathered. Additional samples valued at [200*mission_class]cr per. Now shutting down!")
 		if(active)
 			stop_mining()
+
+	if(num_current > (num_wanted + 3))
+		say("Core sample collector full. Unable to continue drilling.")
+		if(active)
+			stop_mining()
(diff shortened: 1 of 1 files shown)


Erikafox:
I would be okay with even more bonus per node if it capped out at fixed amount of the mission's worth for additional payout (say 15%?)

zimon9:
> I would be okay with even more bonus per node if it capped out at fixed amount of the mission's worth for additional payout (say 15%?)

oh- yes that's a good idea
I'll cap it but I'll keep the reward bonus the same (it's really high as it is, I think? I'll change it as you deem would be best)

PR #6398, opened 2026-06-14
Fixes bullets ignoring directional structures if you're on the same tile

Gristlebee:
## About The Pull Request

Adjusts the the targeting logic of projectiles so they'll hit the directional objects/structures instead of immediately homing in on any mob on the same tile.
<details>
<summary>Video</summary>
Before:

https://github.com/user-attachments/assets/27b044ab-8ac9-4edf-bcaa-6383cdba73e1

After:

https://github.com/user-attachments/assets/1877e830-a1f7-4d16-9e82-10262f987079
</details>

This should also fix flipped tables not providing cover if you're standing on the same tiles as them.

## Why It's Good For The Game

Fixes are good. It doesn't particularly make sense to get hit by a bullet if you're standing behind a pane of reinforced glass or a firelock or whatever.

## Changelog

:cl:
fix: Bullets going through directional border objects if you're on the same tile.
/:cl:


firebudgy:
image

Erikafox:
You may want to consider untangling some of the logic that's attached to some of these objects (flipped tables) if this covers the case the logic was for

Gristlebee:
> You may want to consider untangling some of the logic that's attached to some of these objects (flipped tables) if this covers the case the logic was for

Don't believe anything should overlap. This mainly just concerns targeting order of operations and projectiles erroneously over prioritizing mobs.

rivercadaver:
no more two tap through firelocks on every space ruin ever i kneel

PR #6409, opened 2026-06-17
damage adjustments to the bayonets and gun melee

DIB-DOG:
## About The Pull Request
Increases gun melee from 5 to 10
Increases bayonet damage from 15 to 18
Increases energy bayonet damage from 19 to 20

Lets bayonets have knife stats like faster attack speed and embed chance.

## Why It's Good For The Game
Smacking someone with my big hunk of wood and metal doing the same damage as punching someone bare handed feels bad and doesn't make a lot of sense.

Bayonets up to this point were just survival knives that can stick to your gun and were meh at best. For the increase in price over a survival knife, this will make the bayonet tangibly better but not insanely powerful or as good as a combat knife.

The energy bayonet adjustment is just 1 damage difference and it makes things easier. It cost 500 credits which is more than a combat knife so it's fair to be on par with it's damage.

## Changelog

:cl:
balance: rebalanced gun related melee options
/:cl:

### DIFF
--- a/code/game/objects/items/attachments/bayonet.dm
+++ b/code/game/objects/items/attachments/bayonet.dm
@@ -12,6 +12,8 @@
 	hitsound = 'sound/weapons/bladeslice.ogg'
 	attack_verb = list("slashed", "stabbed", "sliced", "torn", "ripped", "diced", "cut")
 	sharpness = SHARP_POINTY
+	attack_cooldown = LIGHT_WEAPON_CD
+	embedding = list("pain_mult" = 4, "embed_chance" = 35, "fall_chance" = 10)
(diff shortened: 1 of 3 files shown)


Overteh:
Personally, my biggest problem with bayonets is that they are just a weak version of a survival knife. Their attack speed is noticeably lower than that of a knife, and there was no advantage in damage, and even after this change, the difference of 3 damage can do little.

DIB-DOG:
> Their attack speed is noticeably lower than that of a knife, and there was no advantage in damage, and even after this change, the difference of 3 damage can do little.

I’m hearing, “hey try adjusting attack speed” so I’ll touch that when I’m on my pc.

PR #6410, opened 2026-06-17
[DNM] removed flame range on 4.6x30mm tesla flux rounds

zimon9:
## About The Pull Request

Removes the flame range on tesla flux rounds

## Why It's Good For The Game

It always felt a tiny bit inconsistent that 4.6mm flux rounds were the only source of tesla bolts that caused fires -- and rather strong fires, at that. With how strong fires are, I don't think these projectiles should also cause fire damage on top of the shocks that they already provide.

## Changelog

:cl:
balance: removes flame range on tesla flux rounds
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/energy/tesla.dm
+++ b/code/modules/projectiles/projectile/energy/tesla.dm
@@ -32,7 +32,7 @@
 	power = 5000
 
 /obj/projectile/energy/tesla/explosive/on_hit(atom/target)
-	explosion(get_turf(loc),0,0,0,flame_range = 3)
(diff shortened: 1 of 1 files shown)


generalthrax:
Would like to see this in action for at least one round because I feel this may hit the effectiveness of the rounds quite a bit (though I may be wrong)

zimon9:
Placed on draft and the Do Not Merge tag has been applied, because this change makes the rounds rather underwhelming. I believe Bailey had some ideas for this, so I'll leave this up in order to allow us to be able to work on this.

PR #6429, opened 2026-06-23
QOL PGF lineup tweaks

DIB-DOG:
## About The Pull Request
I know I do this super often but they're my babies and I can't keep my hands off em, ¯\_(ツ)_/¯.

Started off adresssing some things that always irked me in the design of the Quasar and Cthonian's engineering areas on. Namely the lack of dedicated suit storage for the engineers and a generally cramped feeling.

After that point it came to my attention that for a medical response ship, the Cthonian was very cramped when trying to actually intake patients and was generally unfun to work inside of, which led me to redesigning the canteen and crew quarters areas to encourage better flow.

An assortment of other minor tweaks for player enjoyment later and now I've reorganized medbay supplies, added a new paper, put in a couple new decorations, and finally doles out coffee machines.

IMAGES:
image

image

## Why It's Good For The Game
These changes improve the quality of player interaction with the ships and in the case of the Cthonian, should make it less miserable to be crew on.

## Changelog

:cl:
add: Added some fluff and QOL reorganization to the PGF lineup
/:cl:


generalthrax (dismissed):
review, number of things

**Cthonian:**
- I would do some work with the Cthonian cargo bay, like here. It suffers badly from Kansatsu disease and is both difficult to manuever and store / access equipment. Expanding it out like this may require moving turrets but it would help significantly with ship flow. I would also add a door button and holofield button closer to the entrance.

Image

- Fix atmo here
Image

- Move this lightswitch
Image

- Dont need the second light imo
Image

- I would avoid continuing the thindow prison cells, try something like this to make better use of the space and avoid it being too claustrophobic, remove curtains
Image

- Move coffee machine and accessories to this table
Image

- Provide some tramal or another painkiller for the doctor

- Provide a few jetpacks to the EVA crate. I may be wrong, but I recall it being a boarding vessel, and these are helpful for EVA boarding

**Quasar:**
- Just make this a normal locker. Wall locker on an empty table may as well be a normal locker
Image

- Remove the curtains here, like on the cthonian. Don't need more pixel hunting on bunk beds
Image

- Remove the windows and windoor here
Image

- From personal experience I also found the quasar cargo bay to be difficult to use. I would remove the railings on the stairs and replace the sec camera console with a console table / reinforced table for easier use and space (replicate this). I would put one of the rechargers from the armoury in here as well (likely on the table)
Image

- You'll probably have to either remove the cameras or somehow fit them in elsewhere, but I think table space is more important than crew access to cameras (either remove them or put a console up in like the CIC or something of that sort)

- Give some tramal or some other painkiller to the medic

- Similarly move coffee accessories off the microwave
image


Also, for the both of these, I would look over them with new wallmounts to make sure theres no clipping (as I am writing this it is on an old branch)

DIB-DOG:
Ty for the review, all should be addressed as discussed.

PR #6457, opened 2026-07-02
Changes Salad Recipes to Prevent Explosions

### LINKED ISSUE #6009: Making a melon fruit bowl causes it to explode (and strangely does not require a bowl)
Issue reported from  Round ID: 4136 (Shiptest)

Reporting client version: 515.1642

Steps to reproduce -
1 - Obtain ingredients for a melon fruit bowl (minus a bowl)
2 - Create the bowlless melon fruit bowl.
3 - Explode upon its creation.

Image

Aquidu:
## About The Pull Request
Changes two salad recipes:
Jungle salad: Watermelon slices replaced with berries
Melon bowl: Bananas replaced with grapes

## Why It's Good For The Game
Fixes #6009 (nice)
https://github.com/user-attachments/assets/cc3f92a2-0230-4f27-b0f3-f5453219d97e

## Changelog

:cl:
fix: 2 salads no longer explode when crafted
/:cl:


generalthrax:
It may be preferable to simply remove potassium from bananas rather than adjusting the recipe they arent supposed to have so much they cause an explosion

Aquidu:
> It may be preferable to simply remove potassium from bananas rather than adjusting the recipe they arent supposed to have so much they cause an explosion

if its not enough to cause an explosion it just makes the food inedible somehow
i wasn't sure if i wanted to remove the reagents themselves in case that had disastrous consequences somewhere else, and thought it would still be fun if you could grind up bananas for potassium. i just thought that removing the connections themselves would be the most harmless solution to it

PR #6490, opened 2026-07-11
Makes planetary lighting inherit from ztrait

rye-rice:
## About The Pull Request

Basically planetary lighting is  put on the vlevel as a ztrait and turfs take from thatata snore honk shoo

More importantly, this means /lit subtypes can be killed, at long last. This PR doesnt do that, as i dont want to do more mapping repaths, but if /lit subtypes were removed tommorow planets would still be lit up, which is fucking cool

Hopefully sometimes we can do the same with planetary atmos, to get rid of that huge mess of planetary subtypes, but this is still great

By very techincality, this is Overmap Expansion #5.01, but its been so long i'm probably not calling it that

This cleans up and fixes up some other overmap code, preparing for a touchup of overmapgen sometime in the future

Mapping changes are to fix canister sprites but to also change areas on outposts to work with the system better

## Why It's Good For The Game

Better than /lit hell

Someday, maybe we can get rid of planet subtypes of plasteel tiles as well?

## Changelog

:cl:
add: Ruins should have somewhat better lighting
add: Hangars should now have weather show in them
/:cl:


code/modules/admin/verbs/randomverbs.dm
```diff
@@ -1103,6 +1103,58 @@
 	message_admins(span_big("Click here to jump to the overmap Jump point: " + ADMIN_JMP(point.token)))
 	BLACKBOX_LOG_ADMIN_VERB("Spawn Overmap Jump Point")
 
+/client/proc/setall_planetlights_inround()
```
Erikafox: You're going to want to make this star system specific

code/modules/overmap/planets/planet_types.dm
```diff
@@ -259,6 +311,10 @@
 	default_baseturf = /turf/open/water/beach/deep
 	weather_controller_type = /datum/weather_controller/waterplanet
 
+	light_color = "#09121a"
```
Erikafox, line 313: all the new colors should be defines

Erikafox (changes requested):

code/modules/admin/verbs/randomverbs.dm
rye-rice: done

code/modules/overmap/planets/planet_types.dm
rye-rice: To be honest, if we go through the idea of  having initial_atmos be a ztratit it probably will be better  to not have them defines for readablity; these are copied from the turfs and the colors on the turfs dont really do anything anymore, not really new

PR #6517, opened 2026-07-16
Portal Buildmode

Gristlebee:
## About The Pull Request

Adds portal buildmode. You click on two tiles as a start and destination and it sets up a portal pair. It also has an option to set up ladders.

## Why It's Good For The Game

Convenient tool for setting up portals/ladders ingame since they take a good bit of effort otherwise.

## Changelog

:cl:
add: Portal buildmode
/:cl:


kaylexis:
savior

PR #6530, opened 2026-07-21
Rockworld Ruin - New Gorlex Republic Mining Colony

Addust:
## About The Pull Request

image

adds new NGR ruin; the mining colony.
overrun by fauna and being scavenged by a Ramzi Clique force that escaped detainment inside the colony during the evacuation through the humble concealed himehabu and RUNNING THE HELL OUT OF THERE

has some crystal fauna among other things

loot:
- breaching hammer
- angle grinder (on a ramzi cliquist)
- some entrenching tools
- NGR mining hardsuits (2 on corpses)
- other assorted NGR milsurp, mostly on corpses
- mauler pistol (confiscation safe)
- super dirty sidewinder
- ringneck (in ramzi territory)
- R-23 viper w/ no spare ammo (in fauna territory)
- gaboon (one on a ramzi)
- a bunch of ramzis with guns, some of them may even drop
- weapon cleaning kit or so
- stocks of ammo, mining explosives, rations, most of what a colony would be ordering

threats:
- scattered ramzi clique force split between an improvised hideout and salvaging the colony
- hideout force: ramzi clique overseer (officer w/ civ boomslang), chef, combat engineer (dude w/ a ringneck), teemeres rover
- salvage force: ramzi hunter (illestren), ramzi doorkicker (gaboon), frontiersmen smuggler (buddy-buddy w/ the clique (they have lore dw) and has an asp), ramzi cutter (angle grinder) and a sidewinder dude
- a bunch of angry fauna, many of which crystal inside the evil hell cave (may add more)

## Why It's Good For The Game

cool ruin for the NGR with a premise that i think is cool

## Changelog

:cl:
add: The New Gorlex Republic has lost contact with several mining colonies in your area following reports of captured Ramzi Clique personnel and overwhelming numbers of fauna. We advise maintaining caution when investigating these sites.
/:cl:


thearbiber (dismissed):
* There is A Ton of money to be made off crystal corpses. I'd just keep the amount of crystal mobs to the alive ones here, maybe even get rid of 1 watcher. Two goliaths and one watcher is already around over 6 grand iirc. In its place I think crystal objects/grubs would get the point across.

Image

* I'd make the room layout here more like this. The rooms conforming to the walls is not pleasant

Image

* Use normal windows or just two air tanks. Thindows dont really look stable even if they are+not up to standard anymore.

Image

* I'd check through the ruin to use the off versions of the scrubbers/vents if they are open to the planetary atmos. Otherwise, it will cause huge wind spikes as the vents try to fight the planet filling the room with co2 IF power gets restored

Image

* fluff on the table is neat but it makes the hall really tight+doesnt really fit in the hall. Would move it someplace else that doesnt get chokepointed

Image

* Shave the table into the armory like this

Image

* Would move these lockers out the way so this space is easier to travel

Image

* Think the bar here would very much benefit from the building being larger, as shown in image. Kill the circled table aswell

Image

* Would shorten the airlock here so this space is also more walkable, like this. Something like a bin or vendor in the circled spot would look good

Image

* Believe a layout like this would be much easier to traverse

Image

Thats about all I can see, really. Some slight layout changes to kill the 1 tile movement should make this pretty good

generalthrax:
image

amulet in the wall here

generalthrax:
I would make the gabagoon enemies use the gaboon fire sound as well

Addust:
> I would make the gabagoon enemies use the gaboon fire sound as well

both done

PR #6540, opened 2026-07-23
Thousand Eyes Smelter fix and general optimizations

Gristlebee:
## About The Pull Request

Smelters on thousand eyes are now linked

Mapped Processing Units now link based off DX and DY vars rather than machinedirs. You now put in how many x or y tiles based on where the smelter is relative to the console is now. Existing mapped consoles have been swapped over to the new linking method.

## Why It's Good For The Game

Smelters work,  and makes mapping in consoles a lot more flexible in the future since you don't need to be on a matching direction with the smelters anymore.

## Changelog

:cl:
add: Processers consoles link off dx and dy
fix: Thousand Eyes smelters
/:cl:


code/modules/mining/machine_processing.dm
```diff
@@ -75,18 +75,21 @@
...
 	///Useless, kept to avoid maploading errors
 	var/input_dir = NORTH
 	var/output_dir = SOUTH
-	var/dist = 1
+	///Position of the smelter relative to the processing machine in tiles. Negative X is left, Positive X is to the right.
+	var/dx = 0
+	///Position of the smelter relative to the processing machine in tiles. Negative Y is down, Positive Y is up.
+	var/dy = 0
+	/// To help hook up to smelters on roundstart.
+	var/roundstart_id
```
Erikafox: assuming this is deadcode since it works off offsets w/ how you set it up. Please remove.

Offsets are functional but ID would be much more convenient for mappers & such, if you feel like aspiring to something.

Erikafox (dismissed):

code/modules/mining/machine_processing.dm
Gristlebee: Deadcode's removed. I might do ID down the line, but this is good enough for now.

Erikafox:
compile maps is currently failing, please take a peek so we can merge

PR #6562, opened 2026-07-28
Ivory remap

thearbiber:
## About The Pull Request
everythings the same except like. cash got dropped to 2.5k from 7k. SRM guide has more gear since its all exclusive anyhow, faction role can rep some faction gear. 1 more passenger slot

image
image
image

## Why It's Good For The Game
A pretty decent size increase for one of the smallest ships (its still semi small) but I feel like the luxury ship should feel more luxury. Also upkeeps some of it to be more in line today (7k starting cash makes it feel awkward to do round progression)

## Changelog

:cl:
add: ivory remap
/:cl:


generalthrax (dismissed):
Get rid of a bookcase here. Would block the window and I doubt anyones using up two bookcases in a round anyhow

Image

Replace the rack here with a pipe dispenser machine

Image

Include some soap for the bathroom, here

Image

I would include an extra cell for the SG-8 and a magazine for the Woodsman respectively. Bring them up to 3 each

Image

Clipboards and folders for the bridge here. Extra tablet as well

Image

I'd put some marker beacons along these exterior wing parts here

Image

Put some mugs / malleable glasses and a single large beaker in here

Image

Fairly small changes all things considered. Penguins a facking stoic 

Image

PR #6590, opened 2026-08-04
Makes painkillers make you less sleepy

googobbug:
## About The Pull Request

Makes tramal only make you sleepy at 30u (up from 5u) and morphine only make you sleepy at 10u (up from 5u).

## Why It's Good For The Game

Taking a tramal pen shouldn't put you in sleep hell for 20 minutes. Low doses of painkillers should be safe for surgery/RP, only hardcore juicers should get sent to sleep world.

## Changelog

:cl: cablefish
balance: Morphine and tramal make you less sleepy now.
/:cl:


IHaveNoOriginality:
I would say 15u isn't exactly low for a dose of morphine, that's half a bottle/a full syringe of a pretty strong opiod, and you can dose it for slight damage resistance. Thoughts about morphine causing sleep from 6u to 10u?

googobbug:
10u would also be fine if 15u is too much, the 15u just came from "half OD threshold"

googobbug:
thought about it for a little bit, morphine eepy threshold is 10u now

PR #6591, opened 2026-08-04
ngr spotter equivalent (EOD replacement)

thearbiber:
## About The Pull Request
image

image

## Why It's Good For The Game
Armor variety so you can think about what you want to buy past "let me just get the blood red again". Still fulfills the EOD role itself, also unhappy with the EOD just being a recolor.

## Changelog

:cl:
add: ngr spotter equivalent and eod replacement
/:cl:


Erikafox (commented):
rather than fluffing them as EOD suits I suggest leaning into having them be new-manufacture boarding suits.  Lower tech than the blood reds, so much easier to maintain for the NGR military. Similar idea to the spotter.

PR #6602, opened 2026-08-07
Remove Phorid TF

Breakingzap:
The code that causes plasma tiles to turn you into a phorid has been commented out, it now just kills you.  Let me know if I should just delete all of it, but idunno if its needed for some reason??? so its staying just disabled.

## Why It's Good For The Game
The transformation was buggy and also forcing people to play a different species is bad. Forcing people to play phorid of all species is double bad.

## Changelog

:cl:
del:Plasma tiles can no longer turn you into a phorid
/:cl:

### DIFF
--- a/code/modules/awaymissions/mission_code/snowdin.dm
+++ b/code/modules/awaymissions/mission_code/snowdin.dm
@@ -66,12 +66,9 @@
 			if(L)
 				L.adjust_fire_stacks(20) //dipping into a stream of plasma would probably make you more flammable than usual
 				L.adjust_bodytemperature(-rand(10,20)) //its cold, man
-				if(ishuman(L))//are they a carbon?
-					var/mob/living/carbon/human/PP = L
-
-					if(prob(35)) //used to be used to slowly convert limbs, I'm not changing the mechnical function outside of removing limb/species conversion, so plasma just kinda random crits now lol.
-						PP.adjustToxLoss(15)
-						PP.adjustFireLoss(25)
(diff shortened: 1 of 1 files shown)


iamthesilly (changes requested):
Might as well just delete it instead of commenting out. If it's ever getting added back in, it could definitely be implemented better.

code/modules/awaymissions/mission_code/snowdin.dm
```diff
@@ -67,48 +67,11 @@
 				L.adjust_fire_stacks(20) //dipping into a stream of plasma would probably make you more flammable than usual
 				L.adjust_bodytemperature(-rand(10,20)) //its cold, man
 				if(ishuman(L))//are they a carbon?
-					var/list/plasma_parts = list()//a list of the organic parts to be turned into plasma limbs
-					var/list/robo_parts = list()//keep a reference of robotic parts so we know if we can turn them into a plasmaman
 					var/mob/living/carbon/human/PP = L
```
iamthesilly: Can remove this too and the human check, applying the tox/fire damage doesn't require the target to be human

iamthesilly (changes requested):

code/modules/awaymissions/mission_code/snowdin.dm
Breakingzap: wait is the resolve the merge button
there was a button previously to do the suggestion, is that the button i just hit

iamthesilly (approved):

PR #6647, opened 2026-08-19
Pirate guns no longer vanish, instead dropping a broken gun

googobbug:
## About The Pull Request

Changes how human simplemobs drop guns.
Firstly: Instead of dropping no gun at all, missing the drop chance now drops an unusable broken gun.
Secondly: Guns from human simplemobs now have a random amount of bullets in the magazine/charge in the cell.

In order to curb the amount of free energy gun cells, they have a separate drop roll to be good or bad. Same as the weapon drop rate. It does not tell you if it's good or not so have your deckhand test them. Undamaged energy guns always have a good cell.

## Why It's Good For The Game

It's always felt kinda weird/gamey to have pirate guns just poof. This fixes that without giving people one million free guns.

## Changelog

:cl: cablefish
add: Human simplemobs, when failing the gun drop chance, now drop a broken gun instead of no gun.
balance: Guns from human simplemobs will drop with a random amount of bullets in the mag/charge in the cell.
/:cl:

### DIFF
--- a/code/datums/components/fullauto.dm
+++ b/code/datums/components/fullauto.dm
@@ -59,6 +59,10 @@
 /datum/component/automatic_fire/proc/autofire_on(client/usercli)
 	SIGNAL_HANDLER
 
+	if(isgun(parent))
+		var/obj/item/gun/stinky = parent
+		if(!stinky.actually_shoots)
+			return // NO
 	if(autofire_stat != AUTOFIRE_STAT_IDLE)
 		return
 	autofire_stat = AUTOFIRE_STAT_ALERT
--- a/code/modules/mob/living/simple_animal/hostile/human/human.dm
+++ b/code/modules/mob/living/simple_animal/hostile/human/human.dm
@@ -94,7 +94,7 @@
 			for(var/i = 0, i < rand(0, cosmetic_damage.magazine.max_ammo), i++)
 				qdel(cosmetic_damage.magazine.get_round()) // this feels kludgy but like. how else
 				cosmetic_damage.magazine.update_ammo_count()
-				cosmetic_damage.update_appearance()
+			cosmetic_damage.update_appearance()
 
 	// ENERGY - drain cell a random amount, apply separate drop chance to cell
 	if(istype(dropped_gun, /obj/item/gun/energy))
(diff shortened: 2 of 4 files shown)


Aquidu:
does this still mean that killing humanmobs will nearly gurantee a magazine/cell drop? i don't think it'd be too bad on ballistics, but a free energy cell per enemy might be pretty strong... also, is there any special treatment for non-magazine guns like pump shotguns or the illestren?

googobbug:
> does this still mean that killing humanmobs will nearly gurantee a magazine/cell drop? i don't think it'd be too bad on ballistics, but a free energy cell per enemy might be pretty strong... also, is there any special treatment for non-magazine guns like pump shotguns or the illestren?

No special treatment for non-mag guns. Those work the same.

Currently this does drop a mag/cell with every gun... Someone suggested that I should make some of the dropped cells blow up like the black market ones, which I'll probably be implementing to curb the amount of free e-gun cells.

googobbug:
Cells have a chance to be bad now. Chance of a good cell is the same as the weapon drop chance. No it doesn't tell you which ones are bad. Have fun.

generalthrax:
It seems the broken Shredders from Commander Tiyak and Frontie Heavy mobs aren't broken and you can use them

EDIT: Seems to be all automatic guns

thearbiber:
Seems to be related to firemode. If you toggle a SKM on semi auto, it doesn't work. If its switched to full, itll shoot.

googobbug:
INTERESTING. Automatic guns must not respect `actually_shoots`... Guh.

googobbug:
Hotfix pushed. Auto guns will now respect it.

googobbug:
Based on feedback:
Ballistic magazines now respect the drop chance too, to avoid handing out 200 free magazines.
Dud guns have a "broken" overlay so you can visually distinguish them without having to examine them all.
image

Randomguy523:
Unsure how to feel about the invisibly botched egun cells; feels like it becomes to either ignore dropped eguns entirely without checking them, get a bit gamey testing them, or get pranked by a surprise mechanic (which could be _occasionally_ funny)

PR #6654, opened 2026-08-20
New Bird Added: Caracara-Class Salvager

PhoenixTheRedandEvil:
## About The Pull Request
image
Behold! The Caracara-Class Salvaging Vessel. Makosso-Warra's finest new industrial vessel!

<details>
<summary>StrongDMM Media</summary>
image
image
</details>

Has a crew of 6:

- Captain
- Security Officer
- Engineer
- 2x Salvage Technicians
- Assistant

## Why It's Good For The Game
MW is broadly lacking of ships designed to engage with the current gameplay loop, and also would benefit from a smaller-crew ship. The Caracara provides both, and also provides a more "working man's" Warraship.

here's what a caracara looks like
image

## Changelog

:cl:
add: Caracara-Class Salvaging Vessel for Makosso-Warra
/:cl:


moffball19 (dismissed):
let's have a watch

cryo's a little bit dark, another light and perhaps an extinguisher cabinet in here would be nice
Image
<br>
this railing should go on the tile above rather than being pixelshifted up
Image
<br>
these portable atmos machines have an odd placement. i'd suggest moving them to the main hall with the pumps being put under the catwalk, then moving the coffee machine between the dorms and canteen. in their place you could leave tables or racks. (maybe replace above table with equipment racks and put reinforced tables by the lathe?)
Image
Image
<br>
a few more meds would be good. give them some dimorlin/tramal painkillers, some charcoal pills ('less' subtype) and a few splints
Image
<br>
the NGR entrenching tool is an odd placement here and doesn't strike me as a scrapping tool either, try a drill or old jackhammer instead
Image
<br>
a spare high capacity cell with the cell charger would be good
Image
<br>
the detail here is just a tiny bit lacking. put a blue hatch mark under the generator and a black catwalk + plating under the SMES. the black siding should continue to wrap around only to separate the tech grid floor from the grid floor
Image
<br>
top window here is missing a grille. change the suit storage units here to be the locked subtype, then give them and the equipment closets an access lock relevant to the salvage techs (should be "10" in this case i believe). if anyone else needs to access them, the captain can just leave it unlocked. also put a fire extinguisher in here
Image
<br>
N2 filter isn't attached to anything! the waste pipes (layer 4) need to find their way to the filters (layer 1) before they can attach
Image
<br>
otherwise, not much to say about this room. please lock the SSU like we did for the salvage techs and add a pipe dispenser next to it
Image
<br>
there's real goggles under your poster of goggles
Image
<br>
this door has a tag to be bolted but lacks a bolt button. if added, only keep the shutter over the window. as before, please lock the SSU and lockers.
Image
<br>
no cameras to monitor except the bodycams. consider adding a few?
Image
<br>
move this photocopier to replace one of the trash cans
Image
<br>
this room feels a bit sparse. another light, an extinguisher, and/or a poster would be good
Image
<br>
gimme a shower cabinet with **nothin**' (add the usual things you'd see, magpie for example). The button in here has the wrong name and wrong functionality to control bolts
Image
<br>
a pizza cutter would be flavourful
Image
<br>
unpixelshift the seats in here and move the button in the top left onto the wall above it. add some fluff folders and junk to the chest drawer. consider removing the desk lamp for table space and/or removing the bottom seat
Image
<br>
nudge your mobile port right by 1 tile so it lines up with other airlocks
Image
Image

PR #6657, opened 2026-08-20
Downed Transport Fix

DIB-DOG:
## About The Pull Request
Changes some minor aspects of the Icemoon downed transport to add to the mystery and make the story it's telling more logical.

Discussing changes with maptie team, keeping it as vague as possible on the github to preserve the element of surprise.

## Why It's Good For The Game
Downed Transport is a ruin I'm proud of, but with time, my plotholes have come to light fairly clearly to me, so this addresses those and I think makes the progression more fun.

## Changelog

:cl:
add: Added more intruige to downed transport
/:cl:


generalthrax (dismissed):
While you're doing this, I would either remove the decals in the shaft (they show improperly on planet turfs) or put some sort of tile underneath them like plating so they show properly. Remove the black border decals too, you can't pixelshift decals

image

DIB-DOG:
> While you're doing this, I would either remove the decals in the shaft (they show improperly on planet turfs) or put some sort of tile underneath them like plating so they show properly. Remove the black border decals too, you can't pixelshift decals
> 
> image


Should all be removed, slipped my mind, ty for the catch.

PR #6673, opened 2026-08-25
Removes Strafing from Melee Pirate Mobs

generalthrax:
## About The Pull Request

Removes the dodge strafing that melee pirate mobs do

## Why It's Good For The Game

Players can't move like them. Feels immersion breaking to have every guy with a shank pulling the Rust crackhead melee moves (which actually slows you down if you do diagonal move as a player) and keeping the same speed

## Changelog

:cl:
balance: Pirate melee mobs no longer strafe
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/human/clip.dm
+++ b/code/modules/mob/living/simple_animal/hostile/human/clip.dm
@@ -11,7 +11,6 @@
 	faction = list(FACTION_ANTAG_DEFECTORS)
 	loot = list()
 	check_friendly_fire = TRUE
-	dodging = TRUE
 	rapid_melee = 2
 	mob_spawner = /obj/effect/mob_spawn/human/corpse/ramzi
 
--- a/code/modules/mob/living/simple_animal/hostile/human/frontiersman.dm
+++ b/code/modules/mob/living/simple_animal/hostile/human/frontiersman.dm
@@ -13,7 +13,6 @@
 	footstep_type = FOOTSTEP_MOB_SHOE
 	mob_spawner = /obj/effect/mob_spawn/human/corpse/frontier
 	r_hand = /obj/item/melee/knife/survival
-	dodging = TRUE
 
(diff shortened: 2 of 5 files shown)


Erikafox (changes requested):
You will need to do this in a way that does not remove sidestepping behavior on the melee mobs (altering dodge_prob.. changing how dodging/sidestepping are called)

generalthrax:
Wholly disagree. For aforementioned higher level melee mobs like the supercollider they already have 60 bullet and 250 health (which players cannot have, even with mammoth) and wallstun sledgehammers (which players cannot do). I think its fair to strip off a little bit of impossibility from the mobs

Erikafox:
The dodging behavior is a balancing measure to keep melee mobs from being flat out chaff, and I agree with your main point about *every* melee pirate doing it being ridiculous, but I think it fits on the ones that are supposed to be elite, drugged up, or otherwise skilled combatants. I'm interested in hearing the why you disagree.

generalthrax:
"Fitting" or not, personally I prefer my mobs requiring skill to defeat rather than having to lucky to land a shot with a non-automatic and hope a mob doesn't spot dodge a bullet moving 700 feet per second just because they're supposedly tweaking out

Erikafox:
At a distance players are capable of the same feat, just not as fast because simplemobs aren't well made. Given that this variable also controls sidestepping in melee, you'll need to edit the code to maintain that element of skillful play. You can also look at the dodge_prob variable which controls how frequently they attempt to dodge on the approach of a mob.

>hope a mob doesn't spot dodge a bullet moving 700 feet per second

Guns are something we skimp realism on because if they were realistic, it would be unfun.

thearbiber:
Dont we already give high level mobs 250 health and full movespeed with 60 bullet armor full body? I dont see the issue on the rubber band shuffle being removed. It just sucks

Aquidu:
i honest to god thought these were intentional to make them more dangerous, it'll be nice being able to predict them better

PR #6681, opened 2026-08-27
Shell Casings no longer trip landmines

Breakingzap:
Shell casings no longer trip landmines.

## Why It's Good For The Game

Shell casings setting off nearby landmines is really frustrating + illogical (landmines usually require some degree of pressure, not just a shell casing bouncing off it) + it killed me in round so I have to nerf it.

## Changelog

:cl:
fix: Shell casings cannot set of landmines
/:cl:

### DIFF
--- a/code/game/objects/items/devices/mines.dm
+++ b/code/game/objects/items/devices/mines.dm
@@ -115,7 +115,7 @@
 
 /// Can this mine trigger on the passed movable?
 /obj/item/mine/proc/can_trigger(atom/movable/on_who)
-	if(triggered || !isturf(loc) || !armed || iseffect(on_who) || istype(on_who, /obj/item/mine))
+	if(triggered || !isturf(loc) || !armed || iseffect(on_who) || istype(on_who, /obj/item/ammo_casing)|| istype(on_who, /obj/item/mine))
 		return FALSE
 	return TRUE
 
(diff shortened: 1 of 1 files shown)


Erikafox:
rather than checking for casing why don't you do an item weight check? light items won't set off mines because they're not heavy enough is good logic

Breakingzap:
> rather than checking for casing why don't you do an item weight check? light items won't set off mines because they're not heavy enough is good logic

The code to check weight didn't like being on the "on_who" bit of code.

Erikafox:
if isitem(on_who)
var/obj/item = on_who
if on_who.w_class < weight_class_here
return false

code/game/objects/items/devices/mines.dm
```diff
@@ -115,8 +115,13 @@
 
 /// Can this mine trigger on the passed movable?
 /obj/item/mine/proc/can_trigger(atom/movable/on_who)
+
 	if(triggered || !isturf(loc) || !armed || iseffect(on_who) || istype(on_who, /obj/item/mine))
 		return FALSE
+	if isitem(on_who)
+		var/obj/item/stupidthing = on_who
+		if(stupidthing.w_class < WEIGHT_CLASS_NORMAL)
```
Erikafox:
```suggestion
		if(stupidthing.w_class < WEIGHT_CLASS_SMALL)
```

ping me when you approve this so i can merge the pr

PR #6682, opened 2026-08-27
Removes the ability for vending machines to advertise + Vending machines can have language now

Breakingzap:
## Why It's Good For The Game

A constant barrage of stationesque "jokes" gets very annoying. There is already a lot going on in the chat box, we don't need to waste chat space with "don't believe the reports! Smoke today!" Also more language is good, clipships can have League Zohillian speaking vendors.

## Changelog

:cl:

del: Ship vending machines recognize that they are objects and have no right to speak unless spoken to. (Some outpost vending machines can still talk)
add: Vending machines can now speak more than galactic common (only when spoken to first)

/:cl:

### DIFF
--- a/code/modules/ruins/lavalandruin_code/syndicate_base.dm
+++ b/code/modules/ruins/lavalandruin_code/syndicate_base.dm
@@ -17,5 +17,4 @@
 		/obj/item/grenade/chem_grenade/cryo = 5,
 		/obj/item/grenade/chem_grenade/adv_release = 5
 	)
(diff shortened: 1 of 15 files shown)


Erikafox (changes requested):
some of these vendor suck but the speaking behavior fits on ones like soda and snack machines. I'd just piggyback off the check for if the vendor is 'free' (in a ship) to see if they make the verbal calls + culling vendors that should stfu instead of wholesale removal.

PR #6688, opened 2026-08-28
Yet Another Kali-Andhi Remap

Addust:
## About The Pull Request

what it says on the tin

notable changes include:
- de-cucking the bridge
- slightly expanded armoury (now with C-20r)
- new Gryke-class subshuttle (the Karst remains so i don't have to mess with agni and because it's slightly smaller)
- hallway length cut by like 2/3 or something
- char said the ship wants a 3-tile-wide hallway so we're getting a 3-tile-wide hallway

image
- char also suggested messing with the gryke
image

## Changelog

:cl:
add: The Kali-Andhi-class Destroyer has been remapped, with such new and exciting features as "not having a cuck bridge" and "a fully-operational plasma-boosted hydrogen-mix thermo-electric generator", whatever that means.
add: The Karst-class dropship has been replaced in frontline service by the Gryke-class. Note that the Karst remains in rear-line service and thus remains posted at Agni Trading Post.
/:cl:


code/modules/atmospherics/auxgm/gas_types.dm
```diff
@@ -188,6 +188,7 @@
 	fire_products = list(GAS_H2O = 1)
 	fire_burn_rate = 2
 	fire_temperature = FIRE_MINIMUM_TEMPERATURE_TO_EXIST - 50
+	enthalpy = FIRE_HYDROGEN_ENERGY_RELEASED // like 2.8 megajoules or something. why was this not set
```
iamthesilly: This wasn't set because of how enthalpy works. The fuel itself is zero and its products are negative, for a net change equal to `FIRE_HYDROGEN_ENERGY_RELEASED`. The actual reason the TEG is so terrible with hydrogen is addressed by #6689.

iamthesilly (dismissed):

code/modules/atmospherics/auxgm/gas_types.dm
Addust: oh shit fr???? ill remove it to avoid conflicts but the teg will be useless until #6689 is merged

generalthrax:
Whoops that was a comment

generalthrax (dismissed):
Number of things:

- Bit of a squished little room here. Could probably bump it up by one more to make it less cramped without meaningfully affecting the top room
image

- Readjust the kitchen tables to be more like this. We also have countertop sinks now
image

- Never really was one for the old Kali remap's line of bunkbeds for the dorms, think cabins like this would be a better use of the space. You can probably put a TV into the mess hall, this is just a mock-up
image

- Advise putting valves instead of pumps in front of the thrusters and simply having a gas pump regulating safe pressure for the thrusters instead. Would also double-up on hydrogen tiles for the open air tank instead of one being airless
image

- These tiles are (as far as I know) not immune to being dislocated by vacuum so you may want to use external hull tiles instead
image

- Do not pixel shift computer consoles
image

- Do not think these firelocks here will help much in the case of leaks (will likely just trap someone in)
image

- You can include a little bit more plastitanium here for repair material (upwards of 20-25)
image

- Could do a similar treatment with the command bedrooms for something like this
image

- Get rid of the thindows here 
image

- Try to move these to walls to make for table space in the bridge
image

- Small suggestion for the cargo bay, just shifting stuff around to make good use of space and give the easy ability to include other things (along with encouraging people to use the whole space for storage / renovation)
image

- Use the new countertop sink here instead of the chemical one
image

- Use railings instead of platforms here (platform perspective looks off)
image

- Can probably just cut the railings here, the layering is spotty
image

- Avoid "diagonal" prechargers like this when possible (the Falmouth doesn't get a pass), arrange it something like this (you have some liberties here)
image

- Far too many buttons on this wall, they'd likely see better use by the doors themselves
image

- Double stairs down here into the TEG area
image

- May be a better place for this fireaxe out in a public area or engineering, only need one of each for the coffee bean packs, and these distress panels fit well on a quarter tile, it'd be good to move
image

Pretty solid so far, we'll have to see how the TEG functions when #6689 goes through

Addust:
> Number of things:
> 
> * Bit of a squished little room here. Could probably bump it up by one more to make it less cramped without meaningfully affecting the top room
> 
> image
> * Readjust the kitchen tables to be more like this. We also have countertop sinks now
> 
> image
> * Never really was one for the old Kali remap's line of bunkbeds for the dorms, think cabins like this would be a better use of the space. You can probably put a TV into the mess hall, this is just a mock-up
> 
> image
> * Advise putting valves instead of pumps in front of the thrusters and simply having a gas pump regulating safe pressure for the thrusters instead. Would also double-up on hydrogen tiles for the open air tank instead of one being airless
> 
> image
> * These tiles are (as far as I know) not immune to being dislocated by vacuum so you may want to use external hull tiles instead
> 
> image
> * Do not pixel shift computer consoles
> 
> image
> * Do not think these firelocks here will help much in the case of leaks (will likely just trap someone in)
> 
> image
> * You can include a little bit more plastitanium here for repair material (upwards of 20-25)
> 
> image
> * Could do a similar treatment with the command bedrooms for something like this
> 
> image
> * Get rid of the thindows here
> 
> image
> * Try to move these to walls to make for table space in the bridge
> 
> image
> * Small suggestion for the cargo bay, just shifting stuff around to make good use of space and give the easy ability to include other things (along with encouraging people to use the whole space for storage / renovation)
> 
> image
> * Use the new countertop sink here instead of the chemical one
> 
> image
> * Use railings instead of platforms here (platform perspective looks off)
> 
> image
> * Can probably just cut the railings here, the layering is spotty
> 
> image
> * Avoid "diagonal" prechargers like this when possible (the Falmouth doesn't get a pass), arrange it something like this (you have some liberties here)
> 
> image
> * Far too many buttons on this wall, they'd likely see better use by the doors themselves
> 
> image
> * Double stairs down here into the TEG area
> 
> image
> * May be a better place for this fireaxe out in a public area or engineering, only need one of each for the coffee bean packs, and these distress panels fit well on a quarter tile, it'd be good to move
> 
> image
> Pretty solid so far, we'll have to see how the TEG functions when #6689 goes through

all requested changes are in

PR #6699, opened 2026-08-30
Fixes missing kepori/vox variations on some clothing

iamthesilly:
:cl:
fix: Fixed missing kepori and vox variations on some clothing
/:cl:


iamthesilly:
Redundant, another fix is already opened

PR #6701, opened 2026-08-31
Radmine Walkmines and Significant Mob Radiation Behavior Changes

zimon9:
## About The Pull Request

This PR changes the behavior of radiation on mobs significantly. The mechanics of radiation were left untouched (so it shouldn't spread out of control as nothing about that has been altered).

Radiation dosage limits were increased across the board, and the amount of vomiting that radiation induces was reduced. Radiation also does significantly less toxin damage and also clears itself more rapidly. In exchange, radiation now causes rad burns more often, starting after passing the safe radiation threshold (increased from 500 to 1000). IPCs now also take rad burn damage, as well.

dreamseeker_hM5IlvtiKl

A new walkmine variant was introduced, the Radmine. Radmine sprites, design and behavior were all done by @thearbiber.

image

The radmine runs up to a mob and starts projecting medium-intensity radiation in a wide cone around it. If destroyed, it spews out heavy radiation in a wide circle around it.

## Why It's Good For The Game

The radiation changes were something I've thought about for a long time, but while working on the radmine, I found that I was presented with a binary: either I make the radmine accumulate radiation in players silently, and therefore appear innocuous and not scary, or I could have it project enough radiation to create a glow overlay and scare players, but in doing so, force them to either spend twenty minutes under a shower, use up a radiation treatment kit, or utilize a SSU's flash-fry setting to burn away the radiation.

I wanted the middle ground of a scary mob that, while a threat, would inflict injuries that could still be treatable in a reasonable way. 

I might've made it a little too forgiving, however, and so I think the behavior could benefit from tweaking.

## Changelog

:cl:
add: adds the radmine, a walkmine that sends high-intensity ionizing radiation into anyone that walks near
balance: significantly altered radiation's effect on mobs. Radiation now primarily deals burn damage with minor toxin damage, or pure burn damage if one is an IPC. Rad immune mobs are unaffected.
/:cl:


Breakingzap:
Hi so I found a rather significant issue with this in round. mining uranium (with a drill) will rapidly give you radiation burns, even through a radiation resistant hardsuit. We didn't even mine a lot and we were constantly ticking damage from the box. This is bad I think.

zimon9:
> Hi so I found a rather significant issue with this in round. mining uranium (with a drill) will rapidly give you radiation burns, even through a radiation resistant hardsuit. We didn't even mine a lot and we were constantly ticking damage from the box. This is bad I think.

Oh, thanks! I'll check this out in a local

code/__DEFINES/radiation.dm
```diff
@@ -11,32 +11,32 @@ Ask ninjanomnom if they're around
...
+#define RAD_MOB_SAFE 1000
 // How much stored radiation to check for hair loss
-#define RAD_MOB_SICKNESS 800
+#define RAD_MOB_SICKNESS 1500
 // How much stored radiation to check for mutation
-#define RAD_MOB_MUTATE 1250
+#define RAD_MOB_MUTATE 5000
 // The amount of radiation to check for vomitting
-#define RAD_MOB_VOMIT 2000
+#define RAD_MOB_VOMIT 2500
```
Erikafox, line 32: we don't mutate mobs w/ radiation any more so you can just remove that define

code/__DEFINES/radiation.dm
```diff
@@ -11,32 +11,32 @@ Ask ninjanomnom if they're around
...
 #define RAD_BURN_LOG_BASE 1.1
 #define RAD_BURN_LOG_GRADIENT 10000
 #define RAD_BURN_CURVE(X) log(1+((X-RAD_BURN_THRESHOLD)/RAD_BURN_LOG_GRADIENT))/log(RAD_BURN_LOG_BASE)
 
 // How much stored radiation in a mob with no ill effects
-#define RAD_MOB_SAFE 500
+#define RAD_MOB_SAFE 1000
 // How much stored radiation to check for hair loss
-#define RAD_MOB_SICKNESS 800
+#define RAD_MOB_SICKNESS 1500
```
Erikafox, line 28: we also don't force bald people any more but if you want to be clever, what you can do is look @ how elzu leaves worked for mexapix, and make a small hair loss overlay when someone is over the radiation sickness amount.

code/__DEFINES/radiation.dm
Erikafox: order of illness should be general malaise / light burning -> vomiting -> hairloss imo

code/__DEFINES/radiation.dm
```diff
@@ -11,32 +11,32 @@ Ask ninjanomnom if they're around
...
 // How much stored radiation to check for mutation
-#define RAD_MOB_MUTATE 1250
+#define RAD_MOB_MUTATE 5000
 // The amount of radiation to check for vomitting
-#define RAD_MOB_VOMIT 2000
+#define RAD_MOB_VOMIT 2500
 // Chance per second of vomitting
-#define RAD_MOB_VOMIT_PROB 0.5
+#define RAD_MOB_VOMIT_PROB 0.1
```
Erikafox, line 34: this is very low for something that should be a warning indicator, lets try halving it first

code/__DEFINES/radiation.dm
```diff
@@ -11,32 +11,32 @@ Ask ninjanomnom if they're around
...
 
 // How much stored radiation to check for stunning
-#define RAD_MOB_KNOCKDOWN 2000
+#define RAD_MOB_KNOCKDOWN 3000
 // Chance of knockdown per second when over threshold
-#define RAD_MOB_KNOCKDOWN_PROB 0.5
+#define RAD_MOB_KNOCKDOWN_PROB 0.1
```
Erikafox, line 39: ditto. this means you've gotten so radiactive that you're just getting blasted

code/modules/mob/living/carbon/human/species.dm
```diff
@@ -1444,7 +1444,7 @@ GLOBAL_LIST_EMPTY(roundstart_races)
  * - times_fired: The number of times SSmobs has fired
  */
 /datum/species/proc/handle_mutations_and_radiation(mob/living/carbon/human/source, seconds_per_tick, times_fired)
-	if(HAS_TRAIT(source, TRAIT_RADIMMUNE))
+	if(HAS_TRAIT(source, TRAIT_RADIMMUNE) || HAS_TRAIT(source, TRAIT_RADRESISTANT))
```
Erikafox, line 1447: rad resistant should reduce the 'effective' radiation a mob suffers rather than completely removing it.

Erikafox (changes requested):

code/__DEFINES/radiation.dm
zimon9: You're right, I'll just remove that, then. There's commented out mutation code, should I remove that too?
I believe I set up the switch a little strangely because I was looking at the number values rather than how the disease would realistically change depending on what thresholds would be crossed (ie. vomiting, balding, ect). I was wondering if I should remove the disease entirely, since it's not quite used?

code/__DEFINES/radiation.dm
zimon9: That could be fun, certainly. I'd need to see if I can figure it out/if it would look okay

code/__DEFINES/radiation.dm
zimon9: Sure, I'll halve it

code/__DEFINES/radiation.dm
zimon9: Sure thing, I'll halve it

code/modules/mob/living/carbon/human/species.dm
zimon9: The RADRESISTANT trait here is mainly for robotic enemies, because I figured it would be a little painful for them to experience toxin damage. I made it a generic RADRESISTANT trait in case I wanted to apply it to non-robotic enemies, but it _is_ named a little strangely

Erikafox: simplemobs don't have species. only humans do. just make it work consistently w/ the naming and we can expand around it later

zimon9: alright, I'll do that, thank you

googobbug:
The frequency for the warning message seems very high right now, on a round today when I went on a waste world I was shaking and getting the message about rad burns every few seconds even though I was not taking an appreciable amount of rads (3 burn total after several minutes of walking around).

zimon9:
> The frequency for the warning message seems very high right now, on a round today when I went on a waste world I was shaking and getting the message about rad burns every few seconds even though I was not taking an appreciable amount of rads (3 burn total after several minutes of walking around).

Ah, yeah, I didn't have a proper check to prevent the message occuring when negligible amounts of damage was occuring. I've tweaked it now so that it only happens for damage amounts of 0.4 per tick, with messages every 7 seconds. I'm probably going to see if I can put a baseline damage instead of using the strange log function for this.

code/__DEFINES/radiation.dm
zimon9: This'd be fun, but I want to make a toggle for particle effects first (because I know some players experience a lot of lag or crashes because of it). I'll potentially revisit this in a future expansion PR.

zimon9:
Reviews are addressed now.

PR #6702, opened 2026-08-31
Oasis Stash - Updates to River Valley Stash

stiel0248:
## About The Pull Request

Overhauls River Valley Stash to have better scenery, newer enemies, and to work fully as intended. Fully implements the Whollop Class Smuggling Pod, both for aspawn shenanigans and for crews to pick up if they desire. The ruin now fully serves as I originally intended as a very open map with significant sniper threats.

Modifies some melee enemies

image

image

Rewards:
So Many Fucking Drugs, Joywires, Beacon, Two Spitters, a cloaking device, a functional subshuttle, SKM Drum Mag (No SKM)

Dangers: Good luck getting there.

## Why It's Good For The Game

Ruin working as intended. Oasis stash's rewards are now more broadly applicable and hopefully provides a more interesting fight as one of the few long-range challenges in the game.

## Changelog

:cl: stiel0248
fix: River Valley Stash > Oasis stash. Large redesign.
/:cl:


generalthrax (changes requested):
Hello, just a couple things here

- So, biggest concern is that theres not any real feasible way to approach this in an intuitive way. The snipers with their enhanced vision range have pretty much all angles covered here, even from mapped in "alternate" routes (the west cave). Approaching from the South and West is suicide, even for potentially docking vessels with the 20 aggro range, leaving your only real way to enter the ruin without a long-ranged rifle is to dig in from the northeast

- I'd suggest decreasing their aggro range since they're F3 fronties, and moving the cave in area further east away from the vision of the eastern-most sniper.  Can bring the mountain forward a bit, as well (along with some greebs to make it more noticeable to people passing) (as a note, the aggro range would be fine if all of their aggro-ranges did not overlap and they were not F3 fronties, though that is not the case here)

Image
Image

- It'll probably more interesting if this west entry way wasn't covered with snipers (its a cave, after all) and was instead covered by traps and automated defenses (so you can get through without being shot at)

Image

- Given the volume of Frontiersmen enemies on this ruin, would probably be flavourful to include some temporary constructions like extra tunnels (similarly to the original ruin) along with cheap concrete hab-blocks along the north-and-east caves while they're getting this place set up to hide their stuff. 

Image

- This is a Zohilan cloaking device from the Zohil Explorat. The Whollop shouldn't have this (a normal cloaking gen should have no problem being charged with a vessel this small)
Image

- I would replace these two SKMs with either 2x Spitter mobs or 1x Spitter / 1x Pounder
Image

- These guys can probably be removed or moved to any new constructions you make  as per the top suggestions
image

- Shuttle will always spawn facing east like this. Would adjust the oasis itself to compensate. There's also no way to enter
Image

- Number of the var-edited trees on the ruin appear to be invisible

- Allow the normal Whollop to be spawned (add it to code/datums/shuttles.dm)

- Seems you have two variants of the populated shuttle
Image

- I would remove these vine covers too for the snipers, I see the intent but I don't think it works here
image

stiel0248:
Lot of great suggestions actually. Working on this.

- [x] Spitter fronties
- [x] Cloaking nerf
- [x] Swap f3's for illies (I feel the sniper's range is really important for making this ruin capital S Scary and makes it unique).
- [x] Figure out how to handle the vines (We're just killing them)
- [x] East entryway changes
- [x] West entryway changes
- [x] Tunnel network adjustments for living (Concrete buildings are too much, this is made to be a very subtle operation to a slight degree)
- [x] Make the Oasis far larger to accommodate the shuttle
- [x] Fix shuttle file and spawning crap
- [x] Why are the trees doing that

generalthrax (changes requested):
- This would be a good place to utilize a wired turret as per ical's pr, simply feed the output from the SMES into the turret with a wire node, and making the turret area itself part of the greater explored area (not the power section, though)

Image

- Given the above average number of high damage enemy weaponry this may be a fine place for an actual chopper (and not a glunked one) to be

Image

- You'll want to restore the removed files from shiptest.dme

- There's already a pr open for glunked guns, I would not include that here

Breakingzap:
Comments from playing this. Firstly the snipers in 1x1 boxes look weird and are weird to fight because you have to go up right next to them, or waste a ton of ammo blowing through all the sandbags. I liked being able to trench raid the last iteration. The hullpiercers are also not very fun, the only real options to deal with them are to either out dps them, or build a wall. Two supercollider type guys in the same room is over the top. The shuttle guys did seem a bit less prone to backshots in this version though, so thats cool.

thearbiber (changes requested):
I'd change out one of the hullpiercers for a shotgun or something. Two of them is shaping to be a little absurd.

Erring:
Two pyro anomalies aren't permanent. They disappear after short period of time after spawning

PR #6715, opened 2026-09-03
fixes gun pointblanks

zimon9:
## About The Pull Request
This PR makes it so that you can fire point blank into an enemy again instead of just smacking them with your gun. (when on harm intent)

If you do want to smack someone with your gun, you can go into disarm intent, or safety your gun.

This PR also fixes a minor runtime that happens when a gun with a magazine and a bullet in the chamber fires the gun while there's no magazine. The gun tries to pull information for the magazine it's supposed to have for ammo counters, but it assumes there'll always be one (and typically, this is true). The check was changed so now it's only a fuzzy check.

## Why It's Good For The Game

I've seen a lot of frustration expressed when characters smack an enemy instead of shooting them, which happens when the player controlling that character wanted to shoot that enemy instead of hitting them. This should resolve that issue. 

## Changelog

:cl:
fix: fixed being unable to point-blank enemies
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -252,8 +252,10 @@
 		zoom(user, user.dir, FALSE) //we can only stay zoomed in if it's in our hands	//yeah and we only unzoom if we're actually zoomed using the gun!!
 
 /obj/item/gun/attack(mob/M as mob, mob/user)
-	if(user.a_intent == INTENT_HARM || !actually_shoots) //Flogging
+	if(user.a_intent == INTENT_DISARM || !actually_shoots) //Flogging
 		return ..()
+	else
+		afterattack(M, user)
 	return
 
 //called after the gun has successfully fired its chambered ammo.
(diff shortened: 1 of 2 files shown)


DIB-DOG:
Does RMB still melee/bayonet with this change?

zimon9:
> Does RMB still melee/bayonet with this change?

It didn't before, but I've fixed it to work now. Thank you. It should work for both disarm and harm intent right-clicks. Disarm left clicks also work. Strangely, help intent fires the gun no matter if it's a left click or a right click.

code/modules/projectiles/gun.dm
```diff
@@ -40,6 +40,7 @@
 	var/fire_sound_volume = 50
 	var/dry_fire_sound = 'sound/weapons/gun/general/dry_fire.ogg'
 	var/dry_fire_text = "click"
+	var/allow_fire = TRUE //temporary variable that switches when attempting to bash an enemy on disarm intent
```
iamthesilly: This doesn't seem to be used anywhere except the check?

zimon9: oh whoops, I think I left that in by accident while trying to fix this
I'll remove that, thank you

zimon9: Should be fixed now, thank you

PR #6722, opened 2026-09-04
prevents runtimes with T4 drill missions

zimon9:
## About The Pull Request

This PR prevents a runtime that occurs with T4 drill missions are removed. The Destroy proc recalls the mission drill, but if the mission was never selected, the drill never spawned. It now has a check that ensures the drill is spawned in before attempting to delete it.

## Why It's Good For The Game

The less runtimes there are, the better

## Changelog

:cl:
fix: fixes runtimes with T4 drill missions
/:cl:

### DIFF
--- a/code/modules/missions/outpost/industrial_drill.dm
+++ b/code/modules/missions/outpost/industrial_drill.dm
@@ -41,7 +41,8 @@
 
 /datum/mission/acquire/industrial_drill/Destroy()
 	. = ..()
-	recall_bound(mission_drill, FALSE)
+	if(mission_drill) //a check to prevent runtimes from an attempt to delete a mission drill that was never spawned in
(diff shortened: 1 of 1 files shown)


code/modules/missions/outpost/industrial_drill.dm
```diff
@@ -41,7 +41,8 @@
 
 /datum/mission/acquire/industrial_drill/Destroy()
 	. = ..()
-	recall_bound(mission_drill, FALSE)
+	if(mission_drill) //a check to prevent runtimes from an attempt to delete a mission drill that was never spawned in
+		recall_bound(mission_drill, FALSE)
```
Erikafox, line 45: set mission_drill to null as /datum/mission/drill/destroy does instead

PR #6723, opened 2026-09-04
Export console fixes

ritorizo:
## About The Pull Request

This PR correct two bugs
1. The price displayed on the export consoel was overestimated if there where more than one item of the same kind on the pad.
2. The decaying of price decayed the price of the first item. The sale price is now exactly the same as if each item where sold individualy in a for loop.

## Why It's Good For The Game

[
exportybug
](url)

## Changelog

:cl:
balance: The price of the first item is no longer decayed at the export console.
fix: The export console now display the actual price at which the items will be sold.
:cl:


ritorizo:
Should I write up a proof?

PR #6724, opened 2026-09-04
Decreases aggro range of the tesla hermits

Sadhorizon:
## Why It's Good For The Game
My deckhand (extremely new) walked a bit too close, aggrod this thing from off-screen, and turned half of my crew into paste that doctors were scared to fix, basically ending the round there and then. I do not think that's particularly fun. 
Counterplay is basically "know this one ruin has those". Strong enemies like this should be reactable to without metaknowledge of their existance.

They still react if you shoot them also.

## Changelog

:cl:
balance: Decreased aggro range of tesla rifle hermits.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/human/survivors.dm
+++ b/code/modules/mob/living/simple_animal/hostile/human/survivors.dm
@@ -193,6 +193,7 @@
 	casingtype = /obj/item/ammo_casing/c46x30mm/tesla
 	r_hand = /obj/item/gun/ballistic/automatic/smg/skm_carbine
 	mob_spawner = /obj/effect/mob_spawn/human/corpse/damaged/whitesands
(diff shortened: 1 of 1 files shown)


Sadhorizon:
Maybe a bit of spite PR but this sucked so hard.

Addust:
gonna throw my two pence in. WHY does this ruin need hermits that shoot explosives as EXTERIOR GUARDS in the first place???? maybe like 1 or 2 in the interior i could justify but getting stunlocked and incinerated by the dude at the front door doesnt seem good for progression

Erikafox:
I looked @ this and I feel the effectiveness of the tesla round is more so related to the fact that explosions have a 'gentle' argument that does nothing. If it stopped explosions from throwing people around, and was flagged as true on the round, I think that might be a good solution.

PR #6735, opened 2026-09-05
updates maploaded sprites for spent autoinjectors

zimon9:
## About The Pull Request

Injectors that spawn in on a map already used don't _appear_ spent, which can be confusing sometimes. This fixes that so that their icons properly update.

Also fixes a bug with beakers so that the cap overlay is properly layered.
Before: 
image

After: 
image

Lastly, an autoinjector pen helper tool was added.

## Why It's Good For The Game

This is a fix to ensure that icons properly update when something is maploaded, that reagent caps are properly layered, and the helper may be helpful? I'm actually not very familiar with them nor what this one can be useful for.

## Changelog 

:cl: @zimon9, @Erikafox 
fix: fixed autoinjector sprites not updating upon mapload
fix: cap overlay layering inhand icon for reagent containers
add: reagent pen helper tool
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/hypospray.dm
+++ b/code/modules/reagents/reagent_containers/hypospray.dm
@@ -124,6 +124,12 @@
 	icon_state = "[base_icon_state][(reagents.total_volume > 0) ? null : 0]"
 	return ..()
 
+/obj/item/reagent_containers/hypospray/medipen/Initialize(mapload, vol)
+	. = ..()
+	if(!reagents || !reagents.reagent_list.len) //maploaded used up reagents
+		update_icon_state()
+		update_appearance()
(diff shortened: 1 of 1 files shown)


Erikafox (changes requested):
discussed in dev channel

zimon9:
> discussed in dev channel

Addressed the reviews, added the helper and fix to beaker cap layers
I attempted to use src + 0.1, but the cap failed to be layered in at all (I suspect that upon initialize, the layer is set to null, so it doesn't get layered in at 0.1, or runtimes and fails to render at all. Either way, I had to hardcode the cap layer at 50 (the default beaker layer) + 0.1.

PR #6748, opened 2026-09-07
Adds a single-use spray for dehusking into cargo.

TheUSSWatercooler:
## About The Pull Request
Adds a 100u, single-shot spray of synthflesh into cargo for de-husking a patient.

## Why It's Good For The Game
This gives crews an option that isn't getting the deckhand to ghettochem.
Original suggestions said to have this option priced at 1000 credits, but I feel like 800 is a better middle ground, still more expensive than making it by yourself via brute kits, but not equal in price to bundles of painkillers and whole crates of blood.

## Changelog

:cl:
add: Adds a high-strength synthflesh spray for dehusking into cargo.
:cl:

### DIFF
--- a/code/modules/cargo/packs/medical.dm
+++ b/code/modules/cargo/packs/medical.dm
@@ -68,6 +68,12 @@
 	cost = 100
 	contains = list(/obj/item/reagent_containers/inhaler_canister/salbutamol)
 
+/datum/supply_pack/medical/dehusk_single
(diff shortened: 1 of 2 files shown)


TheUSSWatercooler:
Drafting this, coming back after a day to think about it and I'm realizing I made something that's a patchjob at best and shoddy at worst with some obvious holes, gonna come back when I have time this week.

PR #6754, opened 2026-09-08
9mm smg firerate "buff"

thearbiber:
## Why It's Good For The Game
Changes them to shoot faster (the speed they are intended to)

So it turns out 1 fire delay rounds up to 1.5 (same as PDWs) which kinda leaves the 9mm smg as just an eternally shittier pdw. Also buffs the spitters spread so it doesnt get immediately outclassed shooting at the same speed as the others

## Changelog

:cl:
balance: 9mm smg firerate buff
balance: spitter spread decrease

/:cl:

### DIFF
--- a/code/modules/projectiles/guns/manufacturer/clip_lanchester/ballistics.dm
+++ b/code/modules/projectiles/guns/manufacturer/clip_lanchester/ballistics.dm
@@ -206,7 +206,7 @@ NO_MAG_GUN_HELPER(automatic/pistol/cm357)
 	weapon_weight = WEAPON_LIGHT
 	fire_sound = 'sound/weapons/gun/smg/cm5.ogg'
 	manufacturer = MANUFACTURER_MINUTEMAN
-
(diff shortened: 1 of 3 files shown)


Apogee-dev:
so i noticed you buffed the spitter because the other smgs now fire at the same speed- there isn’t any way to maintain spitter’s fire rate advantage, is there? or is this an engine limitation thing

thearbiber:
> so i noticed you buffed the spitter because the other smgs now fire at the same speed- there isn’t any way to maintain spitter’s fire rate advantage, is there? or is this an engine limitation thing

As of right now , Fire rate works like this:
Thing can either fire at .9 (which is actually 1 firespeed)
or they can fire at .4 (which is actually .5 firespeed)
Fixing this is like, way out of my league and it straight up might be an engine limitation.

if we want to make the spitter fire faster, it will have to shoot at pounder speed. We could do that, but that is like, ludicrously fast and id probably want to talk to a couple about it first

PR #6773, opened 2026-09-12
clover guns AP increases

thearbiber:
## Why It's Good For The Game

While theyre not performing bad, they definitely feel lacking against armor despite being advertised for armor.

## Changelog

:cl:
balance: increases clover guns AP by a smidge across the board
/:cl:

### DIFF
--- a/code/modules/cargo/packs/gun.dm
+++ b/code/modules/cargo/packs/gun.dm
@@ -448,9 +448,9 @@
 /datum/supply_pack/gun/energy/shillelagh/indie
 	name = "PL-12 'Shillelagh' Energy SMG"
 	desc = "Contains a single PL-12 Submachine Gun, a civilian grade energy gun with a burst fire mode manufactured by Clover Photonics. Uses Eoehoma cells."
-	cost = 2000
+	cost = 3000
 	contains = list(/obj/item/storage/guncase/pl12)
-	faction_discount = 10
+	faction_discount = 0
 	faction_locked = FALSE
 
 /datum/supply_pack/gun/energy/faveleira
--- a/code/modules/projectiles/projectile/beams.dm
+++ b/code/modules/projectiles/projectile/beams.dm
@@ -121,8 +121,7 @@
 	impact_effect_type = /obj/effect/temp_visual/impact_effect/blue_laser
 
 /obj/projectile/beam/laser/clover/smg
-	damage = 20
-	armour_penetration = 10
+	damage = 25
 
 /obj/projectile/beam/laser/clover/magnum
 	icon_state = "clover_pulse"
--- a/code/modules/projectiles/projectile/beams.dm
+++ b/code/modules/projectiles/projectile/beams.dm
@@ -122,17 +122,17 @@
 
 /obj/projectile/beam/laser/clover/smg
 	damage = 20
-	armour_penetration = 0
+	armour_penetration = 10
 
 /obj/projectile/beam/laser/clover/magnum
 	icon_state = "clover_pulse"
 	damage = 25
-	armour_penetration = 0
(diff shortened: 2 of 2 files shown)


Apogee-dev:
it feels sorta odd to me to make the shillelagh PDW-like (comparable to sidewinder and resolution) when it’s the only actual energy smg that’s universally available without playing the black market lottery. seems like the kind of thing that’d be more of a baseline experience, i guess? Ngl i thought the “anti-armor” bit was meant more in comparison to the Volt’s godawful AP

thearbiber:
It kinda suffers from the fact it cant really have expanded magazines 90% of the time, so you get stuck with about 15 shots? At its full potential using extendeds, it does perform similarly but you cant really reach that most the time unless you play CLIP (who have two other smgs) also the fact it costs way more. It also does the same damage as the rush so it does just end up feeling like a rush with less shots. The other alternative would be making it do 25 damage to make it in line with the 10mm SMGs, may just do that instead actually

PhoenixTheRedandEvil:
> it feels sorta odd to me to make the shillelagh PDW-like (comparable to sidewinder and resolution) when it’s the only actual energy smg that’s universally available without playing the black market lottery. seems like the kind of thing that’d be more of a baseline experience, i guess? Ngl i thought the “anti-armor” bit was meant more in comparison to the Volt’s godawful AP

I did sort of intend for the shillelagh to be sort of the 'pdw-like' energy SMG when i was putting them in. I think generally the lack of energy SMGs is also just the byproduct of there being like, no public or indie energy guns.

Apogee-dev:
Why is it that most of our energy guns are designed around extended mags being the “standard” capacity and being sort of crippled without them, anyway? I mean, expanded mags for the ballistics mean they actually feel _expanded_ rather than brought up to baseline when you have access to them. It feels like that whole thing’s been an underlying cause for a good portion of our weird energy weapon feel, tbh. especially since it mostly just means indies who barely have access to any energy weapons to begin with just get shafted while the factions always get extendos

thearbiber:
> Why is it that most of our energy guns are designed around extended mags being the “standard” capacity and being sort of crippled without them, anyway? I mean, expanded mags for the ballistics mean they actually feel _expanded_ rather than brought up to baseline when you have access to them. It feels like that whole thing’s been an underlying cause for a good portion of our weird energy weapon feel, tbh. especially since it mostly just means indies who barely have access to any energy weapons to begin with just get shafted while the factions always get extendos

agreed, but thats kinda out of scope and would require rebalancing every energy gun ever.

Apogee-dev:
> > it feels sorta odd to me to make the shillelagh PDW-like (comparable to sidewinder and resolution) when it’s the only actual energy smg that’s universally available without playing the black market lottery. seems like the kind of thing that’d be more of a baseline experience, i guess? Ngl i thought the “anti-armor” bit was meant more in comparison to the Volt’s godawful AP
> 
> I did sort of intend for the shillelagh to be sort of the 'pdw-like' energy SMG when i was putting them in. I think generally the lack of energy SMGs is also just the byproduct of there being like, no public or indie energy guns.

ah yeah, when i drew up the concepts for the clover line i was thinking of the shilly basically being an energy counterpart to the indie cobra on account of the indies-have-no-lasers problem. bit of a disconnect in thinking/communication.

Though its pretty funny to me that if the shilly gets a damage buff it would land right back in that niche anyway

Apogee-dev:
> agreed, but thats kinda out of scope and would require rebalancing every energy gun ever.

Actually, come to think of it, wouldnt it mostly just impact the balance for indie energy guns? the factions already have open access to upgraded cells for their exclusives (and some come packaged with em), so… Hm. definitely material for a different PR tho

PR #6774, opened 2026-09-12
Slow Down Tortoise! (small tortoise nerf)

PhoenixTheRedandEvil:
## About The Pull Request
Makes the Tortoise significantly slower (speed 4 instead of 1). Increases the amount of charges it does to compensate

## Why It's Good For The Game
Turtle too fast. Supercollider-style mob that ALSO had a shotgun and ALSO laid prox mines at your feet ontop of being a big bullet sponge made it feel like it was garunteed to kill almost everyone who fought it, which made it hard to use in events.

## Changelog

:cl:
balance: turtle slow
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/megafauna/megabot.dm
+++ b/code/modules/mob/living/simple_animal/hostile/megafauna/megabot.dm
@@ -122,6 +122,8 @@
 			spread_shot()
 		if(prob(20))
 			plant_mine()
+		if(prob(20))
+			charge()
 
 /mob/living/simple_animal/hostile/megafauna/megabot/proc/burst_shot()
 	casingtype = /obj/item/ammo_casing/energy/laser/assault/sharplite/megabot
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/hostile/megafauna/megabot.dm
```diff
@@ -25,7 +25,7 @@
 	melee_damage_lower = 50
 	melee_damage_upper = 50
 
-	speed = 1
+	speed = 4
 	move_to_delay = 3
```
thearbiber, line 29: Set move_to_delay to 5 if you want them to be slower while still deadly. Speed does not dictate mob speed

thearbiber (changes requested):

code/modules/mob/living/simple_animal/hostile/megafauna/megabot.dm
PhoenixTheRedandEvil: > Set move_to_delay to 5 if you want them to be slower while still deadly. Speed does not dictate mob speed

addressed this in a timely manner (a week for a line of code)

PR #6781, opened 2026-09-13
Planet Ruin baseturf replacement during generation

prodirus:
## About The Pull Request

image

**This should probably be TMed.**

So ruins are basically dropped on top of generated map terrain.  Destroying the ruin turfs exposes the underlying generated map tile, which in the case of lava planets can be on top of lava lakes. This has the potential to lead to very bad times™.

This PR adds an (optional) new property ``ruin_baseturf`` to planet types, and a proc ``sanitize_ruin_baseturfs`` that runs after the ruin loads to convert a predefined list of turfs into the provided "safe" ``ruin_baseturf``. If none is defined it just skips the proc, so it should be backwards compatible with all existing planet types.

It only runs when the planet is generated, so when someone is first landing in normal cases. Performance impact on my machine is pretty low, taking about an additional 0.2-0.5s to run planet gen on with this proc (with total planet gen taking about 10s)

I've probably approached this issue in an insane manner but idk man. The simplest way would have been to stop ruin turfs from being destroyed at all but that seemed like it would have a larger blast radius. Unsure how this would interact with multi-Z stuff if that ever becomes a thing.

## Why It's Good For The Game

Observed a round recently where someone blowing up a landmine on the Lava planetoid Dynamite Hole ruin dropped them into a lava pool, and it kinda seemed pretty ridiculous.

This hopefully prevents this from happening in the future.

## Changelog

:cl:
code: added a new property and sanitization proc to planet types for ruin generation, allowing the replacement of hazardous turfs that spawn under ruins.
/:cl:


prodirus:
Gonna do it the simpler way, changing baseturfs.

PR #6804, opened 2026-09-18
Hand To Hand Stun Removal

Breakingzap:
## Why It's Good For The Game

Wallshoves are conceptually stupid and not very fun. The ability to chainstun knocked down people by shoving is also bad.  Punches being the only melee attack that can randomly stun and also randomly miss is weird. Agg grab making you drop items makes it weirdly easy to cause someone who isn't immediately lighting you up with gunfire to drop their gun. 

## Changelog

:cl:
balance:Pushing someone into a wall no longer causes them to be completely helpless.
balance:Punches no longer have a chance to cause knockdown.
balance:Punches no longer have a chance to miss
balance:Aggressively grabbing someone no longer causes them to drop whatever they're holding.
balance:Shove chainstuns have been removed.
/:cl:


Erikafox (changes requested):
If we're removing wall shoving (which im not sold on) then we should re-implement actual disarm intent like other servers have.

Erikafox:
Please make a project post if you're interested in pursuing that, and so we can actually discuss this.

